PR #5, opened 2019-04-09
Refactor Plugin PHP as a Class

jeffersonrabb:
This PR brings in a few improvements to the plugin:

1) The main plugin PHP file is refactored as a class: `Newspack_Blocks`
2) View-side block registration and asset loading is more sophisticated. If a block directory contains a `view.php` file, this will be included and used for block registration and rendering. If there is a `view.js` file but no `view.php` then a generic block registration is created which simply enqueues `view.js` & `view.css`.
3) All script enqueuing now relies on generated dependency lists created by `@automattic/wordpress-external-dependencies-plugin`, rather than requiring manual dependency management.
4) Introduces a new example block which demonstrates server-side rendering.

To test:

- Run `npm ci && npm install && npm run clean && NODE_ENV=production npm run build`
- Install/Activate the plugin
- Create a post, open the block picker, verify there is a Newspack category and there are three blocks within it.
- Verify the three blocks work as expected.

### DIFF
--- a/newspack-blocks.php
+++ b/newspack-blocks.php
@@ -84,7 +84,7 @@ class Newspack_Blocks {
 	/**
 	 * Enqueue view scripts and styles for a single block.
 	 *
-	 * @param string $type The block's slug.
+	 * @param string $type The block's type.
 	 */
 	public static function enqueue_view_assets( $type ) {
 		$style_path  = NEWSPACK_BLOCKS__BLOCKS_DIRECTORY . $type . '/view' . ( is_rtl() ? '.rtl' : '' ) . '.css';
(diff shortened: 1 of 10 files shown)


claudiulodro (approved):
After the latest commit, this works well.

Just noting in case it's useful for others: this PR doesn't work on WP 5.1 without the Gutenberg plugin active (editor crashes hard). It works fine on WP 5.2. It shouldn't be an issue since this won't go live until after WP 5.2.

jeffersonrabb:
> Just noting in case it's useful for others: this PR doesn't work on WP 5.1 without the Gutenberg plugin active (editor crashes hard). It works fine on WP 5.2. It shouldn't be an issue since this won't go live until after WP 5.2.

This is most likely https://github.com/WordPress/gutenberg/issues/14580. The quick and dirty fix is to run `localStorage.clear()` in the Javascript console following a downgrade of Gutenberg.

PR #19, opened 2019-05-17
AMP compatibility for home page article block

philipjohn:
This change removes the view.js and related enqueueing code to ensure AMP compatibility. When using the homepage articles block, the enqueueing of view.js, which has `wp-polyfill` as a dependency, breaks AMP compatibility because of the polyfill scripts inserted by WordPress core.

Without view.js the homepage articles block still works as expected.


jeffersonrabb:
Rather than removing entirely, I'd prefer we exclude `view.js` based on being in an AMP environment. Even though the homepage articles block doesn't require any view-side scripting there may be others in the future that do, there may be situations where a block requires these scripts in non-AMP environments, and we want to keep open the possibility that these blocks can be used outside of Newspack. Here's Jetpack's AMP environment check: https://github.com/Automattic/jetpack/blob/master/3rd-party/class.jetpack-amp-support.php#L56-L72

philipjohn:
As an in-between could we only include view.js once we start needing it, and introducing the AMP check at that point? At the moment, it seems like view.js is unnecessary regardless of AMP compatibility (or I might be missing something).

There are possibly other alternatives too:

1. Do we need `wp-polyfill` as a dependency at all?
2. Can we use `amp-script`?

jeffersonrabb:
Alternate approach proposed here: https://github.com/Automattic/newspack-blocks/pull/20

philipjohn:
Closing in favour of #20

PR #20, opened 2019-05-17
Alternate Approach To AMP Compatibility

jeffersonrabb:
This is an alternate approach to https://github.com/Automattic/newspack-blocks/pull/19, in which rather than removing `view.js` entirely we enqueue is only when not in an AMP request. This approach keeps open the possibility of blocks with view-side scripts, which may be used in AMP & non-AMP environments.

Testing Instructions:

1) Install AMP plugin on test site along running Newspack plugin, so to Native mode.
2) Add Newspack Homepage Articles block to a Post or Page and publish.
3) Review AMP errors. There should be none related to `view.js` or `wp-polyfill`.

### DIFF
--- a/newspack-blocks.php
+++ b/newspack-blocks.php
@@ -97,7 +97,7 @@ class Newspack_Blocks {
 				NEWSPACK_BLOCKS__VERSION
 			);
 		}
-		if ( file_exists( NEWSPACK_BLOCKS__PLUGIN_DIR . $script_path ) ) {
+		if ( ! self::is_amp_request() && file_exists( NEWSPACK_BLOCKS__PLUGIN_DIR . $script_path ) ) {
 			$dependencies = self::dependencies_from_path( NEWSPACK_BLOCKS__PLUGIN_DIR . "dist/{$type}/view.deps.json" );
 			wp_enqueue_script(
 				"newspack-blocks-{$type}",
@@ -143,6 +143,15 @@ class Newspack_Blocks {
 			array_push( $classes, $attributes['className'] );
 		}
 		return implode( $classes, ' ' );
+	}
+
+	/**
+	 * Does the page return AMP content.
+	 *
+	 * @return bool $is_amp_request Are we on am AMP view.
+	 */
+	public static function is_amp_request() {
+		return ( function_exists( 'is_amp_endpoint' ) && is_amp_endpoint() );
 	}
 }
 


laurelfulford (approved):
I can confirm that this update removes errors related to view.js and polyfill when AMP is enabled, and that the files are still enqueued when it's not :)

philipjohn:
> only when not in an AMP request

If we go AMP Native, this isn't a situation we're going to have, because everything will be AMP all the time. If we _don't_ go AMP Native, and use this method it'll mean testing everything in both AMP and non-AMP.

jeffersonrabb:
> If we go AMP Native, this isn't a situation we're going to have, because everything will be AMP all the time. If we don't go AMP Native, and use this method it'll mean testing everything in both AMP and non-AMP.

Yes but note that this block may end up being used outside of the context of Newspack, so we should at least leave the door open to scripts if needed.

philipjohn:
> outside of the context of Newspack, so we should at least leave the door open to scripts if needed.

Outside of Newspack, if a script is needed to accompany this block, whoever is using it should be able to simply hook in appropriately to enqueue their own script. Otherwise, we're expecting them to modify our empty script, which essentially means forking the plugin.

philipjohn:
My thinking here to be clearer is:

1. We don't need any JS right now
2. We do need AMP compatibility now
3. view.js breaks AMP compatibility
4. Let's remove the script for now, adding it only when needed, and in an AMP-compatible way

jeffersonrabb:
> Outside of Newspack, if a script is needed to accompany this block, whoever is using it should be able to simply hook in appropriately to enqueue their own script. Otherwise, we're expecting them to modify our empty script, which essentially means forking the plugin.

This isn't exactly what I meant here, I'll clarify. This plugin will likely be the container for numerous  blocks created for the Newspack project, not just Homepage Articles. However, the plugin should be installable and useable on sites that aren't running the main Newspack plugin, or AMP. The homepage articles block doesn't require any view-side scripting, but it's conceivable that a future block might need a bit of scripting to provide certain functionality in a non-AMP context. Leaving [this bit](https://github.com/Automattic/newspack-blocks/blob/045f8fc6011fe7175aa7d3272d7963a50955305d/newspack-blocks.php#L101-L109) intact simply means that future blocks _could_ contain view-side scripting and the infrastructure is already there to handle loading them. The change in this PR already accomplishes the goals of fixing AMP compatibility, by excluding the script in AMP requests.

philipjohn (approved):

PR #28, opened 2019-05-30
Add `ignore_sticky_posts` to the output of the homepage blocks

### LINKED ISSUE #27: Homepage Blocks: sticky posts added on top of selected number of posts
Is is working correctly in WordPress query terms, but if you have a sticky post it's getting added on top of the number of posts you pick (so if you set up a block to show three posts, but the category has a sticky post, it will show four). 

This is only happening on the front-end. 

I think it makes the most sense to exclude the sticky posts from the front-end with `ignore_sticky_posts`.

laurelfulford:
### All Submissions:

This PR adds `ignore_sticky_posts` to the query that generates the posts that display on the front end.

This makes sure the posts shown in the editor match the front end, and also make sure the number of posts shown matches what's set in the editor. 

I can see there being a case for the sticky posts behaving like sticky posts in these blocks, but I think overall it will cause more confusion/problems than it will be useful. 

Fixes #27.

1. Mark a post as 'sticky'; preferably one deeper in the post archives.
2. Set up a homepage block with the same category as the sticky post (or all posts). Set a specific number of posts to display.
3. Save and publish.
4. On the front end, note that the sticky post displays first, and the total posts showing is +1 more than you selected.
5. Apply the PR. 
6. Confirm that the correct number of posts are now displaying, and the sticky post isn't on top (unless it's also the latest post). 

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/view.php
+++ b/src/blocks/homepage-articles/view.php
@@ -15,10 +15,11 @@
 function newspack_blocks_render_block_homepage_articles( $attributes ) {
 	$categories    = isset( $attributes['categories'] ) ? $attributes['categories'] : '';
 	$args          = array(
-		'posts_per_page'   => $attributes['postsToShow'],
-		'post_status'      => 'publish',
-		'suppress_filters' => false,
-		'cat'              => $categories,
+		'posts_per_page'      => $attributes['postsToShow'],
+		'post_status'         => 'publish',
+		'suppress_filters'    => false,
+		'cat'                 => $categories,
+		'ignore_sticky_posts' => true,
 	);
 	$article_query = new WP_Query( $args );
 


philipjohn:
Code looks good :+1: Will give this a proper test in a bit

philipjohn:
> On the front end, note that the sticky post displays first, and the total posts showing is +1 more than you selected.

Working off of `master`, I'm not seeing this behaviour. Here's what I did:

1. Added a homepage block with the following settings:

FireShot Capture 087 - Edit Page ‹ newspack test — WordPress - newspack test

2. I'm using the theme test data, so there is a post called "Template:Sticky" that is in "Uncategorized" and is marked sticky.
3. The first post shown on the homepage is "Hello World!"
4. I changed the category to "Template"
5. The first post shown is "Template: Featured Image (Vertical)" not the sticky "Template: Sticky"

There is a high chance I'm `doing_it_wrong()`

laurelfulford:
Ah, strange! I've flipped back to master and confirmed I can still recreate this. Just in case, I changed my settings to match yours:

image

... and see the correct thing in the editor:

image

But when I publish the post and view the site, that same post leads with the sticky post:

image

Can you confirm if you're seeing the right thing on the published post (and not just in the editor)? Thanks!

philipjohn:
Sorry for the slow reply!

> I changed my settings to match yours

It looks like you have "All" as the category rather than "Uncategorized" but that might not matter unless you have the exact same data.

> But when I publish the post 

Just to clarify - the post you're publishing is the front page with the homepage articles block, right?

> Can you confirm if you're seeing the right thing on the published post (and not just in the editor)? Thanks!

Yeah, I'm seeing the same thing in both the editor and on the frontend - no sticky post.

I did a further test. I removed and re-applied the block with the same settings and updated the page. Then I went to the "Template: Comments Disabled" post (also in Uncategorized) and made it sticky. It did not appear in the homepage articles block.

laurelfulford:
Dang, I'm at a bit of a loss here -- I think I might be creating phantom issues? I'm going to test this out on a second fresh test site, to make sure there isn't something site-specific that's messing things up. Thanks for double-checking this!

laurelfulford:
@philipjohn So that 'All'/'Uncategorized' difference seems to be the key. The _only_ case I can recreate this is when I pick 'All'. If I pick any specific categories, though, it doesn't happen. 

Would you be able to take one last look at this with 'All' categories? It's a much smaller use case, I think, but if a site wanted a generic 'Latest News' section that showed everything, it would be an easy way to do it.

philipjohn:
> So that 'All'/'Uncategorized' difference seems to be the key. 

Hmm yeah looks like it! When I choose "All" I do see the sticky post first in the block.

laurelfulford:
Thanks for the multiple reviews here, @philipjohn!

PR #29, opened 2019-05-31
Menu Renderer

jeffersonrabb:
This is a block that simply renders a pre-defined WordPress menu.


jeffersonrabb:
After discussing this with @laurelfulford we may not need this functionality, so putting this on hold until we know more.

jeffersonrabb:
I don't think we're ever going to need this + there's lots of other work going on with superior approaches to block-based menus.

PR #30, opened 2019-05-31
Add field for a block section header

### LINKED ISSUE #23: Homepage blocks: Add title for entire block
It would be helpful to have a 'title' for an entire block that could be manually edited. 

This could often be set to the category that the block presents, or some more eye-catching version of that. It could also be changed to something more general, depending on the settings. 

It would be helpful to have this inside the block, instead of as a separate header block, so the controls affecting the block (like the type scale, or a future option like background colour) would also affect the this title. 

Some examples from cohort sites:

image

image

laurelfulford:
### All Submissions:

This PR adds a basic section heading to the homepage block -- it's an optional string of text that appears above the articles in the block, to help differentiate the different areas on the homepage.

With this PR I've also included some font-related variables, to be used in this block to make it easier to assign throughout; it can be moved to a more central location once we have more blocks to use it with.  

Fixes #23.

1. Apply patch.
2. Add a Homepage block to a page. 
3. Under 'Display Settings', there should now be a field labelled "Block Section Header".
4. Confirm that adding text here appears in the editor.
5. Save and publish and confirm that the text appears on the front end. 
6. Check out some other settings and make sure you don't see any weirdness (a good one is switching the block to a grid). 

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/edit.js
+++ b/src/blocks/homepage-articles/edit.js
@@ -284,7 +284,7 @@ class Edit extends Component {
 				<div className={ classes }>
 					{ sectionHeader && (
 						<div className="article-section-title" key="article-section-title">
-							{ sectionHeader }
+							<span>{ sectionHeader }</span>
 						</div>
 					) }
 
--- a/src/blocks/homepage-articles/edit.js
+++ b/src/blocks/homepage-articles/edit.js
@@ -21,6 +21,7 @@ import {
 	Dashicon,
 	Placeholder,
 	Spinner,
+	TextControl,
 } from '@wordpress/components';
 import { withSelect } from '@wordpress/data';
 import { withState } from '@wordpress/compose';
@@ -89,6 +90,7 @@ class Edit extends Component {
 		const {
 			postsToShow,
 			categories,
+			sectionHeader,
 			columns,
 			showImage,
 			imageScale,
@@ -126,6 +128,12 @@ class Edit extends Component {
 							required
 						/>
 					) }
+					<TextControl
+						label={ __( 'Block Section Header' ) }
+						value={ sectionHeader }
+						onChange={ value => setAttributes( { sectionHeader: value } ) }
+						help={ __( 'Displays an optional header above the article blocks.' ) }
+					/>
 				</PanelBody>
 				<PanelBody title={ __( 'Featured Image Settings' ) }>
 					<PanelRow>
@@ -224,6 +232,7 @@ class Edit extends Component {
 			categories,
 			typeScale,
 			imageScale,
+			sectionHeader,
 		} = attributes;
 
 		const classes = classNames( className, {
@@ -273,6 +282,12 @@ class Edit extends Component {
 		return (
 			<Fragment>
 				<div className={ classes }>
+					{ sectionHeader && (
+						<div className="article-section-title" key="article-section-title">
+							{ sectionHeader }
+						</div>
+					) }
+
 					{ latestPosts && ! latestPosts.length && (
 						<Placeholder>{ __( 'Sorry, no posts were found.' ) }</Placeholder>
 					) }
--- a/src/blocks/homepage-articles/edit.js
+++ b/src/blocks/homepage-articles/edit.js
@@ -277,15 +277,13 @@ class Edit extends Component {
 			<Fragment>
 				<div className={ classes }>
 					{ latestPosts && ( ! RichText.isEmpty( sectionHeader ) || isSelected ) && (
-						<div className="article-section-title">
-							<span>
-								<RichText
-									onChange={ value => setAttributes( { sectionHeader: value } ) }
-									placeholder={ __( 'Write header…' ) }
-									value={ sectionHeader }
-								/>
-							</span>
-						</div>
+						<RichText
+							onChange={ value => setAttributes( { sectionHeader: value } ) }
+							placeholder={ __( 'Write header…' ) }
+							value={ sectionHeader }
+							tagName="div"
+							className="article-section-title"
+						/>
 					) }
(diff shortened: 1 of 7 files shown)


jeffersonrabb:
Note: https://github.com/Automattic/newspack-blocks/pull/25 has been merged into this branch, so the build script is now `build:webpack` instead of just `build`. E.G.:

`npm rum build:webpack -- --watch`

jeffersonrabb (changes requested):
I suggest we handle this a little differently, as a `RichText` component that can be edited directly on the block as opposed to a text input in the sidebar. This is more in line with block design best practices. The Jetpack GIF block has this feature—an optional title that can be input directly on the block—so it might be a good model to look at. Some relevant code:

https://github.com/Automattic/jetpack/blob/master/extensions/blocks/gif/edit.js#L197-L210
https://github.com/Automattic/jetpack/blob/master/extensions/blocks/gif/gif.php#L63-L65

laurelfulford:
Ah I like that idea! I somehow forgot about that component between initially exploring Gutenberg, and working on this block.

I've updated the PR to use RichText component. It's set up to only show when the block is focused on, which I think is more standard for things like captions, but I'm not 100% if it's confusing in this context. Give it a spin and let me know what you think!

src/blocks/homepage-articles/edit.js
```diff
@@ -273,6 +276,17 @@ class Edit extends Component {
 		return (
 			<Fragment>
 				<div className={ classes }>
+					{ latestPosts && ( ! RichText.isEmpty( sectionHeader ) || isSelected ) && (
+						<div className="article-section-title">
+							<span>
```
jeffersonrabb: What's the reasoning for the `<span/>` wrapper here?

laurelfulford: It's 100% for ease of styling :) For some of the styles I've been exploring, having another HTML element in there makes it a lot easier to pull things off. For example, line the darker border to the right edge of the text in something like this:

image

I'm happy to explore another approach if this seems like overkill!

laurelfulford:
Just noting that I'm making one more change to this PR and removing the font families I added -- it makes it harder for the block to inherit styles from the theme, and it seems better to keep things simpler here so we're not fighting the styles.

src/blocks/homepage-articles/edit.js
jeffersonrabb: Here's why I ask. The `RichText` component has an optional `tagName` property that determines what the editable element will be. Using that plus `className`, you could replace all of that markup with just this, _except_ for that problematic `<span/>`:

```
<RichText
	tagName="div"
	className="article-section-title"
	onChange={ value => setAttributes( { sectionHeader: value } ) }
	placeholder={ __( 'Write header…' ) }
	value={ sectionHeader }
/>
```

laurelfulford: Right! Is it okay if we table this to a separate issue, and I explore other ways to handle styles like this? I added the HTML reluctantly, and would be glad to get rid of it, but I'm not sure off the top of my head the best way to approach it and still have the style flexibility.

jeffersonrabb: The only thing is I believe a `<div/>` within a `<span/>` is illegal markup, so it's not just a question of code brevity. Has the flexible styling been implemented somewhere? Part of why this threw me is that I don't see that styling in either of the block's stylesheets currently.

laurelfulford: I'm confused -- it should be a `span` inside of a `div` (which should be okay)? 

I'm working on styles for the theme right now that match the screenshot above -- it wouldn't be the block default (I'd like to keep that as simple as possible, so we don't have to override a bunch of styles).

jeffersonrabb: The `RichText` element renders out as an element itself (actually a tree of multiple elements). `tagName` is the way to determine what the outermost element is. Here's what the markup of what we have now looks like:

Screen Shot 2019-06-04 at 1 47 31 PM

So we have `div->span->div->div->div->div->div`

laurelfulford: Ohhh I see! I was matching the markup I was adding on the front end (which just ends up being `<div><span></span></div>`. Since the span would more likely than not be display inline, it seemed to make sense to use that as an element, rather than a div and setting it to display inline (which doesn't look like it'd work in the editor anyway). 

Is there a way to add multiple HTML elements to something in the editor, or is it always one? I might be able to do something hacky with the other markup there, too, but I'd like to avoid it in case it backs us into a corner later on.

laurelfulford: I discussed this with @jeffersonrabb, and there doesn't seem to be a great way to add this same markup structure in the editor, since there are so many other pieces to it.

For now, we've opted to remove the markup from the editor in 33ec1e0, and keep it on the front-end; I can explore ways to create a similar-ish preview for the above style in the editor. I think it will just need to have the darker grey part of the border always be the same width, but that seems like a good trade-off compared to trying to get elements of the editor markup to still work after being switched to `display: inline`.

src/blocks/homepage-articles/view.scss
```diff
@@ -99,7 +110,10 @@
 	}
 
 	/* Article meta */
+<<<<<<< HEAD
```
jeffersonrabb: Looks like an unresolved merge conflict here.

laurelfulford: Gah, I'm not sure how I missed that one -- fixed in b119a68!

jeffersonrabb (approved):
🚢 . Nice work!

laurelfulford:
Thanks for your help on this one, @jeffersonrabb!

Issue #16, opened 2019-05-15
Homepage Blocks: Mobile styles


laurelfulford:
Related to the work being done in #8:

The Homepage Article blocks still need mobile styles -- this includes:

- Scaling down appropriately depending on screensize
- Breaking down the 'grid' layout into the appropriate number of columns based on screensize

Also, although not directly related to this block, we also need to consider how things may change when layout blocks -- like Core's Columns blocks -- are used. 

This is related somewhat to work to be done in #15 -- optimizing the image sizes loaded in different circumstances.

laurelfulford:
See #40.

dkoo:
@laurelfulford I'm closing this because Homepage Posts definitely has mobile styles by now, but feel free to reopen if you think more work is needed.

PR #31, opened 2019-06-02
Add small screen-specific font sizes

laurelfulford:
This PR copies over some screen size variables and a `@media` mixin from the Newspack theme, and adjusts the font size for the block down depending on screen size. 

This is a partial fix for #16; the blocks still need some work on how the columns break down, to make sure that when they're multi-column, they don't get too crowded, _especially_ when they're nested inside of another block that creates column (like the default Columns block).

## To test

1. Add 8 test blocks to a page -- make each 1 article long, and give them each a different type scale value, ranging from 1 to 8. 
2. View the blocks on a desktop, tablet, and mobile sized screen.
3. Make sure the article title and avatar sizes are scaling up and down, and that nothing feels to overwhelmingly large on the smaller screens.

### DIFF
--- a/src/blocks/homepage-articles/sass/_mixins.scss
+++ b/src/blocks/homepage-articles/sass/_mixins.scss
@@ -1,0 +1,26 @@
+
+@mixin media( $res ) {
+	@if mobile == $res {
+		@media only screen and (min-width: $mobile_width) {
+			@content;
+		}
+	}
+
+	@if tablet == $res {
+		@media only screen and (min-width: $tablet_width) {
+			@content;
+		}
+	}
+
+	@if desktop == $res {
+		@media only screen and (min-width: $desktop_width) {
+			@content;
+		}
+	}
+
+	@if wide == $res {
+		@media only screen and (min-width: $wide_width) {
+			@content;
+		}
+	}
+}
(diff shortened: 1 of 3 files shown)


jeffersonrabb (changes requested):
Looks good - I followed the testing instructions and used Browser Stack to verify on tablet and phone. I had trouble getting avatars to load in Browser Stack but I'm sure that's fully unrelated to this PR. The only issue I spotted is that at type scale 4, the avatar is actually smaller than 3. Here are the avatar sizes by type scale I'm seeing, and you can see that 4 looks like an aberration:

1 - 24x24
2 - 28x28
3 - 32x32
4 - 28x28
5 - 40x40
6 - 44x44
7 - 48x48
8 - 48x48

jeffersonrabb:
Note: https://github.com/Automattic/newspack-blocks/pull/25 has been merged into this branch, so the build script is now `build:webpack` instead of just `build`. E.G.:

`npm rum build:webpack -- --watch`

laurelfulford:
🤦‍♀ Thanks for catching that, @jeffersonrabb! The sizes have been updated so 4 shows the avatar at 36x36, makes a lot more sense in the context of that list.

jeffersonrabb (approved):
🚢 !

laurelfulford:
Thanks @jeffersonrabb!

PR #33, opened 2019-06-05
Tighten up layout of the article blocks

laurelfulford:
### All Submissions:

This PR is a bit of a mixed bag of changes, but they collectively tighten up the layout of the article blocks a bit: 

* Reduces the space between the columns when you've set the article block to display as grid.
* Gets the author avatar, name and published date on one line when there's room for it (the display when there isn't room for all three still needs work; the date sits below the avatar if it's there, which looks strange). 
* Adjusts the vertical space between articles, and with the post meta. 
* Removes an extra `<p>` tag from around `the_excerpt()`, and reworks the post meta markup a bit to allow for the current layout.

1. Set up a couple article blocks - one with multiple items displayed in one column, and second with the grid.
2. Check it over.
3. Apply the PR; confirm a few things:
   * That when there's space, the author avatar, name and published date appear on one line.
   * That the space between columns is reduced in the grid layout.
   * That the vertical space between articles when displayed in one column is increase. 

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/view.scss
+++ b/src/blocks/homepage-articles/view.scss
@@ -6,6 +6,10 @@
 		margin-bottom: 0.75em;
 		word-break: break-word;
 		overflow-wrap: break-word;
+
+		&:not(:last-child) {
+			margin-bottom: 1.5em;
+		}
 	}
 
 	/* Section header */
@@ -32,7 +36,7 @@
 	@include media(tablet) {
 		@for $i from 2 through 6 {
 			&.columns-#{ $i } article {
-				width: calc( ( 100% / #{$i} ) - 24px );
+				width: calc( ( 100% / #{$i} ) - 12px );
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


src/blocks/homepage-articles/view.scss
```diff
@@ -6,6 +6,10 @@
 		margin-bottom: 0.75em;
 		word-break: break-word;
 		overflow-wrap: break-word;
+
+		&:not(:last-child) {
```
jeffersonrabb: [Note] Very, very minor point that shouldn't block anything: this won't quite have the desired effect when using column layout since the `margin-bottom` will be applied to all but the final item in the last row, so if the next-to-last one has greater height we end up with `1.5em` margin at the bottom anyway.

laurelfulford: Oh, good catch -- I totally missed that!

I did rework it a bit so the articles displayed in grid always have `1em` between them, unless you're on smaller screens, when they're stacked.

I also removed the 'last article has less spacing' style -- it doesn't really make sense, since space is needed between it and the next set of articles.

PR #35, opened 2019-06-06
Break Up Homepage Article Block

jeffersonrabb:
### All Submissions:

This PR breaks creates a structure to make multiple blocks that share the structure of the big homepage articles block, without duplicating code. In this approach, there can be numerous blocks that  are basically identical except for name, icon, and which fields are shown/hidden. If needed these blocks could also have unique stylesheets if they needed to diverge from the main one. The big Newspack homepage articles block is replaced by two examples: Hero and List. They are identical except that Hero displays the image and excerpt, while List does not. 

### Other information:


laurelfulford:
I really like this approach, @jeffersonrabb! It seems like a solid way to avoid duplication, and still create different blocks with different settings. 

I'm running into some issues, though:

* The styles aren't previewing for me in the editor -- it looks like it's because the blocks uses the classes with "-hero" and "-list" on the end there and the styles do not; the blocks on the front-end both use the more generic `wp-block-newspack-blocks-homepage-articles`. I think switching the class in the CSS to `[class^="wp-block-newspack-blocks-homepage-articles"]` should fix -- it will apply the styles to any class that starts with `wp-block-newspack-blocks-homepage-articles`. We could also add both classes to all blocks (the more general and more specific). 
* The date isn't showing at all for me in the articles; also, I'm wondering if it should still be toggle-able? I'm not sure publications that publish infrequently will want to show it. 
* The Featured Image panel is empty by default for both block types; it is populated for the Hero image when you switch to left or right-aligned images. Do you think it would be weird to hide the whole panel until the image alignment is switched for that block? Or should we maybe move these out of separate panels with this multi-block approach, since there are less options? 
* For some reason the 'Show Author Avatar' toggle isn't working for me in the editor -- no errors, it just doesn't move, it's fixed to 'on' for the Hero block, and 'off' for the List block.

Just let me know if you have any questions about the above, or if you run into issues recreating!

jeffersonrabb:
Given new thinking, this isn't the route we're going to take... Closing.

Issue #24, opened 2019-05-23
Homepage Blocks: 'View More' link on a per-block basis


laurelfulford:
Many sites include a link from a section of articles on the homepage, to that section's archive.

We could do something similar with the homepage blocks: include an option to have a link to the current category archive.

Out of the box it could have static text ("View More", "View More Articles", "See More"), but we could add an option to edit the text down the road.

One thing to note is that this could become trickier to do with more complex queries (author + tag), since there isn't a default archive page to point to. It's something to look into as we explore this option.

laurelfulford:
The link is added in #37; the link doesn't work in the editor yet; will leave this open until that can get sorted, but it does work on the front-end and can be styled.

jeffersonrabb:
This has been handled in a different way via the dynamic Load More button implemented in https://github.com/Automattic/newspack-blocks/pull/259.

PR #37, opened 2019-06-27
Adding 'more' link option to the Newspack blocks

laurelfulford:
### All Submissions:

This PR adds a 'More' link option to the block. When a block uses a specific category, the 'More' link leads to that Category's archive page. When a block uses 'All' categories, the link will go to the blog posts page. 

Known issues:

* In the editor, the 'More' link just links to `#`, rather than the correct link it uses on the front-end. The primary reason is complexity -- I got blocked trying to find a way to pull the current query's category's URL. I'm happy to circle back to this in a separate ticket, but it seems like it can be moved forward without. 
* The link has no styles included in the block; it uses the generic class `.button`, so it inherits the editor's button styles; on the front-end it successfully picks up the Newspack theme's styles for `.button`. My plan is to update the theme's styles to make sure it looks correct in the editor, and leave it as is for the default, but I'm open to feedback on that! 
* It's currently not possible to edit the button label; I think that will be a valuable enhancement down the road. 

See #24.

1. Apply the PR.
2. Add two Newspack blocks -- one with no specific category setting, and one with a specific category selected.
3. Turn on the 'More' links for both.
4. Verify that the 'More' link displays for each block.
5. Save and publish.
6. Verify that the 'More' link appears for both blocks on the front-end.
7. Click each 'More' link; verify that the one on the block with 'All' categories goes to your blog posts page, and the one on the block with a specific category leads to that category's archive page.

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/edit.js
+++ b/src/blocks/homepage-articles/edit.js
@@ -35,7 +35,15 @@ const MAX_POSTS_COLUMNS = 6;
 class Edit extends Component {
 	renderPost = post => {
 		const { attributes } = this.props;
-		const { showImage, showExcerpt, showAuthor, showAvatar, showDate, sectionHeader } = attributes;
+		const {
+			showImage,
+			showExcerpt,
+			showAuthor,
+			showAvatar,
+			showDate,
+			sectionHeader,
+			moreLink,
+		} = attributes;
 		return (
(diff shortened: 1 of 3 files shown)


laurelfulford:
Thanks @claudiulodro!

PR #39, opened 2019-07-02
Google Ad Manager block

philipjohn:
### All Submissions:

Provides a Gutenberg block for inserting pre-defined ad inventory into the editor. Ads are pulled from the Newspack plugin where the Ads Wizard allows users to insert ad codes from Google Ad Manager.

1. Apply the patch
1. Add some ads using the Google Ad Manager wizard, with the following names and codes:
    1. Name: `Sidebar 300`, Code: `<amp-ad width=300 height=250
    type="doubleclick"
    data-slot="/21819266620/ll_sidebar_med_rect">
</amp-ad>`
    1. Name: `Leaderboard`, Code: `<amp-ad width=468 height=60
    type="doubleclick"
    data-slot="/21819266620/sdfsdf"
    data-multi-size="320x50">
</amp-ad>`
    1. Name: `Large Leaderboard`, Code: `<amp-ad width=728 height=90
    type="doubleclick"
    data-slot="/21819266620/fsadfsdf">
</amp-ad>`
1. Add a new post, and add an Ad Manager block
1. Observe the drop down box in the block settings for selecting an ad unit, which will contain the ad units you added through the wizard
1. Choose the ad units in turn and observe the block dynamically changing size
1. Choose the different block alignment options and verify the ad is placed as expected.
1. Save and view the post to verify the ad code is printed as expected

### Other information:


philipjohn:
@sonjaleix and I talked about how this block works today and decided to try a change.

When we use the Ad Manager block we give users two things:

1. Ad size in block controls
2. Ad unit in inspector controls

The ad size is purely visual, to allow the user to see the ad in the editor in the right size. If it doesn't match the size of the ad unit selected the user may end up confused as the ad unit will render in whatever size it is _supposed_ to be, not the size selected in the block control.

The ad unit code includes width and height parameters, so instead of the block control we are going to try and extract the width/height from the ad code and dynamically alter the size of the block based on the selected ad unit.

sonjaleix:
Great chat today. To add to the above, we've also discussed various approaches of how we can offer a way to support mobile ads – this has been a concern that came up during the charter meeting this week. 

We've discussed the following options we've discovered so far:

- Ad Manager has an option parameter called `data-multi-size`, that allows to pick multiple sizes (e.g. 468x60 and mobile ad banner 320x50). When testing to set up this option on the Ad Manager side, we ran into an error. It seems you can't randomly select two ad sizes, somehow they need to be "compatible" (unclear of the definition here) to each other for this option. 
- We could also borrow from the Gutenberg ads block toggle to control if a slotted ads block will display on mobile or not. This would be a work around to help editors add mobile ads

We'll be testing these options to explore which direction is most feasible and user-friendly.

jeffersonrabb:
@philipjohn could you add instructions to this PR about minimum steps to set up Ad Units in Ad Manager for testing?

philipjohn:
Thanks for the assist, @jeffersonrabb .

I've also updated the PR description with better testing instructions and sample ad codes so that it's super easy to test, as well as adding more thorough testing suggestions.

One outstanding issue: there's an issue with centre alignment. Looks like the AMP plugin might be interfering there, so looking into that. Any thoughts welcome!

philipjohn:
https://github.com/Automattic/newspack-theme/blob/eb2e43ac39fa0c421b997d5776c9aee3618f421b/styles/style-1.css#L3099-L3105

philipjohn:
> One outstanding issue: there's an issue with centre alignment. Looks like the AMP plugin might be interfering there, so looking into that. Any thoughts welcome!

Actually, it seems to be the theme. Specifically, [this code](https://github.com/Automattic/newspack-theme/blob/eb2e43ac39fa0c421b997d5776c9aee3618f421b/styles/style-1.css#L3099-L3105) so with that as a caveat, I think this PR is ready to go.

sonjaleix (commented):
I've tested the ad block. 

Can't tell if you've implemented what we've discussed: adjusting the ad unit display size in the Editor based on the selected ad unit. If you have, this isn't working for me. Front-end size is displaying correctly.

The main issue I'm running into, is that it seems to only render one ad on the page, even though there are multiple ad blocks set up on a post (with different ad units for each selected). _I'm seeing some AMP conflicts in my local instance, might have something to do with it?!_

A small odd bug I ran into with the ad unit name field: 
To replicate, create an ad unit and call it "Ad 300x250". Once you've saved the ad unit and you land back on the ad units overview screen you'll see the ad unit name display as "Ad 300&#215;250". See screenshot: 
Screenshot 2019-07-08 14 05 42

src/blocks/google-ad-manager/edit.js
```diff
@@ -0,0 +1,74 @@
...
+	render() {
+		/**
+		 * Constants
+		 */
+		const { attributes, setAttributes } = this.props;
+		const { activeAd } = attributes;
+		const { adUnits } = this.state;
+		const activeAdData = adUnits.find( adUnit => parseInt( adUnit.id ) === parseInt( activeAd ) );
+		const style = activeAdData && {
+			width: `${ activeAdData.width }px`,
```
jeffersonrabb: Following [this PR closure](https://github.com/Automattic/newspack-plugin/pull/167#issuecomment-509213072) I don't see any dimensions extraction regex in here? 

Here's a branch with a stab at one: https://github.com/Automattic/newspack-blocks/compare/add/gam-block...add/gam-block+regex

src/blocks/google-ad-manager/edit.js
```diff
@@ -0,0 +1,74 @@
+/**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { Component, Fragment } from '@wordpress/element';
+import { InspectorControls } from '@wordpress/editor';
+import { SelectControl, Placeholder } from '@wordpress/components';
+import { withState } from '@wordpress/compose';
```
jeffersonrabb: This can be removed.

src/blocks/google-ad-manager/view.php
```diff
@@ -0,0 +1,44 @@
...
+
+	$active_ad = isset( $attributes['activeAd'] ) ? (int) $attributes['activeAd'] : 0;
+	if ( 1 > $active_ad ) {
+		return '';
+	}
+
+	$ad_code = get_post_meta( $active_ad, 'newspack_ad_code', true );
+
+	Newspack_Blocks::enqueue_view_assets( 'google-ad-manager' );
+	return $ad_code;
```
jeffersonrabb: This needs to be wrapped in a div with classes identifying the block and alignment. This is probably why @sonjaleix  is only seeing one ad per page, and why block alignment isn't working. Here's a suggested approach:

```suggestion
	return sprintf(
		'<div class="%s">%s</div>',
		esc_attr( Newspack_Blocks::block_classes( 'google-ad-manager', $attributes )),
		$ad_code
	);
```

src/blocks/google-ad-manager/editor.scss
```diff
@@ -0,0 +1,31 @@
...
+	.components-toggle-control__label {
+		line-height: 1.4em;
+	}
+
+	.components-base-control__field {
+		padding: 7px;
+	}
+}
+
+.newspack-gam-ad-leaderboard .components-placeholder {
```
jeffersonrabb: Neither of these are used anymore.

src/blocks/google-ad-manager/editor.scss
```diff
@@ -0,0 +1,31 @@
+[data-type='newspack-blocks/google-ad-manager'][data-align='center'] .newspack-gam-ad {
```
jeffersonrabb: This file can probably be much simpler. I was able to get more or less the same results with only this:

```
.newspack-gam-ad .components-placeholder {
	display: block;
	height: 100%;
	display: flex;
	min-height: 31px;
}
```

src/blocks/google-ad-manager/view.scss
```diff
@@ -0,0 +1,288 @@
+@import "sass/variables";
```
jeffersonrabb: This file looks like it was copied directly from the `homepage-articles` block. This block doesn't seem to need any view-side styling, so this can be deleted completely. You can also delete `src/blocks/google-ad-manager/view.js`. And in `src/blocks/google-ad-manager/index.js` you can remove line 16: `import './view.scss';`

src/blocks/google-ad-manager/sass/_mixins.scss
```diff
@@ -0,0 +1,26 @@
+
```
jeffersonrabb: Unused. Remove pls.

src/blocks/google-ad-manager/sass/_variables.scss
```diff
@@ -0,0 +1,22 @@
+$mobile_width: 600px;
```
jeffersonrabb: Unused. Remove pls.

src/blocks/google-ad-manager/view.js
```diff
@@ -0,0 +1,7 @@
+//alert( "hello!");
```
jeffersonrabb: This can be removed.

src/blocks/google-ad-manager/index.js
```diff
@@ -0,0 +1,46 @@
...
+ * Internal dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import edit from './edit';
+
+/**
+ * Style dependencies - will load in editor
+ */
+import './editor.scss';
+import './view.scss';
```
jeffersonrabb: This line should be removed.

jeffersonrabb (changes requested):
Most of this PR is unused and can be deleted. There doesn't appear to be any width/height regex in this PR — I pushed a branch with a proposed approach to this (https://github.com/Automattic/newspack-blocks/compare/add/gam-block...add/gam-block+regex). Added a note about how to wrap the front-end rendered content, and a few other notes.

src/blocks/google-ad-manager/view.php
philipjohn: So, I'd already done this and hadn't committed and pushed :facepalm:

philipjohn:
> Can't tell if you've implemented what we've discussed: adjusting the ad unit display size in the Editor based on the selected ad unit. If you have, this isn't working for me. Front-end size is displaying correctly.

Sorry, Sonja, I think this is my fault - I was still running it locally using the API to send the sizes through which you would have been missing.

> The main issue I'm running into, is that it seems to only render one ad on the page, even though there are multiple ad blocks set up on a post (with different ad units for each selected). I'm seeing some AMP conflicts in my local instance, might have something to do with it?!

Hmm, I saw this too, once - it flashed up then disappeared. A reload of the page shows both ad units fine now though.

> To replicate, create an ad unit and call it "Ad 300x250". Once you've saved the ad unit and you land back on the ad units overview screen you'll see the ad unit name display as "Ad 300×250". See screenshot:

I'm gonna drop that into a new issue (#51) as I think it might be a bit more involved.

src/blocks/google-ad-manager/editor.scss
```diff
@@ -0,0 +1,6 @@
+.newspack-gam-ad .components-placeholder {
```
jeffersonrabb: What would you think about centering the `Placeholder` in the block? It looks a little funny when you have a small, square unit left justified within  full-width block.

src/blocks/google-ad-manager/index.js
```diff
@@ -0,0 +1,45 @@
...
+		<path d="M11.8 10.9c-2.27-.59-3-1.2-3-2.15 0-1.09 1.01-1.85 2.7-1.85 1.78 0 2.44.85 2.5 2.1h2.21c-.07-1.72-1.12-3.3-3.21-3.81V3h-3v2.16c-1.94.42-3.5 1.68-3.5 3.61 0 2.31 1.91 3.46 4.7 4.13 2.5.6 3 1.48 3 2.41 0 .69-.49 1.79-2.7 1.79-2.06 0-2.87-.92-2.98-2.1h-2.2c.12 2.19 1.76 3.42 3.68 3.83V21h3v-2.15c1.95-.37 3.5-1.5 3.5-3.55 0-2.84-2.43-3.81-4.7-4.4z"/>
+		<path d="M0 0h24v24H0z" fill="none"/>
+	</svg>
+);
+
+export const settings = {
+	title,
+	icon,
+	category: 'newspack',
+	keywords: [ __( 'ad' ), __( 'advert' ), __( 'ads' ), __( 'Google' ) ],
```
jeffersonrabb: I believe there's a three keyword limit.

src/blocks/google-ad-manager/edit.js
```diff
@@ -0,0 +1,87 @@
...
+		const style = {
+			width: `${ width }px`,
+			height: `${ height }px`,
+		};
+		const dimensions = width && height ? ` (${ width } x ${ height })` : '';
+		return (
+			<Fragment>
+				<InspectorControls>
+					<SelectControl
+						label={ __( 'Advert' ) }
```
jeffersonrabb: For consistency's sake, could we use `Ad unit` here and in the placeholder?

src/blocks/google-ad-manager/edit.js
```diff
@@ -0,0 +1,87 @@
...
+						value={ activeAd }
+						options={ this.adUnitsForSelect( adUnits ) }
+						onChange={ activeAd => setAttributes( { activeAd } ) }
+					/>
+				</InspectorControls>
+				<div className="wp-block-newspack-blocks-google-ad-manager">
+					<div className="newspack-gam-ad" style={ style }>
```
jeffersonrabb: It would be nice to change the label if no ad has been selected. Here's a proposed approach:

```
<div className="wp-block-newspack-blocks-google-ad-manager">
					{ !! activeAdData && (
						<div className="newspack-gam-ad" style={ style }>
							<Placeholder icon={ icon } label={ __( 'Ad Unit' ) + dimensions } />
						</div>
					) }
					{ ! activeAdData && (
						<div className="newspack-gam-ad">
							<Placeholder icon={ icon } label={ __( 'Select an Ad Unit' ) } />
						</div>
					) }
				</div>
```

src/blocks/google-ad-manager/index.js
```diff
@@ -0,0 +1,45 @@
...
+ * Style dependencies - will load in editor
+ */
+import './editor.scss';
+
+export const name = 'google-ad-manager';
+export const title = __( 'Google Ad Manager' );
+
+/* From https://material.io/tools/icons */
+export const icon = (
+	<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24">
```
jeffersonrabb: For the icon we should use the `wordpress/components` SVG components: `SVG` and `Path` (they're already imported so all you need to do is replace the tags).

jeffersonrabb (changes requested):
Left a few more comments, all pretty minor.

For the bad HTML characters @sonjaleix is seeing when there's an `x` in the Ad Unit name, let's not worry about that in this PR. We'll have to dig in, but it's probably a `newspack-plugin` issue, since the same formatting problem is seen in the Wizard.

When I test with AMP units I see what looks like a placeholder with the word "Ad". When I test with non-AMP units, I haven't seen anything. Ideally I'd like to see at least one real ad render in each environment before approving this.

philipjohn:
> When I test with AMP units I see what looks like a placeholder with the word "Ad". When I test with non-AMP units, I haven't seen anything.

Yep, there's definitely an assumption of AMP here, sorry that isn't clear. However, the improvements I've discussed with Sonja this week would change that, by providing more flexibility and adding the ability to add header code (which is needed for non-AMP ads).

src/blocks/google-ad-manager/view.php
```diff
@@ -19,10 +19,20 @@ function newspack_blocks_render_block_google_ad_manager( $attributes ) {
 		return '';
 	}
 
-	$ad_code = get_post_meta( $active_ad, 'newspack_ad_code', true );
+	$classes = Newspack_Blocks::block_classes( 'newspack-blocks-google-ad-manager', $attributes );
+	if ( isset( $attributes['className'] ) ) {
```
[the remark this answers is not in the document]
philipjohn: Oh nice

src/blocks/google-ad-manager/index.js
philipjohn: Ugh, so there is

src/blocks/google-ad-manager/index.js
philipjohn: TIL!

jeffersonrabb:
This functionality has been extracted to a feature plugin. Closing in favor of https://github.com/Automattic/newspack-plugin/pull/184 + https://github.com/Automattic/newspack-gam.

PR #58, opened 2019-07-22
Update font sizing to use ems

laurelfulford:
### All Submissions:

This PR updates the homepage block's font sizing to switch it from pixels to ems, so it inherits the general font sizing from the theme using it.

It also updates the font size variables to match sizing used in the theme; they're not used in the homepage article blocks, but I left them in place in case we need them for another block. If it turns out we don't, and we stick with using ems, they can be removed. 

1. Set up the homepage block using the different sizes, or copy-paste this example into the code view of the block editor: https://cloudup.com/cKNYeV2Gil3
2. Apply the PR and run `npm run build`.
3. Confirm that the font sizing hasn't really changed.

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/sass/_variables.scss
+++ b/src/blocks/homepage-articles/sass/_variables.scss
@@ -6,15 +6,18 @@ $wide_width: 1379px;
 $font__size_base: 20px;
 $font__size-ratio: 1.125;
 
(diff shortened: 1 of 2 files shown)


laurelfulford:
Thanks for the review!

Issue #12, opened 2019-05-15
Duplicate post management


laurelfulford:
This came up in the review for #8 from @jeffersonrabb.

We need a way to make sure posts are NOT duplicated across multiple sections when there is overlap in the homepage blocks.

This could be manually avoided by very careful section selection, or by having some kind of offset control, but having a programmatic way to handle it would be the least fragile.

We would also need to consider how this interacts with things like 'Most Popular' and 'Most Recent' post sets, or other sections that may need to be exempt.

jeffersonrabb:
Resolved in https://github.com/Automattic/newspack-blocks/pull/289.

PR #61, opened 2019-07-22
Avoid Post Duplication (View Only)

jeffersonrabb:
### All Submissions:

This PR is an attempt to ensure that multiple articles blocks on the same page will not show the same post more than once. Some questions and caveats:

- This PR only affects the view (preview/published content). In the editor there will still be duplicate posts. This will be addressed in a separate PR.
- Currently this logic will be applied to every instance of `Newspack Homepage Articles` block. Would we need the flexibility to exempt certain blocks from the logic? 
- Blocks will be processed in the order they appear on the page, with higher prioritized stories going to the first ones. It is possible that an unusual layout could be produced where the topmost block is not the visually dominant one, which could lead to unexpected distribution of posts.

This is a partial resolution of https://github.com/Automattic/newspack-blocks/issues/12

1. Add a few instances of the block to a post or page, with identical query settings.
2. Observe there are duplicate posts visible in the Editor.
3. Publish or preview. Observe that there is no duplication.

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/view.php
+++ b/src/blocks/homepage-articles/view.php
@@ -13,10 +13,15 @@
  * @return string Returns the post content with latest posts added.
  */
 function newspack_blocks_render_block_homepage_articles( $attributes ) {
+	global $newspack_blocks_post_id;
+	if ( ! $newspack_blocks_post_id ) {
+		$newspack_blocks_post_id = array();
+	}
 	$author        = isset( $attributes['author'] ) ? $attributes['author'] : '';
 	$categories    = isset( $attributes['categories'] ) ? $attributes['categories'] : '';
+	$posts_to_show = intval( $attributes['postsToShow'] );
 	$args          = array(
-		'posts_per_page'      => $attributes['postsToShow'],
+		'posts_per_page'      => $posts_to_show + count( $newspack_blocks_post_id ),
 		'post_status'         => 'publish',
 		'suppress_filters'    => false,
 		'cat'                 => $categories,
@@ -45,6 +50,7 @@ function newspack_blocks_render_block_homepage_articles( $attributes ) {
 	if ( isset( $attributes['className'] ) ) {
 		$classes .= ' ' . $attributes['className'];
 	}
+	$post_counter = 0;
 
 	ob_start();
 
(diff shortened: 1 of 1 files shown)


jeffersonrabb:
> I looked into using offset instead of oversized posts_per_page (for much better performance), but I wasn't able to figure out how to make it work nicely when there are homepage blocks with categories. 

I tried this approach too, but as you say it only really works if all blocks use the exact same query, which they won't. Do you think the oversized `posts_per_page` will be a significant drag? I was under the impression this won't be too costly, and the most important thing was to avoid using `post__not_in`, which is the primary rationale for the approach used here.

claudiulodro:
Good sleuthing. I think the issue may be as simple as [this being an off-by-one error](https://github.com/Automattic/newspack-blocks/pull/61/files#diff-0bc1373b8d23b0320f8a67d468c37a1eR67)?

e.g. 
```
$posts_to_show = intval( $attributes['postsToShow'] ); // Let's say postsToShow = 1
$post_counter = 0

// Show posts until $post_counter > $posts_to_show
// Display a post. $post_counter is now 1. Go again since $post_counter > $posts (1 > 1) is false.
// Display a post. $post_counter is now 2. $post_counter > $posts, so stop showing posts. 2 posts were displayed and we requested only 1.
```

jeffersonrabb:
Nice sleuthing both of you! This should fix it: https://github.com/Automattic/newspack-blocks/pull/61/commits/365d8d1673a4b741145cc99d0b53be5bcfd91538

laurelfulford:
Looks good to me! I turned my front page into a dumpster fire of multi-category posts; there aren't any dupes, and the post counts are right; parsing through the /wp-admin posts screen, it looks like everything is displaying in order and where expected. 🎉

jeffersonrabb:
OK! Onward to solving the issue in the editor now...

iamtakashi:
> Onward to solving the issue in the editor now...

Could you let me know where this work is happening? The difference between the editor and the front-end of the site is very confusing at the moment. cc @Automattic/ajax

jeffersonrabb:
> Could you let me know where this work is happening?

This editor de-duplication work is occurring in https://github.com/Automattic/newspack-blocks/pull/176 which is a  different block. It's conceivable the work in that block could be applied to Homepage Articles, but this isn't being actively explored at the moment. This is the commit where editor de-dupe was first introduced: https://github.com/Automattic/newspack-blocks/pull/176/commits/672ff1903cde3e5f68af2a955eafc5a87e93d57c

cc: @georgeh

Issue #41, opened 2019-07-02
Homepage Blocks: Building queries


laurelfulford:
Right now, the homepage blocks will display all the blocks, or you can filter by a specific category.

We should also explore: 

1. Filtering by other options (author, tag) (pretty sure we've already talked about this one)
2. Ability to pick a specific story (you could stack multiple blocks to create a section of specific stories).
3. ~~Has not been requested yet, but multiple filters (2-3 categories if it makes sense, or author + category).~~ On second thought, I think this could be needlessly complicated (in options you'd have to pick from, and the resulting query), and have limited applications you couldn't recreate by adding an additional category or tag for sorting. I think it makes sense to hold off until there are requests/real use cases we can't tackle in another way.

PR #63, opened 2019-07-25
Specify Single Post

jeffersonrabb:
### All Submissions:

Adds a Select to the top of the **Display Settings** Panel for selecting a specific post. If a selection is made, only this post will be shown, and any category, author, or limit settings will be ignored. In the sidebar, the category/author/limit will actually be hidden if a post selection has been made to avoid confusion. 

Big caveat! A select control is definitely not the right UI for this because of the potentially huge number of posts. A select is used in the customizer for setting the homepage Page so I opted to go this way at first, but it will need to be replaced. The Select will only display the 20 most recent posts, so the current version is not yet as flexible as it needs to be. I propose we consider this PR as-is so that the basic functionality is addressed, then consider alternate UI in a new PR (cc: @sonjaleix).

Screen Shot 2019-07-25 at 8 35 04 AM

This partially resolved https://github.com/Automattic/newspack-blocks/issues/41

1. Check out this branch, run `npm run build:webpack`
2. Add a block. Observe **Display One Specific Post** `SelectControl` at the top of the sidebar. 
3. Select a post. 
4. Observe that only that post is displayed in the block. 
5. Observe that **Category**, **Author**, and **Number of Items** UI are hidden.
6. Preview the post and verify that only the select post is shown.
7. In the editor, set the `SelectControl` back to `-- Select Post --` and observe that all UI fields are revealed again and the posts shown return to normal.

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/edit.js
+++ b/src/blocks/homepage-articles/edit.js
@@ -17,11 +17,11 @@ import { __ } from '@wordpress/i18n';
 import { Component, Fragment, RawHTML } from '@wordpress/element';
 import { InspectorControls, RichText, BlockControls } from '@wordpress/editor';
 import {
-	ToggleControl,
 	PanelBody,
 	PanelRow,
 	RangeControl,
 	Toolbar,
+	ToggleControl,
 	Dashicon,
 	Placeholder,
 	Spinner,
@@ -124,6 +124,7 @@ class Edit extends Component {
 			postLayout,
 			mediaPosition,
 			moreLink,
+			singleMode,
 		} = attributes;
 		return (
 			<Fragment>
@@ -134,6 +135,7 @@ class Edit extends Component {
 							onNumberOfItemsChange={ value => setAttributes( { postsToShow: value } ) }
 							authorList={ authorList }
 							postList={ postList }
+							singleMode={ singleMode }
 							categoriesList={ categoriesList }
 							selectedCategoryId={ categories }
 							selectedAuthorId={ author }
@@ -147,6 +149,7 @@ class Edit extends Component {
 							onSingleChange={ value =>
 								setAttributes( { single: '' !== value ? value : undefined } )
 							}
+							onSingleModeChange={ value => setAttributes( { singleMode: value } ) }
 						/>
 					) }
 					{ postLayout === 'grid' && (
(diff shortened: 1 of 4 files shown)


sonjaleix:
Just some initial feedback on the Single Post selector UI. It's not ideal to show both options (select single post, select category etc.) next to each other. It feels confusing. 

I propose one of two solutions: 
1. The UI pattern should have a top-level selector between "individual post" and "all posts" (probably need to find a better label). By default I would propose to select "all posts" and show the category/author/ etc. options. The user can then switch to "individual post" easily and select the post. If we can reuse something like the auto-complete feature from the UI of adding a text link to a post in the Editor.
Pros: all options are in one block and clearly separated out
Cons: Adds more layers of UI, there's a risk if not implemented well for this option to be missed by the user

2. Create a separate block
Pros: simplified UI, we could make it a related block so the user can switch between that and the regular post block
Cons: Probably not different enough of a block to separate it out

Additionally the "Read more" option should probably also be hidden when "Single Post" is selected

I agree with the approach to ship this and then iterate.

jeffersonrabb:
One possibility would be to add the Toggle here, and address the ideal UI for post selection separately. The Toggle is the easy part. @sonjaleix, @laurelfulford do you think we should do this? Or hold off and address all UI optimization in a single, separate PR?

sonjaleix:
@jeffersonrabb How do you envision the toggle to work?

jeffersonrabb:
> How do you envision the toggle to work?

In the simplest form it would look like this. Switching it on hides all of the Category, Number of Items, Author, and Show More Link, and reveals the Post dropdown. Switching it off does the inverse.

Screen Shot 2019-07-25 at 2 37 41 PM

sonjaleix:
This might work. Will have to test in action to make a proper ux assessment. I'd vote to implement it this way for now.

jeffersonrabb:
> This might work. Will have to test in action to make a proper ux assessment. I'd vote to implement it this way for now.

Added the toggle in https://github.com/Automattic/newspack-blocks/pull/63/commits/7f9e4c069e746b1436bb96e63884a599b8af452e. See what you think?

sonjaleix:
Thanks for creating this. On initial review, this might work. 

One thing we'll need to fix however: when a user first turns "choose specific story" on and no story is selected in the dropdown, the block shouldn't still show the content from the category selection. We might want to add a note for the user to pick a story.

jeffersonrabb:
There's definitely improvements to be made here, but the team needs this feature quickly so I'm calling this good enough for now and 🚢 ing

PR #65, opened 2019-07-31
First pass at adding a 'post meta' (author and date) block

laurelfulford:
### All Submissions:

This PR creates a first pass at a block that displays the post's author and publish date. It's _very_ rough -- there aren't really editor styles -- but it's a start.

This is what the block currently outputs in the editor: 

image

... and on the front-end: 

image

(It currently uses the same markup as the theme, so it's inheriting the styles on the front-end).

One thing I was wondering was should we put _everything_ together -- author, date and share icons -- or create individual blocks for each? 

I can see benefits to both -- if we keep them separate, it will be more flexible for layout and choosing what's shown. However, it could make it harder to achieve layouts like the one currently in the theme:

image

You could use a column block and align the text to the right on the column with the share icons... but that's not going to look great on mobile since the share icons will still be right aligned.

Bundled together, it limits the layout/options, but we might be able to offer some pre-set choices, similar to the columns:

image

We also need to consider how we get something like this working with multi-author?

I also expect this block needs a new name; what it is will depend on our approach.

There might also be other things I'm not considering -- any thoughts are welcome!

1. Add the 'Post Meta' block to a post.
2. Confirm that the correct author and published date are displaying relative to the current post. 

### Other information:

### DIFF
--- a/class-newspack-blocks-api.php
+++ b/class-newspack-blocks-api.php
@@ -35,6 +35,17 @@ class Newspack_Blocks_API {
 			)
 		);
 
+		/* Add published date source. */
+		register_rest_field(
+			'post',
+			'newspack_published_date',
+			array(
+				'get_callback'    => array( 'Newspack_Blocks_API', 'newspack_blocks_get_published_date' ),
+				'update_callback' => null,
+				'schema'          => null,
+			)
+		);
+
 		/* Add first category source */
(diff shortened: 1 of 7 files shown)


laurelfulford:
Closing this in favour of other work happening in the block.

Issue #43, opened 2019-07-02
Homepage blocks: Block styles


laurelfulford:
Dig into adding some block styles to the blocks -- it needs some more exploration, but my current ideas are:

* Borders in-between (vertical on single-column blocks, or horizontal when its displayed as a grid)
* Backgrounds on a per-article basis (backgrounds for the entire block can be achieved by nesting inside of the Group block, so I think it's safe to leave out). 

Next steps: 

Further exploration into things it makes sense to add.

laurelfulford:
Another option for the styles could be to match the articles heights in a section, and pin the post meta to the bottom, to create a more consistent grid. This probably should be paired with a background colour, or at least a border.

laurelfulford:
Another one for the list: featured images as article backgrounds. Perhaps this option could also allow you to pick a solid colour as a background, as a fallback for articles that don't have images.

laurelfulford:
There's also full-width blocks -- [The Intercept](https://theintercept.com/) pairs this with a background image, for a bold featured story:

![image](https://user-images.githubusercontent.com/177561/60904045-3e0d6400-a227-11e9-9aaf-8824a7541efa.png)

laurelfulford:
I think this one is safe to close for now -- both the borders and background image options have been added, and we can create issues for any specific styles that may be suggested down the road.

PR #73, opened 2019-08-19
Adds first pass at a 'background image' style for the block.

laurelfulford:
### All Submissions:

This PR adds a 'background image' style to the block; if a story has a featured image, it will be used as a background, and the text inside of the block will be set to white. 

I'm curious what other options could be helpful for this block. I'm thinking height/padding, or vertical alignment could help with placing the text at the top/bottom, or left/right; there's probably more.

Related to #43

2. Add a newspack block; make sure at least some of the articles in it have featured images.
3. In the style switcher, pick 'Background Image':

image

4. Confirm that the featured image is set as the background, and the text is overlaid and white:

image

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/edit.js
+++ b/src/blocks/homepage-articles/edit.js
@@ -165,7 +165,6 @@ class Edit extends Component {
 			showImage,
 			showCaption,
 			imageScale,
-			topPadding,
 			showExcerpt,
 			typeScale,
 			showDate,
@@ -248,17 +247,6 @@ class Edit extends Component {
 							max={ 4 }
 							beforeIcon="images-alt2"
 							afterIcon="images-alt2"
-							required
-						/>
-					) }
-					{ showImage && mediaPosition === 'behind' && (
-						<RangeControl
-							className="top-padding-slider"
-							label={ __( 'Top padding' ) }
-							value={ topPadding }
-							onChange={ value => setAttributes( { topPadding: value } ) }
-							min={ 1 }
-							max={ 12 }
 							required
 						/>
 					) }
@@ -358,7 +346,6 @@ class Edit extends Component {
 			categories,
 			typeScale,
 			imageScale,
-			topPadding,
 			sectionHeader,
 			showCaption,
 		} = attributes;
@@ -373,7 +360,6 @@ class Edit extends Component {
 			[ `image-shape${ imageShape }` ]: imageShape !== 'landscape',
 			'has-text-color': textColor,
 			'show-caption': showCaption,
-			[ `top-padding${ topPadding }` ]: mediaPosition === 'behind' && showImage,
 		} );
 
 		const blockControls = [
--- a/src/blocks/homepage-articles/edit.js
+++ b/src/blocks/homepage-articles/edit.js
@@ -395,6 +395,12 @@ class Edit extends Component {
 				isActive: mediaPosition === 'right',
 				onClick: () => setAttributes( { mediaPosition: 'right' } ),
 			},
+			{
+				icon: 'admin-page',
+				title: __( 'Show media behind' ),
+				isActive: mediaPosition === 'behind',
+				onClick: () => setAttributes( { mediaPosition: 'behind' } ),
+			},
 		];
 
 		const blockControlsImageShape = [
--- a/src/blocks/homepage-articles/edit.js
+++ b/src/blocks/homepage-articles/edit.js
@@ -165,6 +165,7 @@ class Edit extends Component {
 			showImage,
 			showCaption,
 			imageScale,
+			topPadding,
 			showExcerpt,
 			typeScale,
 			showDate,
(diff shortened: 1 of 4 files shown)


rickybanister:
Beautiful!

I agree that the text alignment options would add more code-free flexibility, but this is already going to be super useful as is.

laurelfulford:
Thanks @rickybanister! 

I'll set this to 'needs review' to get some eyes on the code, and if it goes in as is we can iterate on it.

jeffersonrabb (commented):
If Newspack Theme is current and custom colors are in use, the author name turns dark (this is not actually one of the theme colors). Also, in non-columnar layouts perhaps we need a `min-height`?

Screen Shot 2019-08-19 at 11 50 54 PM

rickybanister:
I don't know if we need a min-height, that could be just as frustrating in some other contexts as not having one here. Maybe we provide a height slider sidebar option? And maybe eventually we could provide a handle on the image itself to pull the block taller like the proposed column and image adjustment UIs? 

Regarding the author style, do we just need to add more specificity to override the custom color?

laurelfulford:
> Also, in non-columnar layouts perhaps we need a min-height?
> Maybe we provide a height slider sidebar option? 

I agree we need some way to control the height, and probably a base-line min height -- right now, assuming it's not displaying as a grid, the height is determined by the amount of text inside. 

I think it would also be good to be able to align things left, right, up, down, but that's kind of getting ahead of things :) 

One challenge is that, as far as I know, you can't conditionally show/hide controls based on block style (like you can with other toggles -- like showing/hiding the featured image-related controls based on whether the images are actually displaying). 

So we may always have to show the height controls, or perhaps have to move this very-style-like control to a non-style placement (like as one of the image alignment options, or to the sidebar). 

@jeffersonrabb & @rickybanister - I'd be really interested in your thoughts on the above!

rickybanister:
The image alignment options makes sense to me actually—we could provide options for 'crop to text' or 'full size image' and show the image in its native aspect ratio?

laurelfulford:
I've updated the block so that the 'background image' option is one of the image alignment options -- I also realized that it, as a style, would clash with the left and right options, so it really does make sense to have it there instead.

For the time being, I used a pretty weak icon (below) but it works for now as a placeholder:

image

I also added some padding to the top, and aligned the text to the bottom, but I think we really do need some kind of control to adjust depending on how large the block is -- a v2? 

> we could provide options for 'crop to text' or 'full size image' and show the image in its native aspect ratio?

I think it should be doable, but that might need to be a v3, 4 and 5 :)

rickybanister:
That all sounds right to me. Maybe we should just use the image icon to represent this style since it's image-dominant?

laurelfulford:
> Maybe we should just use the image icon to represent this style since it's image-dominant?

Oh, I like that better! I updated it in 7c1af2c:

image

Right now I'm just using the dashicons that are immediately available; we could also replace it with a more appropriate SVG once we track one down :)

laurelfulford:
I made an update to this PR in 4e6eb9a, adding the top-padding slider, since it seemed important for this style to actually be usable. 

I ranges from 1 to 12, and each value has a padding based on screen size, following the same pattern:

* For `4`, it's `2em` on mobile, `4em` on tablet, and `8em` on desktop.
* For `6`, it's `3em` on mobile, `6em` on tablet, and `12em` on desktop.
* For `10`, it's `5em` on mobile, `10em` on tablet, and `20em` on desktop.

... etc.

jeffersonrabb (changes requested):
I'm getting a white bar (white text on a white background) for the title when testing this.

Screen Shot 2019-10-08 at 12 29 27 AM

laurelfulford:
Ah, foiled by Style 3 again! 

It looks like I'll need to add a complementary PR for that style pack to make sure this doesn't happen (similar to how it's being handled when the article block is nested inside of a cover block). 

I'll circle back shortly with a PR for the theme to test with this, and make sure it's resolved.

laurelfulford:
There's an update for style 3 that can be tested with this PR here:

laurelfulford:
In testing the above, I also noticed there's an issue with `object-fit: cover` + AMP.

(This also happens for two of the theme featured image styles, the behind and beside options.)

I found this fix for an object fit issue in Twenty Nineteen that we might want to explore: https://github.com/ampproject/amp-wp/pull/1619/commits/4fb548aa8883a9fd80c7849fb74e51cf5edb8366

laurelfulford:
Ok, the above is fixed in 4e211a9  -- thanks for the big assist, @jeffersonrabb!

laurelfulford:
@jeffersonrabb as per our convo in Slack, I've removed the padding control in 560d5db for now, so we can handle it in a separate PR and explore some other approaches before sticking with this one.

One other thing that's changed since this PR was created: we can now show the image caption in each block :| That could be something else to mull over approaches for after this PR is merged. I'm leaning towards hiding it and the option to show it entirely when this image position is picked, but there might be a better approach!

jeffersonrabb (approved):
This is good to go. Thanks for your patience on this one!

laurelfulford:
Thanks Jeff!

> Thanks for your patience on this one!

Not a problem! I appreciate that you have reserve to step back and rethink approaches like this -- I get so swept up in solving the problem sometimes, I run with the first solution I see without considering the ways we could end up "stuck" with it. :) 

I'll create a separate issue in which we can hash the scaling question and the captions questions out!

PR #85, opened 2019-08-22
Add a 'border' style for the columns block

### LINKED ISSUE #81: Columns Block: 'borders' styles
Add a style for the columns block that will place borders in-between the columns.

laurelfulford:
### All Submissions:

This PR adds a 'border' style to the column block.

It also removes an unneeded style from the re-ordering styles (`flex-wrap: nowrap`).

This PR does need a couple adjustments to the theme, as far as when the columns break -- the theme overrides the Gutenberg default for some reason, so you get some odd column spacing as you shrink the browser window down. I've added a fix for this in https://github.com/Automattic/newspack-theme/pull/314; that PR will need to be applied to the theme to get the same results as some of the screenshots below.

I realize this is a pretty odd approach -- I couldn't find a good way to use `border-right` and maintain the even column spacing, and I wanted to mess with the main styles of the block (like left and right margins) as little as possible. I'm open to suggestions for other approaches, though!

Closes #81

2. Copy-paste [this test content into the code editor](https://cloudup.com/cIOddZlCYVb); to best test the spacing of each column, it has an image in each so you can see exactly where they are.
3. Switch the column style to use the 'Borders' style:

image

4. Confirm that a border appears between each block, evenly spaced:

image

5. Shrink down the window and confirm that when the columns start displaying in two columns, the border does not appear next to the end of each row (PR https://github.com/Automattic/newspack-theme/pull/314 needs to be applied to see this; otherwise the second and third block will be on the same line, with no space between):

image

6. Shrink down the browser window further and confirm that the border appears under each column once they're down to one column, excluding the last one:

image

### Other information:

### DIFF
--- a/src/block-styles/core/columns/editor.scss
+++ b/src/block-styles/core/columns/editor.scss
@@ -20,4 +20,42 @@
 			order: -1;
 		}
 	}
+
+	&.is-style-borders {
+		[data-type="core/column"] {
+			border-bottom: 1px solid $color__border;
+			position: relative;
+
+			&:last-child {
+				border-bottom: 0;
+			}
+
+			@include media( mobile ) {
+				border-bottom: 0;
+
+				&:after {
+					border-right: 1px solid transparent;
+					bottom: 0;
+					content: '';
+					position: absolute;
+					right: -30px;
+					top: 0;
+				}
+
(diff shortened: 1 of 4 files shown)


laurelfulford:
Thanks!

PR #87, opened 2019-08-24
Adds optional image caption to the homepage article blocks.

### LINKED ISSUE #46: Homepage Blocks: Add captions to images
The homepage blocks should be able to display some form of caption or credit for the featured image.

I'm not sure we want to show the full caption -- if the text is really long, it might not make sense for showing in smaller spaces -- or add a custom field/means of adding a credit to a photo to show (either custom in one of the plugins, or by using an [existing](https://wordpress.org/plugins/image-credits-nofollow/) [plugin](https://wordpress.org/plugins/better-image-credits/)).

laurelfulford:
### All Submissions:

This PR adds the option to show the image caption for the featured image in the Homepage Article Block.

It uses a very standard tag -- `figcaption` -- that right now is styled too specifically in the theme, so it's not being applied. I'm hoping to clean that up in a separate PR for the theme; ~~I may need to circle back and make adjustments for the font-scaling in the block after that, but for the time being, there are no styles for the caption in the block (so it looks a bit blah).~~ 

Edited to add: I did make some size and spacing adjustments for the block in 63bbae8, but tried to keep them in line with the level of styles added to the block so far.

I updated the markup used for the featured image to a `<figure>` tag, which is more standard with what's used for the image block.

I set the caption to show by default, but I'm wondering if we should set it to 'off' by default instead -- would be interested in feedback on that!

image

image

Closes #46.

2. Add a article block with a featured image and a caption. 
3. Confirm that the toggle 'Show Featured Image Caption' shows/hides the caption in the editor and on the front-end.
4. Confirm that that toggle disappears when the featured image is hidden.
5. Confirm that when an image doesn't have a caption, the `<figcaption></figcaption>` is not output, even when it's set to show a caption.

### Other information:

### DIFF
--- a/class-newspack-blocks-api.php
+++ b/class-newspack-blocks-api.php
@@ -103,13 +103,7 @@ class Newspack_Blocks_API {
 	 * @param Array $object  The object info.
 	 */
 	public static function newspack_blocks_get_image_caption( $object ) {
-		if ( 0 === $object['featured_media'] ) {
-			return;
-		}
-
-		$featured_image_caption = wp_get_attachment_caption( $object['featured_media'] );
-
-		return $featured_image_caption;
+		return (int) $object['featured_media'] > 0 ? trim( wp_get_attachment_caption( $object['featured_media'] ) ) : null;
 	}
 
 	/**
--- a/class-newspack-blocks-api.php
+++ b/class-newspack-blocks-api.php
@@ -24,6 +24,16 @@ class Newspack_Blocks_API {
 			)
 		);
 
+		register_rest_field(
+			array( 'post', 'page' ),
+			'newspack_featured_image_caption',
+			array(
+				'get_callback'    => array( 'Newspack_Blocks_API', 'newspack_blocks_get_image_caption' ),
+				'update_callback' => null,
+				'schema'          => null,
+			)
+		);
+
 		/* Add author info source */
 		register_rest_field(
 			'post',
@@ -85,6 +95,21 @@ class Newspack_Blocks_API {
 		$featured_image_set['full'] = $feat_img_array_full[0];
 
 		return $featured_image_set;
+	}
+
+	/**
+	 * Get thumbnail featured image captions for the rest field.
+	 *
+	 * @param Array $object  The object info.
+	 */
+	public static function newspack_blocks_get_image_caption( $object ) {
+		if ( 0 === $object['featured_media'] ) {
+			return;
+		}
+
+		$featured_image_caption = wp_get_attachment_caption( $object['featured_media'] );
+
+		return $featured_image_caption;
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


laurelfulford:
> Before digging in, what's the argument in favor of showing a featured image caption in the block?

It was a request we had in feedback (I think from when we originally shared the designs -- I can try to find the source!). It's also something that does show up on other news sites - for example, here's Washington Post's front page:

image

And NYT:

image

I don't think it would make sense for all cases -- like for smaller articles it could start looking crowded. But it could be helpful for adding some additional context or credits to a prominent image.

class-newspack-blocks-api.php
```diff
@@ -87,6 +97,21 @@ public static function newspack_blocks_get_image_src( $object ) {
 		return $featured_image_set;
 	}
 
+	/**
+	 * Get thumbnail featured image captions for the rest field.
+	 *
+	 * @param Array $object  The object info.
+	 */
+	public static function newspack_blocks_get_image_caption( $object ) {
```
jeffersonrabb: Here's a slightly leaner version, which also has trims leading and trailing space:

```
	public static function newspack_blocks_get_image_caption( $object ) {
		return (int) $object['featured_media'] > 0 ? trim( wp_get_attachment_caption( $object['featured_media'] ) ) : null;
	}
```

src/blocks/homepage-articles/index.js
```diff
@@ -52,6 +52,10 @@ export const settings = {
 			type: 'boolean',
 			default: true,
 		},
+		showCaption: {
+			type: 'boolean',
+			default: true,
```
jeffersonrabb: I personally would advocate for defaulting to false. I feel like captioned photos on a homepage should be a very clear editorial decision.

class-newspack-blocks-api.php
laurelfulford: Ah thanks! Updated in 8aae960.

src/blocks/homepage-articles/index.js
laurelfulford: Good call! Default flipped to 'false' in 3569c27.

laurelfulford:
Thanks Jeff!

> Center alignment clashes a bit with the left alignment of the other meta fields:

> Wondering if we should do away with the line. For inspiration I looked at the New York Times home where images are sometimes captioned. There are also lots of horizontal rules separating stories and sections (something we might add at some point) and if we also put a rule under the caption there could be visual conflict.

The styles for the caption right now are coming from the theme (without the theme, the captions would be aligned left and not have a bottom border), so it's probably best to deal with them there. 

I'll spin up a separate PR with left aligned text when the captions don't appear in a regular post/page, and dig into removing the border in another.

laurelfulford:
Related theme tickets:

https://github.com/Automattic/newspack-theme/issues/391
https://github.com/Automattic/newspack-theme/issues/392

Issue #96, opened 2019-08-28
Article Block: Image sizes


laurelfulford:
Right now the block just uses the media library's large image size, which makes for odd image sizes. 

Let's do the following:

* Give the block a fixed image size (6:4) to avoid this
* Add a image sizing option (1:1, 4:6).

laurelfulford:
Looks like this should've been closed with #98 being merged!

PR #98, opened 2019-08-28
Article Block: Add image size options

laurelfulford:
### All Submissions:

This PR adds three fixed image sizes to the blocks plugin -- Landscape (800 x 600), Portrait (600 x 800) and Square (800 x 800). It's not a perfect solution to our image size issue (and I've outlined why below), but hopefully it's a good starting point towards a solid, useful fix. 

Some known issues:

* If the original image is smaller than the custom image size, it won't be cropped.
* The image sizes I've picked -- 800 x 600, 600 x 800 and 800 x 800 -- are a bit to small to fill the content space (1200); especially if we're talking about images for retina screens, where we'd ideally be starting with images twice as large as they need to be (so a minimum of 2400px).

These two images kind of push and pull each other: we need to make the image sizes smaller to work with small archive images (or small new images! Sometimes you just can't get a larger photo). But we also need to go larger to make them work in more layouts -- like a one column post -- and work better on a variety of screen sizes. 

One way around this might be to have different sizes of each shape (a small, medium, and large landscape image, for example); I'd be interested in getting some feedback on this overall, so we can land on a good way to approach it that isn't a big knotted mess. 

We also might want to incorporate the DropdownMenu for our block's Toolbar options, to turn some of these sets of buttons into a single dropdown of options (like how the block alignment is done).

See #96.

2. Install and run the [Regenerate Thumbnails](https://en-ca.wordpress.org/plugins/regenerate-thumbnails/) plugin.
3. Add an article block to a page; set it so it's showing images, and the posts in there actually have featured images. Ideally these image will be over 800px wide and over 800px tall to see the resizing in action, but having smaller images could also help you see how it doesn't work.
4. Clicking on the block, there should now be three new options in the Toolbar, for the different image shapes: 

image

5. Try each of the sizing and confirm that your larger featured images switch between Landscape, Portrait, and Square.
6. Publish your post between each change, and confirm those image shapes are being used on the front end, when you have images large enough.

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/view.php
+++ b/src/blocks/homepage-articles/view.php
@@ -102,18 +102,25 @@ function newspack_blocks_render_block_homepage_articles( $attributes ) {
 				?>
 
 				<article <?php echo has_post_thumbnail() ? 'class="post-has-image"' : ''; ?>>
-					<?php if ( has_post_thumbnail() && $attributes['showImage'] ) : ?>
 
+					<?php if ( has_post_thumbnail() && $attributes['showImage'] && $attributes['imageShape'] ) : ?>
 						<figure class="post-thumbnail">
 							<a href="<?php echo esc_url( get_permalink() ); ?>" rel="bookmark">
-								<?php the_post_thumbnail( 'large' ); ?>
+								<?php
+								if ( 'landscape' === $attributes['imageShape'] ) {
+									the_post_thumbnail( 'newspack-article-block-landscape' );
+								} elseif ( 'portrait' === $attributes['imageShape'] ) {
+									the_post_thumbnail( 'newspack-article-block-portrait' );
+								} else {
+									the_post_thumbnail( 'newspack-article-block-square' );
+								}
+								?>
 							</a>
 
 							<?php if ( $attributes['showCaption'] && '' !== get_the_post_thumbnail_caption() ) : ?>
 								<figcaption><?php the_post_thumbnail_caption(); ?>
 							<?php endif; ?>
 						</figure><!-- .featured-image -->
-
 					<?php endif; ?>
 
 					<div class="entry-wrapper">
@@ -259,6 +266,10 @@ function newspack_blocks_register_homepage_articles() {
 				'imageScale'    => array(
 					'type'    => 'integer',
 					'default' => 3,
+				),
+				'imageShape'    => array(
+					'type'    => 'string',
+					'default' => 'landscape',
 				),
 				'sectionHeader' => array(
 					'type'    => 'string',
--- a/src/blocks/homepage-articles/view.php
+++ b/src/blocks/homepage-articles/view.php
@@ -112,53 +112,11 @@ function newspack_blocks_render_block_homepage_articles( $attributes ) {
 								$image_size = 'newspack-article-block-landscape-tiny';
 
 								if ( 'landscape' === $attributes['imageShape'] ) {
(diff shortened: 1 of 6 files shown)


laurelfulford:
I reworked this PR a bit to iron out some sizing issues -- it's not 100% yet, but I think it will take care of the image scaling issues we're seeing on the charter sites. 

The approach is:

Set up three image shapes: landscape, portrait, and square (as described above).

For each, set up four sizes: large, medium, small and tiny. 

By default, the 'tiny' version of the image will be used, then, as available (AKA as there is a generated image that matches the size we've set for the large, medium, or small sizes) the next largest version of the image will be used. That way, the images are most likely to be the same aspect ratio.

Then, with CSS, the images are scaled up to fill the available space. 

Known issues: 

* Images will get blurry if they're too small and scaled up; there's only so much we can do here if the originals are too small. The alternative is not scaling them up, and rather than being blurry, they're just weirdly small compared to the rest.
* On the front-end, we're using images up to 1200px wide; we should optimize this further so we're not loading too-large images. 
* I wrote some pretty repetitive, clunky code to do this; if it seems like a workable approach, I can rework it. 

### Screenshots:

**Landscape:**

image

**Portrait:**

image

**Square:**

image

### How to test the changes in this updated Pull Request:

2. Install and run the [Regenerate Thumbnails](https://en-ca.wordpress.org/plugins/regenerate-thumbnails/) plugin.
3. Add [these four images](https://cloudup.com/csINxGVskGQ) as featured images for the four most recent articles, for testing -- they are all too small, and the wrong shapes, to be cropped correctly.
4. Add an article block to a page and set it to show at least four posts. 
5. Try the different image shapes with different settings (the articles in columns, so they're narrower; the images floated left and right). 
image
6. Confirm that the images are the correct shape in the editor on the front end. You will see some blurring (especially on that 'Vertical' image), but it should be the same size as the other images in the group. 

And as a general note: going forward we can recommend minimum image sizes for this block; this fallback behaviour is mostly for making sure existing articles don't end up with too weirdly sized images.

laurelfulford:
@jeffersonrabb discussed this PR in Slack; in f477163 I made some updates based on his feedback to optimize the code and reduce repetition.

src/blocks/homepage-articles/view.php
```diff
@@ -102,18 +102,30 @@ function newspack_blocks_render_block_homepage_articles( $attributes ) {
...
-					<?php if ( has_post_thumbnail() && $attributes['showImage'] ) : ?>
+
+					<?php if ( has_post_thumbnail() && $attributes['showImage'] && $attributes['imageShape'] ) : ?>
 
 						<figure class="post-thumbnail">
 							<a href="<?php echo esc_url( get_permalink() ); ?>" rel="bookmark">
-								<?php the_post_thumbnail( 'large' ); ?>
+								<?php
+
+								$image_size = 'newspack-article-block-landscape-tiny';
```
jeffersonrabb: You should be able to simplify all of this to just one line:

```
$image_size = newspack_blocks_image_size_for_orientation( $attributes['imageShape'] );
```

laurelfulford: Oh whoops! Yeah, still stuck on the idea of making this needlessly complicated! This has been fixed in 847ba3b

laurelfulford:
Thanks Jeff! The outstanding issue above should be fixed!

laurelfulford:
Thanks Jeff!

PR #102, opened 2019-09-03
Remove 'more' button option from the article block.

### LINKED ISSUE #94: Article Block: Kill the more link
Right now when you use the 'More' button, it goes to the archive page of the current category (soon to be expanded).

Is there a smart way to offset where you end up when clicking 'more' -- like an anchor, or even offsetting the posts on the archive page?

Note that for either, you may have a block with three articles, but there could be articles from that same category further up the page -- so the offset would need to be based on which articles these are (the third, fourth and fifth on the page from this category?), not just how many they are (three posts from this category).

laurelfulford:
### All Submissions:

Since it mostly duplicates the button block, but with less functionality and more complexity, we've opted to kill the in-block 'more' button. The recommended alternative is using the button link; if we find we need something 'smarter' (like automatically generating a link with an offset) we can approach that separately.

Closes #94.

1. Apply the PR and run `npm run build:webpack`; confirm there are no errors.
2. Confirm that the 'more' button option is no longer there.
3. Double-check it's no longer referenced in the code. 

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/edit.js
+++ b/src/blocks/homepage-articles/edit.js
@@ -39,15 +39,7 @@ const MAX_POSTS_COLUMNS = 6;
 class Edit extends Component {
 	renderPost = post => {
 		const { attributes } = this.props;
-		const {
-			showImage,
-			showExcerpt,
-			showAuthor,
-			showAvatar,
-			showDate,
-			sectionHeader,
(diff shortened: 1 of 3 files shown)


laurelfulford:
Thanks!

PR #114, opened 2019-09-16
Article Blocks: Add text colour picker

### LINKED ISSUE #97: Article Block: Add text colour picker
The ability to change the block's text colour would make it work against more situations.

Right now it's always dark grey, except when nested in a cover block it's white.

This would help it work against more backgrounds (like with a darker group block background you could change it to white, or if using a lighter cover block background, you could override the default white text for all nested blocks in that case and pick black).

laurelfulford:
### All Submissions:

This PR adds a text colour picker to the Article Blocks; the purpose is to change the text to white or another light colour when the block is placed on something with a dark background, but it should allow you to change the text to any colour.

Note: For style packs that use the primary colour for the text in the header (Default, 3 and 4), if the primary colour in the theme is customized, it will override the block's setting in the editor.

It's more appropriate to fix this in the theme, so I'll tackle that in a separate PR. 

Closes #97.

2. Add an article block to a page, and switch it to a pre-defined colour from one of the first four spots (the primary and secondary colour); confirm it's applied.
3. Add a second block to the page, and pick a custom colour; confirm it's applied.
4. Save and publish; confirm the colours are applied on the front-end.
5. Navigate to Customize > Colours; switch the colour (primary or secondary) that you applied to the first block.
6. View that first block on the front-end and in the editor; confirm that the colour now used is your updated primary/secondary colour.

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/edit.js
+++ b/src/blocks/homepage-articles/edit.js
@@ -22,7 +22,6 @@ import {
 	RangeControl,
 	Toolbar,
 	ToggleControl,
-	withFallbackStyles,
 	Dashicon,
 	Placeholder,
 	Spinner,
@@ -38,16 +37,6 @@ const { decodeEntities } = wp.htmlEntities;
  * Module Constants
  */
 const MAX_POSTS_COLUMNS = 6;
-
-const applyFallbackStyles = withFallbackStyles( ( node, ownProps ) => {
-	const { textColor } = ownProps.attributes;
-	const editableNode = node.querySelector( '[contenteditable="true"]' );
-	//verify if editableNode is available, before using getComputedStyle.
-	const computedStyles = editableNode ? getComputedStyle( editableNode ) : null;
-	return {
-		fallbackTextColor: textColor || ! computedStyles ? undefined : computedStyles.color,
-	};
-} );
 
 class Edit extends Component {
 	renderPost = post => {
@@ -399,7 +388,6 @@ class Edit extends Component {
 
 export default compose( [
 	withColors( { textColor: 'color' } ),
-	applyFallbackStyles,
 	withSelect( ( select, props ) => {
 		const { postsToShow, author, categories, tags, single, singleMode } = props.attributes;
 		const { getAuthors, getEntityRecords } = select( 'core' );
--- a/src/blocks/homepage-articles/edit.js
+++ b/src/blocks/homepage-articles/edit.js
@@ -114,7 +114,6 @@ class Edit extends Component {
 			latestPosts,
 			isSelected,
 			textColor,
-			customTextColor,
 			setTextColor,
 		} = this.props;
 		const hasPosts = Array.isArray( latestPosts ) && latestPosts.length;
@@ -285,7 +284,6 @@ class Edit extends Component {
 			hasPosts,
 			categoriesList,
 			textColor,
-			customTextColor,
 		} = this.props; // variables getting pulled out of props
 		const {
 			showExcerpt,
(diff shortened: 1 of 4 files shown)


src/blocks/homepage-articles/edit.js
```diff
@@ -100,6 +102,9 @@ class Edit extends Component {
 			setAttributes,
 			latestPosts,
 			isSelected,
+			textColor,
+			customTextColor,
```
jeffersonrabb: I'm confused about `customTextColor` — I don't see it being set anywhere. Is it still in use?

src/blocks/homepage-articles/edit.js
```diff
@@ -38,7 +40,7 @@ const MAX_POSTS_COLUMNS = 6;
 
 class Edit extends Component {
 	renderPost = post => {
-		const { attributes } = this.props;
+		const { attributes, textColor } = this.props;
```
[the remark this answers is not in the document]
laurelfulford: Whoops! I really was throwing everything at the wall to see what sticked :) This was removed in b0b9bf3

src/blocks/homepage-articles/edit.js
laurelfulford: I tested removing it and everything still works as expected, so no, doesn't look like it's doing anything :) Removed in 3689c17 !

laurelfulford:
Thanks Jeff! These issues should now all be resolved.

src/blocks/homepage-articles/index.js
```diff
@@ -108,6 +108,14 @@ export const settings = {
 			type: 'boolean',
 			default: false,
 		},
+		textColor: {
+			type: 'string',
+			default: '',
+		},
+		customTextColor: {
```
jeffersonrabb: We should be able to remove this.

src/blocks/homepage-articles/view.php
```diff
@@ -64,13 +64,27 @@ function newspack_blocks_render_block_homepage_articles( $attributes ) {
...
+	if ( isset( $attributes['textColor'] ) ) {
+		$classes .= ' has-text-color';
+		if ( '' !== $attributes['textColor'] ) {
+			$classes .= ' has-' . $attributes['textColor'] . '-color';
+		}
+	}
+
+	$styles = '';
+
+	if ( '' !== $attributes['customTextColor'] ) {
```
jeffersonrabb: With `customTextColor` gone this block should be removable.

src/blocks/homepage-articles/view.php
```diff
@@ -253,6 +268,14 @@ function newspack_blocks_register_homepage_articles() {
 					'type'    => 'boolean',
 					'default' => false,
 				),
+				'textColor'       => array(
+					'type'    => 'string',
+					'default' => '',
+				),
+				'customTextColor' => array(
```
jeffersonrabb: We can remove this.

src/blocks/homepage-articles/index.js
laurelfulford: I tried removing all of this, and it stops a custom text colour from saving -- it will preview initially, but isn't applied if you navigate away from the editor and return -- or appearing on the front-end. Picking a colour from the pre-defined palette will still work though. 

I think need to leave it all in place for the value to save (index.js), and then for it to add the inline colour style (view.php). I am guessing, though 😬

laurelfulford:
It looks like that customTextColor value is still needed to actually save a custom colour and display it on the front-end. Without it, you can pick a custom colour, and it will preview, but if you save and return to the editor, it will be back to black, and it never shows on the front end. 

(Just let me know if you're experiencing differently!)

jeffersonrabb:
> It looks like that customTextColor value is still needed to actually save a custom colour and display it on the front-end. Without it, you can pick a custom colour, and it will preview, but if you save and return to the editor, it will be back to black, and it never shows on the front end.

Appears you're absolutely right! I wasn't understanding how `PanelColorSettings` works.

laurelfulford:
Thanks for digging into that, Jeff! I wasn't really sure how it worked, either, but I'm glad it does 😄

PR #129, opened 2019-09-26
Columns Style: Fix editor preview of column reordering

laurelfulford:
### All Submissions:

There appears to have been a change in the core column block styles, so the editor preview of our 'reordering' block style was overlapping a bit.

This PR fixes that by making the selector a bit more specific.

1. Make sure you're running the latest version of the Gutenberg plugin.
2. Add a columns block to a page, and change the style to the 'Move first column to second' style.

image

3. Note the appearance in the editor -- the first and second columns should be kind of overlapping:

image

4. Apply the PR and run `npm run build:webpack`
5. Confirm that the issue is resolved.

image

### Other information:

### DIFF
--- a/src/block-styles/core/columns/editor.scss
+++ b/src/block-styles/core/columns/editor.scss
@@ -1,7 +1,7 @@
 @import "../../../shared/sass/variables";
 @import "../../../shared/sass/mixins";
 
-.wp-block-columns {
+.wp-block .wp-block-columns {
 	@include media(mobile) {
 		&[class*='is-style-first-col-to'] [data-type="core/column"] {
 			margin-left: 46px;


laurelfulford:
Thanks for the test, @claudiulodro! Huh, I can't get this _not_ to work for some reason 😕 Could you please try clearing your cache, and if that doesn't do the trick, please let me know what browser and version you're using? Hoping one of those two things helps narrow this down. 

> I don't see anything in columns/editor.scss that would actually affect the desktop styling of the columns?

I think this might be due to this very confusing Sass:

`@include media(mobile)`

... which reads like the styles are applied to mobile screens and smaller. But when compiled it actually turns into:

`@media only screen and (min-width: 600px)`

It might be a code good readability improvement to switch this in the block and the theme to something clearer -- like `mobile-plus` or `mobile-minwidth`.

laurelfulford:
Thanks @claudiulodro! Heh, caching has been the story of my life, with all the JavaScript 😄 -- thanks for circling back to this!

PR #133, opened 2019-09-27
Add autocomplete search to article block author/tag/category inputs

claudiulodro:
### All Submissions:

The REST API limit for fetching a bunch of stuff is 100 items. Virtually all large sites have >100 users and >100 tags, so the current select-dropdown-based author/category/tag select will not be usable on a real site. 

To address this problem, I've created and added an autocompleting search component for those fields. 

The fields in various states look like this:

Screen Shot 2019-10-01 at 12 04 44 PM

### Component

I've introduced a new `AutocompleteTokenField` component to handle all of the heavy lifting. This component should be fairly general-purpose, so it can be used in @georgeh's query block work if needed. One important thing this does compared to the regular `FormTokenField` is enforce known values; you can't e.g. just create new tags like with the normal one.

I designed it primarily to be super-usable for block attributes. You just pass in the raw attributes and two functions that convert those attributes into objects with `label` and `value` properties:

```
// Convert a string into autocomplete results.
const fetchTagSuggestions = ( search ) => {
	return apiFetch( {
		path: addQueryArgs( '/wp/v2/tags', {
			search,
			per_page: 20,
			_fields: 'id,name',
			orderby: 'count',
			order: 'desc',
		} ),
	} ).then( function( tags ) {
		return map( tags, ( tag ) => ( {
			value: tag.id,
			label: decodeEntities( tag.name ) || __( '(no title)' ),
		} ) );
	} );
};
// Convert saved attribute value into recognizable data.
const fetchSavedTags = ( tagIDs ) => {
	return apiFetch( {
		path: addQueryArgs( '/wp/v2/tags', {
			per_page: 100,
			_fields: 'id,name',
			include: tagIDs.join(','),
		} ),
	} ).then( function( tags ) {
		return map( tags, ( tag ) => ( {
			value: tag.id,
			label: decodeEntities( tag.name ) || __( '(no title)' ),
		} ) );
	} );
};

<AutocompleteTokenField
	tokens={ tags || [] }
	onChange={ ( tokens ) => setAttributes( { tags: tokens } ) }
	fetchSuggestions={ fetchTagSuggestions }
	fetchSavedInfo={ fetchSavedTags }
	label={ __( 'Tag' ) }
/>
```

1. `npm run build:webpack`
2. Insert an Article block. Try out the fields and verify they work as described above.
3. Save the page. Verify the fields save and load correctly on a saved page. Verify the fields filter correctly on the site frontend.

### Other information:


rickybanister:
Is there a 'token' component for gutenberg? I know we had one in Calypso.

rickybanister:
Could we use the tag field from the editor sidebar:

image

claudiulodro:
I think we should be able to do something like that if you think it's a better way to go.

jeffersonrabb:
> Is there a 'token' component for gutenberg? I know we had one in Calypso.

I believe it's this: https://github.com/WordPress/gutenberg/tree/master/packages/components/src/form-token-field

thomasguillot:
I will +1 what @rickybanister said. We should be using the token component here so no need for the "change" button.

claudiulodro:
I've updated everything to use the multiselect token input as recommended.

src/blocks/homepage-articles/components/autocomplete-tokenfield.js
```diff
@@ -0,0 +1,151 @@
+import { Component } from '@wordpress/element';
+import { FormTokenField, Spinner } from '@wordpress/components';
+import { withState } from '@wordpress/compose';
```
jeffersonrabb: Looks like `withState` isn't being used.

```suggestion
```

src/blocks/homepage-articles/components/autocomplete-tokenfield.js
```diff
@@ -0,0 +1,151 @@
+import { Component } from '@wordpress/element';
```
jeffersonrabb: Missing comments block:

```
/**
 * WordPress dependencies
 */
```

src/blocks/homepage-articles/components/autocomplete-tokenfield.js
```diff
@@ -0,0 +1,151 @@
...
+
+/**
+ * An multi-selecting, api-driven autocomplete input suitable for use in block attributes.
+ */
+class AutocompleteTokenField extends Component {
+	constructor( props ) {
+		super( props );
+		this.state = {
+			suggestions: [],
+			validValuesByLabel: {},
```
jeffersonrabb: Having two objects in state that are basically the same data makes me a bit nervous—if somehow they get out of sync things could get screwy. Is there any way to store the data just once and use Javascript Array methods to convert to the orientation needed at various points in the flow?

src/blocks/homepage-articles/components/autocomplete-tokenfield.js
```diff
@@ -0,0 +1,151 @@
...
+	 */
+	componentDidMount() {
+		const { tokens, fetchSavedInfo } = this.props;
+
+		if ( ! tokens.length || ! fetchSavedInfo ) {
+			return;
+		}
+
+		this.setState( { loading: true }, () => {
+			fetchSavedInfo( tokens )
```
jeffersonrabb: For the Saved info, is there a way to use `withSelect` to retrieve these rather than making our own API calls?

src/blocks/homepage-articles/edit.js
```diff
@@ -136,35 +138,147 @@ class Edit extends Component {
...
+
+		const fetchAuthorSuggestions = ( search ) => {
+			return apiFetch( {
+				path: addQueryArgs( '/wp/v2/users', {
+					search,
+					per_page: 20,
+					_fields: 'id,name',
+				} ),
+			} ).then( function( users ) {
+				return map( users, ( user ) => ( {
```
jeffersonrabb: Any reason this uses `lodash` `map` instead of normal Javascript `Array.prototype.map`?

src/blocks/homepage-articles/index.js
```diff
@@ -80,14 +80,14 @@ export const settings = {
 			type: 'string',
 			default: 'top',
 		},
-		author: {
-			type: 'string',
+		authors: {
```
jeffersonrabb: Changing the tense of author is going to mean any existing versions of the block with an Author set will break. I think this is probably fine given we're pre-launch, just noting.

src/blocks/homepage-articles/edit.js
```diff
@@ -136,35 +138,147 @@ class Edit extends Component {
...
+											fetchSuggestions={ fetchAuthorSuggestions }
+											fetchSavedInfo={ fetchSavedAuthors }
+											label={ __( 'Author' ) }
+										/>
+									</BaseControl>
+									<BaseControl>
+										<AutocompleteTokenField
+											tokens={ categories || [] }
+											onChange={ ( tokens ) => setAttributes( { categories: tokens } ) }
+											fetchSuggestions={ fetchCategorySuggestions }
```
jeffersonrabb: For all the autocomplete lookups I suggest using `lodash` `debounce` to limit the number of API calls made as the user types. Here's an example of similar usage in the Jetpack Map block: https://github.com/Automattic/jetpack/blob/master/extensions/blocks/map/lookup/index.js

src/blocks/homepage-articles/components/autocomplete-tokenfield.js
claudiulodro: https://github.com/Automattic/newspack-blocks/pull/133/commits/022db0ba07c0e06b374c91a3af20db822283e106

src/blocks/homepage-articles/components/autocomplete-tokenfield.js
claudiulodro: Good catch. https://github.com/Automattic/newspack-blocks/pull/133/commits/022db0ba07c0e06b374c91a3af20db822283e106

src/blocks/homepage-articles/components/autocomplete-tokenfield.js
claudiulodro: There should be. I wasn't able to easily refactor it after a couple hours of fiddling with it, so I'll open a follow-up issue after this is merged to do that.

src/blocks/homepage-articles/edit.js
claudiulodro: No good reason. The core code I was looking at used lodash map and I got tunnel vision. :) https://github.com/Automattic/newspack-blocks/pull/133/commits/7e941b38671b1df9020fa44d3684133af0237394

src/blocks/homepage-articles/components/autocomplete-tokenfield.js
claudiulodro: Yeah good point. That's possible. It will be less performant if the user has been doing a ton of searches because there could be a few hundred valid elements recognized, but the browser can trivially handle that so I don't think it will be an issue. https://github.com/Automattic/newspack-blocks/pull/133/commits/a077b3fc9cd20e14954d585af289db88f2cc92bb

src/blocks/homepage-articles/edit.js
claudiulodro: Good idea. I've added a 500ms delay for the autocomplete: https://github.com/Automattic/newspack-blocks/pull/133/commits/fa1785a647cff20c905f061cf2c178c43abfa6d4

src/blocks/homepage-articles/components/autocomplete-tokenfield.js
jeffersonrabb: That commit removes it from `src/blocks/homepage-articles/edit.js`, not `src/blocks/homepage-articles/components/autocomplete-tokenfield.js` (although it wasn't in use there either, so it was a good catch!)

claudiulodro: Oops! https://github.com/Automattic/newspack-blocks/pull/133/commits/3275bff88c909577b82757aa37a9f95f560989fd

src/blocks/homepage-articles/components/autocomplete-tokenfield.js
```diff
@@ -0,0 +1,186 @@
...
+		this.debouncedUpdateSuggestions.cancel();
+	}
+
+	/**
+	 * Get a list of labels for input values.
+	 *
+	 * @param array values Array of values (ids, etc.).
+	 * @return array Array of valid labels corresponding to the values.
+	 */
+	getLabelsForValues( values ) {
```
jeffersonrabb: You could also get at this with `Array.reduce` like this:

```
getLabelsForValues( values ) {
		const { validValues } = this.state;
		return values.reduce(
			( accumulator, value ) =>
				validValues[ value ] ? [ ...accumulator, validValues[ value ] ] : accumulator,
			[]
		);
	}
```

src/blocks/homepage-articles/components/autocomplete-tokenfield.js
```diff
@@ -0,0 +1,186 @@
...
+		return labels;
+	}
+
+	/**
+	 * Get a list of values for input labels.
+	 *
+	 * @param array labels Array of labels from the tokens.
+	 * @return array Array of valid values corresponding to the labels.
+	 */
+	getValuesForLabels( labels ) {
```
jeffersonrabb: Slightly leaner approach:

```
getValuesForLabels( labels ) {
		const { validValues } = this.state;
		return labels.map( label =>
			Object.keys( validValues ).find( key => validValues[ key ] === label )
		);
	}
```

src/blocks/homepage-articles/components/autocomplete-tokenfield.js
claudiulodro: I like it. Thanks for the suggestion. https://github.com/Automattic/newspack-blocks/pull/133/commits/b4b6e693f380eabbdd89e3a920f75b56cdace845

src/blocks/homepage-articles/components/autocomplete-tokenfield.js
claudiulodro: Very nice. Thanks for the suggestion. https://github.com/Automattic/newspack-blocks/pull/133/commits/b4b6e693f380eabbdd89e3a920f75b56cdace845

PR #135, opened 2019-09-30
Reduce size of smallest type scale options

### LINKED ISSUE #130: Article Block: Make the small font scale size smaller
The small font scale size doesn't really get small enough -- consider knocking down the smaller sizes to make it closer to 10-12 than 14.

Also make sure that the post meta doesn't display larger than the article title.

laurelfulford:
### All Submissions:

This PR reduces the size of the smallest type scale option for the block (based on feedback), and refines the differences between different sizes a bit.

Closes #130.

1. Copy-paste [this test content](https://cloudup.com/comkHU8FNUj) into a post or page; it has a copy of the article block in each size. 
2. Note its appearance.
3. Apply the PR and run `npm run build:webpack`
4. Confirm that the smallest size is now smaller (14px); also confirm that the difference between each of the different sizes seems fairly consistent, and they're noticeably different from each other. 

(And just let me know if anything is "off"!). 

### Other information:

(I messed something up with the version bump PR, causing the commit to show up here -- master should already have it set to '6').

### DIFF
--- a/newspack-blocks.php
+++ b/newspack-blocks.php
@@ -7,7 +7,7 @@
  * Author URI:      YOUR SITE HERE
  * Text Domain:     newspack-blocks
  * Domain Path:     /languages
- * Version:         1.0.0-alpha.5
+ * Version:         1.0.0-alpha.6
  *
  * @package         Newspack_Blocks
  */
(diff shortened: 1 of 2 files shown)


laurelfulford:
Thanks @claudiulodro!

PR #138, opened 2019-10-04
Group Block: Add styles to align block left and right

### LINKED ISSUE #103: Group Block: Align left or right
It would be great to be able to align the group block left or right, so you could nest, say, a couple paragraphs inside it and have it sitting to the side of the content. 

I need to double-check if this is in the queue for a Gutenberg setting; for the short-term, though, we could add a block style.

laurelfulford:
### All Submissions:

By default, the group block has a default alignment, and lets you align it wide and full. This PR adds two block styles that also let you align it right and left.

image

Because we can't extend the core block to add proper alignments, this turned into a bit of a hacky work around, primarily in the editor:

1. **The options live in a different location than the typical alignment options:**

image

Because of that, you can apply both a 'wide' align and an 'align left' style. In this case, the 'wide' or 'full' align will override the left and right align styles -- you can only left or right align the block when it has the default width.

2. **We can't use the typical editor selectors:** When using the default alignments, the editor will place a `data-align` attribute on the top level element for the block. When using a style, the class is placed further inside of the block -- this means the styles we can add to align the block left and right can't "grab" all of the block pieces... meaning some of the Group block controls will not be part of the left or right alignment:

Aligned left:
image

Aligned right:
image

Hopefully this is minor enough not to be a problem; I'm not quite sure how we would work around it. 

I will make a separate PR in the theme to pull these blocks outside of the content area where appropriate. 

Closes #103.

2. Add a group block to the editor; click the styles and align it left. 
3. Save and publish; confirm it displays aligned left on the front end; scale down the browser window and confirm it turns full-width at the mobile breakpoint (600px).
4. Add a second block and align it right; repeat step 3 for this block.
5. Try setting each of these blocks align wide and align full; confirm that this completely overrides the left and right styles, and they display as normal wide and full alignments. 

### Other information:

### DIFF
--- a/src/block-styles/core/group/editor.scss
+++ b/src/block-styles/core/group/editor.scss
@@ -1,0 +1,27 @@
+@import "../../../shared/sass/variables";
+@import "../../../shared/sass/mixins";
+
+@include media( tablet ) {
+	[data-type="core/group"]:not([data-align]) {
+		.wp-block-group {
+			position: relative;
+
+			&.is-style-alignleft,
+			&.is-style-alignright {
+				margin-bottom: 1em;
+				width: 50%;
+				z-index: 5;
+			}
+
+			&.is-style-alignleft {
+				float: left;
+				margin-right: 1em;
+			}
+
+			&.is-style-alignright {
+				float: right;
+				margin-left: 1em;
+			}
+		}
+	}
+}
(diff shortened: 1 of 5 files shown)


laurelfulford:
I had no idea that was possible!

> If this works is this PR unnecessary?

Yeah I think so -- let me give it a try this morning and confirm. I'll also need to update https://github.com/Automattic/newspack-theme/pull/465.

laurelfulford:
Closing in favour of #141, which uses the above suggested approach -- thanks @jeffersonrabb!

Issue #48, opened 2019-07-09
Homepage Blocks: Carousel


laurelfulford:
This may be achievable by nesting the existing Homepage Blocks, one each, inside of a separate existing slideshow block.

Example from [The Center for Public Integrity](https://publicintegrity.org/):

![image](https://user-images.githubusercontent.com/177561/60904241-8e84c180-a227-11e9-923c-e2614a281e27.png)

laurelfulford:
Another request for carousel functionality came up in a design call -- the example given was [AJC](https://www.ajc.com/)'s local politics section:

![image](https://user-images.githubusercontent.com/177561/61143549-609db800-a487-11e9-8c28-8761f83d398e.png)

In this case, the carousel also allows you to include ads mixed in among the stories.

PR #143, opened 2019-10-08
Carousel Block

jeffersonrabb:
### All Submissions:

Adds a Carousel block that uses `amp-carousel` to display posts.  `amp-carousel` cannot currently be used in the editor because it does not support DOM mutation after page load, which occurs whenever the array of posts to be displayed changes. Eventually Bento AMP will resolve this, but until this is available we will use the Swiper library to make the block functional in the editor. Currently, there is only a placeholder in the editor.

Screen Shot 2019-10-08 at 12 18 29 AM

Closes https://github.com/Automattic/newspack-blocks/issues/48.

1. Check out the branch and execute `npm run build:webpack`
2. Add Articles Carousel block to a post.

### Other information:

### DIFF
--- a/src/blocks/carousel/edit.js
+++ b/src/blocks/carousel/edit.js
@@ -4,6 +4,7 @@
 import QueryControls from '../homepage-articles/query-controls';
 import createSwiper from './create-swiper';
 import classnames from 'classnames';
+import AutocompleteTokenField from '../homepage-articles/components/autocomplete-tokenfield.js';
 
 /**
  * External dependencies
@@ -25,13 +26,15 @@ import {
 	RangeControl,
 	Spinner,
 	ToggleControl,
+	BaseControl,
 } from '@wordpress/components';
 import { withSelect } from '@wordpress/data';
 import { withState, compose } from '@wordpress/compose';
+import apiFetch from '@wordpress/api-fetch';
+import { addQueryArgs } from '@wordpress/url';
+import { decodeEntities } from '@wordpress/html-entities';
 
 import { PanelColorSettings, withColors } from '@wordpress/block-editor';
-
-const { decodeEntities } = wp.htmlEntities;
 
 class Edit extends Component {
 	constructor( props ) {
@@ -87,13 +90,10 @@ class Edit extends Component {
 			isSelected,
 			latestPosts,
 			hasPosts,
-			authorList,
-			categoriesList,
-			tagsList,
 		} = this.props;
 		const { autoPlayState } = this.state;
 		const {
-			author,
+			authors,
 			autoplay,
 			categories,
 			delay,
@@ -109,6 +109,97 @@ class Edit extends Component {
 			'swiper-container',
 			autoplay && autoPlayState && 'wp-block-newspack-blocks-carousel__autoplay-playing'
 		);
+
+		const fetchAuthorSuggestions = search => {
+			return apiFetch( {
+				path: addQueryArgs( '/wp/v2/users', {
+					search,
+					per_page: 20,
+					_fields: 'id,name',
+				} ),
+			} ).then( function( users ) {
+				return users.map( user => ( {
+					value: user.id,
+					label: decodeEntities( user.name ) || __( '(no name)', 'newspack-blocks' ),
+				} ) );
+			} );
+		};
+		const fetchSavedAuthors = userIDs => {
+			return apiFetch( {
+				path: addQueryArgs( '/wp/v2/users', {
+					per_page: 100,
+					include: userIDs.join( ',' ),
+				} ),
+			} ).then( function( users ) {
+				return users.map( user => ( {
+					value: user.id,
+					label: decodeEntities( user.name ) || __( '(no name)', 'newspack-blocks' ),
+				} ) );
+			} );
+		};
+
+		const fetchCategorySuggestions = search => {
+			return apiFetch( {
+				path: addQueryArgs( '/wp/v2/categories', {
+					search,
+					per_page: 20,
+					_fields: 'id,name',
+					orderby: 'count',
+					order: 'desc',
+				} ),
+			} ).then( function( categories ) {
+				return categories.map( category => ( {
+					value: category.id,
+					label: decodeEntities( category.name ) || __( '(no title)', 'newspack-blocks' ),
+				} ) );
+			} );
+		};
+		const fetchSavedCategories = categoryIDs => {
+			return apiFetch( {
+				path: addQueryArgs( '/wp/v2/categories', {
+					per_page: 100,
+					_fields: 'id,name',
+					include: categoryIDs.join( ',' ),
+				} ),
+			} ).then( function( categories ) {
+				return categories.map( category => ( {
+					value: category.id,
+					label: decodeEntities( category.name ) || __( '(no title)', 'newspack-blocks' ),
+				} ) );
+			} );
+		};
+
+		const fetchTagSuggestions = search => {
+			return apiFetch( {
+				path: addQueryArgs( '/wp/v2/tags', {
+					search,
+					per_page: 20,
+					_fields: 'id,name',
+					orderby: 'count',
+					order: 'desc',
+				} ),
+			} ).then( function( tags ) {
+				return tags.map( tag => ( {
+					value: tag.id,
+					label: decodeEntities( tag.name ) || __( '(no title)', 'newspack-blocks' ),
+				} ) );
+			} );
+		};
+		const fetchSavedTags = tagIDs => {
+			return apiFetch( {
+				path: addQueryArgs( '/wp/v2/tags', {
+					per_page: 100,
+					_fields: 'id,name',
+					include: tagIDs.join( ',' ),
+				} ),
+			} ).then( function( tags ) {
+				return tags.map( tag => ( {
+					value: tag.id,
+					label: decodeEntities( tag.name ) || __( '(no title)', 'newspack-blocks' ),
+				} ) );
+			} );
+		};
+
 		return (
 			<Fragment>
 				<div className={ classes } ref={ this.carouselRef }>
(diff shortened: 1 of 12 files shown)


thomasguillot:
Any chance we can have some settings for the autoplay like for Jetpack Slideshow? (and maybe the effect).

https://cld.wthms.co/tjUxYc

jeffersonrabb:
> Any chance we can have some settings for the autoplay like for Jetpack Slideshow? (and maybe the effect).

We can support the toggle and the interval, but AFAIK `amp-carousel` does not support alternative transitions. Docs are [here](https://amp.dev/documentation/components/amp-carousel/). See anything I missed?

thomasguillot:
> We can support the toggle and the interval, but AFAIK `amp-carousel` does not support alternative transitions. Docs are [here](https://amp.dev/documentation/components/amp-carousel/). See anything I missed?

Ok, no problem for the transitions. The autoplay is much more needed 😊

jeffersonrabb:
> The autoplay is much more needed

Added in https://github.com/Automattic/newspack-blocks/pull/143/commits/07a9db473bba9d52ff0889489156c0b2b4f9ab7c

thomasguillot:
> Added in [07a9db4](https://github.com/Automattic/newspack-blocks/commit/07a9db473bba9d52ff0889489156c0b2b4f9ab7c)

That was quick! 😊
Can we have a play/pause button when using the autoplay?

thomasguillot:
@jeffersonrabb I committed a bunch of fixes:

* Update block icon
* Resize avatar size on > tablet
* Have links in white for better contrast against black
* Make sure `.entry-wrapper` doesn't overflow
* Add a `cursor: pointer` to the prev/next buttons

How about we add a new style? Something like this:

Screenshot 2019-10-09 at 14 46 46

Can we have 2 toggles (on by default) to:
1. show/hide the bullets, and
2. show/hide the arrows

jeffersonrabb:
> Can we have a play/pause button when using the autoplay?

Added in https://github.com/Automattic/newspack-blocks/pull/143/commits/07a9db473bba9d52ff0889489156c0b2b4f9ab7c

jeffersonrabb:
> show/hide the bullets, and show/hide the arrows

Wouldn't hiding these be compromising the block's usability and accessibility?

jeffersonrabb:
> How about we add a new style? 

I like the alternate style. Another possible style might be to use thumbnails instead of bullets?

thomasguillot:
> Wouldn't hiding these be compromising the block's usability and accessibility?

Yes and no 😊

* The arrows are technically not needed since you can navigate with the bullets.
* If you hide the bullets, a visitor will have no idea how many slides there are or which one is on. Maybe we could have a counter if the bullets are hidden, like "2 / 5".
* You can only hide both if you are on autoplay (otherwise how would a visitor see the rest of the carousel 😅). But yes, this option isn't great for usability and accessibility. Maybe we could display a warning message.

I think allowing these to be hidden offers a bit more flexibility to the block and front-end design.

thomasguillot:
> Another possible style might be to use thumbnails instead of bullets?

Thumbnails would be great!

thomasguillot:
In b981df6 I added a few more changes:

I moved the play/pause button to the top right corner like for Jetpack Slideshow.
I increased the maximum delay time to 20 s because I felt that 5 s was not always enough time to process, the image, the title, the author and the date.
And I set the default time to 5 s rather than 3 s.

thomasguillot:
One thing I noticed is if you completely disable the AMP plugin, there's no fallback and you get all the posts, one after the other.

newspack test_(1440 x 1048)

src/blocks/carousel/view.php
```diff
@@ -0,0 +1,230 @@
...
+			NEWSPACK_BLOCKS__VERSION,
+			true
+		);
+		wp_enqueue_script(
+			'amp-selector',
+			'https://cdn.ampproject.org/v0/amp-selector-0.1.js',
+			null,
+			NEWSPACK_BLOCKS__VERSION,
+			true
+		);
```
westonruter: The AMP plugin is already registering this script in `amp_register_default_scripts()`: https://github.com/ampproject/amp-wp/blob/039554da947563bd7a3432aa724282b1c20a86b9/includes/amp-helper-functions.php#L435-L499

These scripts are registered even when the page is not AMP. They are registered always. It gets `amp-runtime` as a dependency.

So no script is registered here.

So all you need to do here is:

```php
wp_enqueue_script( 'amp-carousel' );
wp_enqueue_script( 'amp-selector' );
```

In reality, you can even remove the `Newspack_Blocks::is_amp()` check because enqueueing those scripts will not hurt in AMP pages. The AMP plugin will ensure that those scripts are present if they weren't already enqueued when there is an `amp-carousel` or `amp-selector` in the page.

jeffersonrabb:
> One thing I noticed is if you completely disable the AMP plugin, there's no fallback and you get all the posts, one after the other.

Should be resolved in https://github.com/Automattic/newspack-blocks/pull/143/commits/b2c2a5b6883c12b8b7601fd1de2685de6cbb75b8

jeffersonrabb:
Editor implementation using Swiper: https://github.com/Automattic/newspack-blocks/pull/143/commits/5dc0ddbbb60e083ec463836a9aaa7fc9941bd85b

thomasguillot:
I added a few changes to the editor's stylesheet in 4f4523c

src/blocks/carousel/view.php
```diff
@@ -0,0 +1,301 @@
...
+	);
+	$article_query = new WP_Query( $args );
+	$counter       = 0;
+	ob_start();
+	if ( $article_query->have_posts() ) :
+		while ( $article_query->have_posts() ) :
+			$article_query->the_post();
+			if ( ! has_post_thumbnail() ) {
+				continue;
+			}
```
claudiulodro: Rather than fetching all posts and skipping posts with no thumbnails, we should ensure that we're fetching posts that contain thumbnails by adding a meta query for `_thumbnail_id` to the WP_Query args above.

src/blocks/carousel/edit.js
```diff
@@ -0,0 +1,321 @@
...
+					{ ! latestPosts && (
+						<Placeholder>
+							<Spinner />
+						</Placeholder>
+					) }
+					{ latestPosts && (
+						<Fragment>
+							<div class="swiper-wrapper">
+								{ latestPosts.map( post => (
+									<article className="post-has-image swiper-slide" key={ post.id }>
```
claudiulodro: Might be good to skip posts that don't have images here. In my brief research I didn't see a way to do a meta query with `getEntityRecords`, so it's not as easy to get only posts with thumbnails here as it is with WP_Query (still possible somehow, but it's going to be more work).

Articles that don't have a thumbnail render as a blank white slide (because they have a height of 0px, none of the title/author/etc. info is visible):

Screen Shot 2019-10-13 at 5 06 13 PM

Easiest solution (and most accurate preview of the frontend) is to skip articles that don't have a featured image here.

src/blocks/carousel/view.php
```diff
@@ -0,0 +1,301 @@
...
+			$counter++;
+			?>
+
+			<article <?php echo has_post_thumbnail() ? 'class="post-has-image"' : ''; ?>>
```
claudiulodro: This `has_post_thumbnail` check is redundant, because it will only ever be reached by posts that have already passed a `has_post_thumbnail` check.

src/blocks/carousel/edit.js
```diff
@@ -0,0 +1,321 @@
...
+						<Placeholder>{ __( 'Sorry, no posts were found.' ) }</Placeholder>
+					) }
+					{ ! latestPosts && (
+						<Placeholder>
+							<Spinner />
+						</Placeholder>
+					) }
+					{ latestPosts && (
+						<Fragment>
+							<div class="swiper-wrapper">
```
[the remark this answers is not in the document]
claudiulodro: https://github.com/Automattic/newspack-blocks/pull/143/commits/262f2d15da9b9f3960a6041964b76cfd1eb0e6e4

src/blocks/carousel/edit.js
```diff
@@ -0,0 +1,321 @@
...
+										<figure className="post-thumbnail">
+											{ post.newspack_featured_image_src && (
+												<a href="#" rel="bookmark">
+													<img src={ post.newspack_featured_image_src.landscape } />
+												</a>
+											) }
+										</figure>
+										<div className="entry-wrapper">
+											{ showCategory && post.newspack_category_info.length && (
+												<div class="cat-links">
```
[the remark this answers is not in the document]
claudiulodro: https://github.com/Automattic/newspack-blocks/pull/143/commits/262f2d15da9b9f3960a6041964b76cfd1eb0e6e4

src/blocks/carousel/edit.js
claudiulodro: https://github.com/Automattic/newspack-blocks/pull/143/commits/45586426ca5db8e9742330d6c49b65af6354649f Not 100% ideal because the preview doesn't exactly match up to the frontend (less elements in preview vs frontend), but it's an improvement.

src/blocks/carousel/view.php
claudiulodro: https://github.com/Automattic/newspack-blocks/pull/143/commits/003741c27571380cd475c49673d192d48db57d13

src/blocks/carousel/view.php
claudiulodro: https://github.com/Automattic/newspack-blocks/pull/143/commits/003741c27571380cd475c49673d192d48db57d13

src/blocks/carousel/edit.js
```diff
@@ -0,0 +1,321 @@
...
+				</div>
+				<InspectorControls>
+					<PanelBody title={ __( 'Display Settings' ) } initialOpen={ true }>
+						{ postsToShow && categoriesList && (
+							<QueryControls
+								enableSingle={ false }
+								numberOfItems={ postsToShow }
+								onNumberOfItemsChange={ value => setAttributes( { postsToShow: value } ) }
+								authorList={ authorList }
+								tagsList={ tagsList }
```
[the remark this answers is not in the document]
claudiulodro: https://github.com/Automattic/newspack-blocks/pull/143/commits/91ca6a8230b0f1d19af4793737724ad8a1bdaeda

PR #149, opened 2019-10-09
Improves column plus borders behaviour

### LINKED ISSUE #148: Column Block: Borders removed at wrong time
Right now, the border styles don't entirely behave as expected: when you have columns with custom widths, the borders disappear at unexpected times. 

Discovered when reviewing #147.

laurelfulford:
### All Submissions:

This is a partial fix for an issue with the columns and the border styles.

Right now, the even width columns display the borders correctly at the tablet breakpoint, but the (fixed) custom width columns do not. All but the last set of columns have custom widths; the sets with two columns are okay, but the one with three is messed up:

image

This PR switches it so that the borders display correctly when you have custom column widths (but breaks the 'default' widths):

image

I'm not sure if there's a great way around this - perhaps if we extend the column block further to add a class if it uses custom block widths? I'd be very interested in getting some feedback on possible approaches.

In the meantime, if you want even-width columns, you can manually set them to even widths (50%, 50%, or 33.33%, 33.33%. 33.33%) and it "fix" the issue. 

(Style pack 3 will need an update based on what we choose to do here, since they have custom border styles. **Edited to add** PR added here: https://github.com/Automattic/newspack-theme/pull/483 Style 5 -- the other style pack that changes these borders -- does it in a smarter way, and doesn't need any changes). 

Closes #148.

1. Copy paste [this test content](https://cloudup.com/csJfug8JLsO) into the editor, or recreate the above by adding column blocks with 2-3 columns, setting them either at custom or default widths, and adding borders.
2. View on front-end; reduce the size of the browser window and note where the columns break/borders disappear.
3. Apply the PR and run `npm run build:webpack`
4. Confirm that you get something closer to the second screenshot; the column blocks using the default width will break kind of oddly, but the ones with custom widths will not.

### Other information:

### DIFF
--- a/src/block-styles/core/columns/view.scss
+++ b/src/block-styles/core/columns/view.scss
@@ -34,17 +34,17 @@
 			@include media( mobile ) {
 				border-bottom: 0;
 
+				&:not(:first-child) {
+					margin-left: 24px;
+				}
+
 				&:after {
-					border-right: 1px solid transparent;
+					border-right: 1px solid $color__border;
 					bottom: 0;
 					content: '';
 					position: absolute;
-					right: -16px;
+					right: -12px;
 					top: 0;
-				}
-
-				&:nth-child(odd):after {
-					border-color: $color__border;
 				}
 
 				&:last-child:after {
@@ -53,8 +53,12 @@
 			}
 
 			@include media( tablet ) {
+				&:not(:first-child) {
+					margin-left: 32px;
+				}
+
 				&:after {
-					border-color: $color__border;
+					right: -16px;
 				}
 			}
 		}


laurelfulford:
@jeffersonrabb I'd appreciate your thoughts on this one -- maybe we can chat about it tomorrow -- to make sure I'm not weirdly complicating things in a bizarre direction :)

thomasguillot:
LGTM. It's much better and it definitely works 😊

laurelfulford:
Thanks Thomas!

I made a new issue to cover next steps for the columns block, as per our conversation in Slack: https://github.com/Automattic/newspack-blocks/issues/151 It's really not holding up for what we need it to do, and it will be better to 'phase out' with something else that is a bit more robust. 

If you have any further thoughts on what we should do there, please add to the comments!

PR #160, opened 2019-10-11
Add transforms to and from core/latest-posts

marekhrabe:
### All Submissions:

This makes it easy for users to switch to Newspack Homepage Articles if they were already using the core block for showing latest posts. It transforms all configuration attributes that are shared between the two blocks.

It also provides a transformation for the opposite direction. Due to Newspack having much more options for configuration, the way back is not lossless. Things like specific styling, filtering categories/authors/tags will be lost on the way because it's just not supported on the other side.

Shared options: number of posts to show, layouts (list/grid), number of grid columns, showing a date, and showing an excerpt.

1. Start a new page and insert "Latest Posts" block from the core.
2. Configure the block in different ways and always try to transform it into Newspack Homepage Articles and back to latest posts.
3. Confirm there are no errors coming from the transformation and it always succeeds. 
4. Undo button should undo the transformation in a single click.

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/index.js
+++ b/src/blocks/homepage-articles/index.js
@@ -2,6 +2,7 @@
  * External dependencies
  */
 import { Path, SVG } from '@wordpress/components';
+import { createBlock } from '@wordpress/blocks';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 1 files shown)


jeffersonrabb:
Could we transform the category filter as well? Homepage Articles supports multiple categories, so we probably transform by using the first item in the array.

marekhrabe:
> transform the category filter as well

I have missed this one! Thanks!

Support just added

PR #165, opened 2019-10-15
Generate separate editor assets per each block

marekhrabe:
### All Submissions:

Apart from the combined build of all the blocks together (current usage), this makes them also available as individual JS & CSS files.

This makes it easier to enqueue assets for a single block and a possibility to publish single blocks, as that currently looks like the future of blocks in WP.org block repo.

Closes # .

0. `npm install`
1. `npm run build:webpack`
2. Explore the `dist` folder generated
3. You show now see additional files - for example `dist/homepage-articles/editor.{js,css,rtl.css,deps.json}`

### Other information:

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -32,17 +32,22 @@ const blocks = fs
   .readdirSync( blocksDir )
   .filter( block => fs.existsSync( path.join( __dirname, 'src', 'blocks', block, 'editor.js' ) ) );
 
-// Helps split up each block into its own folder view script
-const viewBlocksScripts = blocks.reduce( ( viewBlocks, block ) => {
+// Helps split up each block into its own folder with view and editor scripts and styles.
+// Also creates a script one that combines all editor scripts for all blocks.
+const blocksScripts = blocks.reduce( ( scripts, block ) => {
 	const viewScriptPath = path.join( __dirname, 'src', 'blocks', block, 'view.js' );
 	if ( fs.existsSync( viewScriptPath ) ) {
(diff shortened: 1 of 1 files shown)


marekhrabe:
It was quite easy to split the frontend parts but what about the PHP? Can we also cleanly separate it plugin-by-plugin or is there too much shared stuff and connective tissue?

marekhrabe:
Things I'm exploring further:

- block assets are enqueued though non-standard way (using for example `Newspack_Blocks::enqueue_view_assets`). Updating it to use the standard Gutenberg way would loosen the requirement to have the shared class
- Homepage Articles require (I _think_) some additional things in the API, however the code is separated from the block - https://github.com/Automattic/newspack-blocks/blob/61217f85fcef2f108afd293a20cca8b15707df80/class-newspack-blocks-api.php#L16-L58 - maybe we can collocate?

obenland:
This looks like a pretty safe, backwards compatible step forward. I like it.

> block assets are enqueued though non-standard way (using for example Newspack_Blocks::enqueue_view_assets). Updating it to use the standard Gutenberg way would loosen the requirement to have the shared class

Yeah, I wonder if it would make sense to add reusable php file (and readme) where we could replace placeholders to bundle with the assets you generate here.

> Homepage Articles require (I think) some additional things in the API, however the code is separated from the block

This could be broken out and bundled with the block?

getdave:
I'm really sorry but I accidentally marked this as `Ready for Review` cos I clicked the wrong UI button in Github!!! 😭

marekhrabe:
This one is sooo not ready 😀

PR #166, opened 2019-10-15
Experiment with a build script

marekhrabe:
This continues my exploration from #165. It adds a very barebones bash script that takes the source of each block and collocates it with built files belonging to the block.

If we make sure that there is some main entry point for PHP in each block (and it doesn't require any additional files/classes) and there is a readme in each block, those generated directories in `assets/blocks` could be used individually as fully functional WP plugins. 

This is an exploration and I'm not sure it's the right way. While making this, I think I'd like to change the approach in a way, that each block in `src` will have their own `dist` built by webpack and we won't need to copy anything. `src/block/xyz` will literally be a fully functional WP plugin

-----

### All Submissions:

Closes # .

1.
2.
3.

### Other information:


obenland:
> I think I'd like to change the approach in a way, that each block in src will have their own dist built by webpack and we won't need to copy anything. src/block/xyz will literally be a fully functional WP plugin

Can you help me understand the ideal state for you? Would every block directory contain a root php file and readme file? Would those be removed from a regular news pack build?

getdave:
> src/block/xyz will literally be a fully functional WP plugin

We should note that many of these Blocks expect and require modifications to existing Core Blocks to function correctly. For example, the [Group Block config and styles get modified here](https://github.com/Automattic/newspack-blocks/tree/70adb0186055d4c9ed5dc30050f9fb77562fa275/src/block-styles/core/group). Therefore your goal of self-contained Plugin-per-Block would need to include these mods as well.

marekhrabe:
> Can you help me understand the ideal state for you? Would every block directory contain a root php file and readme file? Would those be removed from a regular news pack build?

Yeah, nothing in stone but that's what I'm thinking might work the best. You could zip up any of those block dirs and be able to upload them to WP install as plugins. They will have a readme, main PHP file, source JS as well as build assets. Literally everything they need to run them.

At the same time, you can run the build script (the one already present) to generate the combined plugin which ships all the blocks + all the other stuff in this repo (some core block modifications).

If you install the combined plugin and a single block at the same WP instance, I think the single block should take a preference and get used (maybe we can define some constants to do this easily?)

getdave:
Also note that you could use Webpack to copy everything to your target directory, thereby keeping everything "build" related within Webpack.

So that would be:

1. Compile and copy over all JS for the Block
2. Copy over all the PHP for the Block (using Webpack Copy Plugin).
3. Potentially do a inlining of constants within the PHP files using WebPack Define Plugin.
4. Build README files...etc

Just a thought...

Also I thought that we could use `apply_filter` to allow us to modify the wording used within the Block PHP code. We could then bundle a custom PHP file in our Dotcom build which would `add_filter`s to customise the output as we require.

Issue #106, opened 2019-09-04
Article Block: 5-6 columns don't look great


laurelfulford:
This came up in the review of #105: 

The 5 and 6 column layouts for the article blocks don't look great:

![image](https://user-images.githubusercontent.com/177561/64276878-d0677680-cefd-11e9-8ea8-159a59a0c419.png)

Some options to improve them could be:

* Changing the breakpoint for when they drop to 2-3 columns from tablet to desktop.
* Doing something about the font sizing so they don't use larger font sizes (though my worry with this one is that it could be confusing, especially since there is a separate font size control. But we might be able to set a different 'starting' size that the font size control then scales up and down, but you can't get the same font size in the 6 column layout as you can with a 1 column layout).

laurelfulford:
Something else that came up related to these column counts (and perhaps the four column option as well): 

These stories are small even on desktop; changing it to a single column on mobile makes them quite large, and really changes the visual hierarchy (especially the image size). We need them to 'break down' in a more appropriate way -- like stick to two columns, even on small screens.

thomasguillot:
How about:

* We allow the block to display wide or full width
* If wide, then you get the 5 columns layout
* If full-width, you get the 5 && 6 columns layout
* If "normal" width, then we limit it to 4 columns

On tablet:
* 5 columns can translate to a row of 2 then a row of 3
* 6 columns -> 3, 3

On mobile:
* 5 columns -> 1, 2, 2
* 6 columns -> 2, 2, 2

dkoo:
@laurelfulford is this still an issue? I think we adopted much of what you suggested already, right? Feel free to reopen if there's more to do.

PR #175, opened 2019-10-19
Article Blocks: Remove bottom border of last item

laurelfulford:
### All Submissions:

In an attempt to create some predictable behaviour, I changed the article block to include a border under the last item. Unfortunately, this leads to a lot of doubled borders, especially on smaller screens, or at the very bottom of the page.

I think it ultimately makes more sense to remove the bottom border from the last item; the only time it seems to make sense to keep it is if you only have the block set to show one article; if it doesn't show on the last article in that case, it won't show at all.

1. Set up three article blocks: one with one article, one with several articles displayed as a list.
2. Apply the 'borders' style to all three.
3. Confirm that the last article in each block does not have a bottom border, unless it is the block only displaying one article. 

Note: I noticed some screwy behaviour with the borders and grid styles when testing this change -- the borders move to the bottom of the articles before they actually break into separate columns. I can look at resolving this in a separate PR; it could also change again down the road, depending on what we go with for #106. 

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/view.scss
+++ b/src/blocks/homepage-articles/view.scss
@@ -502,6 +502,10 @@
 
 		&:last-of-type {
 			margin-bottom: 0;
+
+			&:not(:first-of-type) {
+				border-bottom: 0;
+			}
 		}
 	}
 


thomasguillot:
Since the new columns have been merged, the styles should be using `:after` and not border anymore

laurelfulford:
@thomasguillot That looks like screenshots of the columns block, not the article blocks -- is that right?

I did double-check the article block styles with Style 3, though, and the borders aren't right (they should be dotted!), so I'll address that in another PR there:

image

Just let me know if I'm missing something above!

thomasguillot:
> @thomasguillot That looks like screenshots of the columns block, not the article blocks -- is that right?

🤦‍♂ 

Now I understand why I couldn't see the changes... 🤪

laurelfulford:
Thanks Thomas!

> Now I understand why I couldn't see the changes... 

Haha! We have too many border things in this plugin... 😆

Issue #164, opened 2019-10-14
Article Block: Category placement option?


laurelfulford:
Right now the category is displayed over the image, when the image is at the top, but we've had a request to move it above the image in that case instead.

Would it make sense to add an option to the block for this? Change the default styles? Or make it something that changes depending on style pack?

laurelfulford:
A second publication has also asked that the category be moved.

I think it might make sense to just move it out of the image area, and potentially re-add it later if it's requested.

laurelfulford:
The category has been moved, and the reception seems positive; I'm going to close this one for now, until we have a request to move it again.

PR #178, opened 2019-10-21
Do not overlap the category over the image.

laurelfulford:
### All Submissions:

This PR removes the styles that place the category on top of the image, when you opt to display it on blocks. 

We've had two requests to move it so far; I suggested adding an option to move it in #164, but if we do it seems like having it below the image by default would be preferable. 

I'm starting that process here, but recognize this is something pretty distinct that is now available on staging sites. So I'd be open to discussing how to approach this further, what the 'default' should be and if we should make something changeable on the block. 

1. Add a an article block to a page; make sure at least one of the articles displayed has an image.
2. Set the block to display categories.
3. View on the front-end; note the category placement is different:

image

4. Apply the PR and run `npm run build:webpack`
5. Confirm that the category now displays in the same spot when the block has an image, and when it does not:

image

I'm going to make a complementary theme PR for this as well, since I know some things in the style packs will need tweaking.

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/view.scss
+++ b/src/blocks/homepage-articles/view.scss
@@ -170,19 +170,6 @@
 		}
 	}
 
-	&.image-aligntop .post-has-image .cat-links {
-		background-color: #111;
-		left: 0;
-		padding: 0.3em 0.5em 0.25em;
-		position: absolute;
-		top: 0;
-
-		a {
-			color: #fff;
-			display: inline-block;
-		}
-	}
-
 	.entry-meta {
 		display: flex;
 		flex-wrap: wrap;


claudiulodro (approved):
I think this is going to be a more widely-accepted way of displaying the categories. I don't know that we need the style where they are displayed in the top-left corner; unless someone specifically requests those I wouldn't worry about making it a customizable option (it's more combinations to keep track of for styling, and we have enough I think!)

Two visual things I noticed that should probably be updated in the theme after this is merged:

With Style 1, on posts with images there is a long strip of background color:
Screen Shot 2019-10-22 at 10 02 12 AM

On Style 3 with the latest PR you did to bring back the overlap styles, the category is hidden underneath the title:
Screen Shot 2019-10-22 at 10 02 39 AM

The other style packs look fine IMO.

laurelfulford:
Thanks @claudiulodro! 

I'll merge this as is, and follow up with a separate theme PR to fix the fun new stuff there :)

Issue #153, opened 2019-10-09
Article Block: Image as background + Vertical spacing


laurelfulford:
https://github.com/Automattic/newspack-blocks/pull/73 Adds a first pass at being able to set the featured image as a background in the article blocks.

Originally the PR included a control for setting the padding on top of the content, so you could adjust the overall width. 

We removed it to simplify things, and to discuss how to best handle this further. 

The problem that needs solving is, generally, you'd want to be able to control "how much" image to show, based on circumstances. However, is there a smart way we can do this automatically depending on where the block is situation/its size? 

Is there a better way to display distance if we have to use a control (percentages, or pixels like the cover block)? 

This issue is to explore and discuss the best next steps.

thomasguillot:
I think we should bring the settings from the Cover Block (Fixed Background, Focal Point Picker) and some from the Hero Block by CoBlocks (Padding, Fullscreen, Height)

<img width="277" alt="Screenshot 2019-10-11 at 17 20 32" src="https://user-images.githubusercontent.com/177929/66668032-3efdc980-ec4c-11e9-87ba-ad6aca8129ec.png">

<img width="277" alt="Screenshot 2019-10-11 at 17 24 27" src="https://user-images.githubusercontent.com/177929/66668113-5ccb2e80-ec4c-11e9-91ae-2132ef728922.png">

laurelfulford:
> I think we should bring the settings from the Cover Block (Fixed Background, Focal Point Picker) 

Focal Point Picker especially would be handy! I think it would need to be a post-level option, though, since the focal point would probably vary from featured image to featured image. Does that make sense? 

> ... and some from the Hero Block by CoBlocks (Padding, Fullscreen, Height) 

I like the use of things like 'huge' for padding :) For the height, do you think the pixels could be be confusing when scaling it down for smaller screens (instead of vague numbers?). Or am I overthinking this :D 

The height/padding question is definitely the bit we need to solve first, to get the option to be more useful. I think these other options would help take it to another level!

thomasguillot:
> Focal Point Picker especially would be handy! I think it would need to be a post-level option, though, since the focal point would probably vary from featured image to featured image. Does that make sense?

Hmm, I didn't think of that... If we display 1 Focal Point Picker per post in the sidebar it might get a bit crowded 😕, unless we can select a post (like a child block) and then the setting appears in the sidebar

[...]

PR #196, opened 2019-10-31
WIP Article Block Height: Add input for setting `min-height` (option A)

laurelfulford:
### All Submissions:

This PR adds a 'Minimum height in pixels' field to the article block, when you have the image positioned in the background. 

It's a starting point for some of the ideas suggested in #153; two other PRs (#196 and #197) illustrate a couple other approaches we could look at. 

The main issue I see with this approach could be mobile: in looking at CoBlock's Hero block, it lets you set a mobile and tablet min-height in pixels separately. However, the way they override the inline desktop height is by adding `!important`, which we can't use with AMP. 

I think this is something we need to resolve to some extent before merging -- a 1000px tall article on desktop may make sense, and then look terrible on mobile. But I wanted to put this out there first to see how it feels :) 

3. Set the article block to use the `Show media behind` image position.
4. Confirm you now have a field in the Featured Image Settings labelled 'Minimum height in pixels':

image

6. Make sure the minimum height isn't applied to an article in the set that doesn't have a featured image.
7. Try hiding the featured image; make sure the 'Minimum height' option is hidden, and the height is no longer applied.
8. Try changing the image positioning to top, left, or right, and make sure the 'Minimum height' option is hidden, and the height is no longer applied. 
9. Give some feedback on how to approach this for mobile (please and thank you!). 

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/edit.js
+++ b/src/blocks/homepage-articles/edit.js
@@ -15,12 +15,13 @@ import moment from 'moment';
  * WordPress dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { Component, Fragment, RawHTML } from '@wordpress/element';
+import { Component, Fragment, RawHTML, useState, useCallback } from '@wordpress/element';
 import { InspectorControls, RichText, BlockControls } from '@wordpress/editor';
 import {
 	PanelBody,
 	PanelRow,
 	RangeControl,
+	TextControl,
 	Toolbar,
 	ToggleControl,
 	Dashicon,
@@ -31,7 +32,7 @@ import {
 	SVG,
 } from '@wordpress/components';
 import { withSelect } from '@wordpress/data';
-import { compose } from '@wordpress/compose';
+import { compose, withInstanceId } from '@wordpress/compose';
 import apiFetch from '@wordpress/api-fetch';
 import { addQueryArgs } from '@wordpress/url';
 import { decodeEntities } from '@wordpress/html-entities';
@@ -41,6 +42,7 @@ import { PanelColorSettings, withColors } from '@wordpress/block-editor';
  * Module Constants
  */
 const MAX_POSTS_COLUMNS = 6;
+const ARTICLE_MIN_HEIGHT = 100;
 
 /* From https://material.io/tools/icons */
 const landscapeIcon = (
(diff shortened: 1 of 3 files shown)


thomasguillot (commented):
I feel like this option doesn't offer as much flexibility as the other two, especially on mobile. If you set the min-height to 1000px, then it looks huge on mobile whereas the vh approach looks a bit more balanced.

Maybe adding options for mobile/tablet would make it more flexible, but is it worth the effort?

Nov-01-2019 16-32-14

thomasguillot:
If we go ahead with this. Icons that I would suggest:

* Desktop: https://fonts.gstatic.com/s/i/materialicons/desktop_windows/v1/24px.svg
* Tablet: https://fonts.gstatic.com/s/i/materialicons/tablet/v1/24px.svg
* Mobile: https://fonts.gstatic.com/s/i/materialicons/smartphone/v1/24px.svg

laurelfulford:
Yeah, I definitely agree that it's less flexible than the other two.

The issue I ran into offering three heights -- desktop, mobile and tablet -- is more _how_ to add the styles. CoBlocks adds the min-height inline for the desktop styles, but if you add a mobile and tablet height, it'll generate a unique CSS class and those styles to the `head`, like:

```<style id='coblocks-frontend-inline-css' type='text/css'>
@media only screen and (max-width: 768px) {
        .coblocks-hero-101111828337 > [class*="__inner"]:not(.is-fullscreen) {
                min-height:300px !important
        }
}

@media only screen and (max-width: 514px) {
        .coblocks-hero-101111828337 > [class*="__inner"]:not(.is-fullscreen)  {
                min-height:400px !important
        }
}
</style>```

Since AMP ignores `!important`, we wouldn't be able to do exactly this; we'd probably have to always generate unique classes and add styles for all three to the `head`. So yeah, definitely more effort; it might be best to start with one of the other approaches, and if there's a call for this level of control, build on top of that.

thomasguillot:
Ok, thank you for the explanation @laurelfulford. I've only realised you already mentioned this in the description of the PR  🤦‍♂ 
Alright so, I think we should disregard this option.

laurelfulford:
Haha it's okay! That comment was buried in multiple paragraphs, I should've highlighted it more as a concern. 

Closing this approach due to lack of flexibility for different sized screens.

PR #197, opened 2019-10-31
WIP Article Block Height: Add input for setting `min-height` (option B)

laurelfulford:
### All Submissions:

This is a riff on #196, switching the text input for pixels with a range control for `vh` to set the minimum height for articles with the image as background.

The hope with using a measurement relative to the screensize is that it will behave a bit more predictably on small screens (compared to pixels, which are fixed regardless of screensize). I'm not 100% `vh` is the best option, but figured it would give a rough feel for how this kind of thing would behave. 

See #153.

4. Confirm you now have a field in the Featured Image Settings labelled 'Minimum height':

image

6. Add a block with a taller minimum height (60-70). On the front-end, try adjusting the height of the browser window and confirm that the minimum height on the block visually adjusts too.

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/edit.js
+++ b/src/blocks/homepage-articles/edit.js
@@ -84,7 +84,8 @@ class Edit extends Component {
 		const {
 			showImage,
 			imageShape,
-			imageFileSize,
+			mediaPosition,
+			minHeight,
 			showCaption,
 			showExcerpt,
 			showAuthor,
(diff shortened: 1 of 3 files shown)


thomasguillot (approved):
I really like this option. As a non-technical user, I think this would make more sense than the "padding-top" option ("what's a padding on a website???")

laurelfulford:
> As a non-technical user, I think this would make more sense than the "padding-top" option

I agree! 

My one issue with this approach is that the added height can completely disappear on smaller screens (when the `min-height` becomes less than the height the article would be without it). But I _think_ I may be able to do a mash-up of the two -- where the one control adding a 'min-height' also adds some padding. I'm going to experiment with that in a separate PR, and if it works, I think it'd be the best approach.

If not: I'm still kind of leaning on the 'padding' approach in #198 since it'll be added at any screen size, but I agree that the label is kinda terrible. Since the value is pretty arbitrarily to the average user, too, do you think we could get away with calling that 'height' instead?

laurelfulford:
Closing this PR in favour of going forward with #199.

PR #198, opened 2019-10-31
WIP Article Block Height: Add range control for top padding (option C)

laurelfulford:
### All Submissions:

This is another riff on #196 and #197, but rather than setting a minimum height, it sets a top padding.

It's similar to #197 in that it uses a measurement that changes with the viewport (in this case, `vw`, to better get a feel for how much that shifting affects things). 

One plus of this approach, over a minimum height, is that it doesn't fully disappear on smaller screens (even if the padding value is relatively smaller, it's always a value on top of the current height, vs. the minimum height becoming shorter than the actual height of the article, based on the content height.

See #153 

1. Apply the PR and run npm run build:webpack
4. Confirm you now have a field in the Featured Image Settings labelled 'Top Padding', with a default of `0`:

image

6. Add a block with a taller top padding (40-50). On the front-end, try adjusting the width of the browser window and confirm that the padding on the block visually adjusts too.

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/view.php
+++ b/src/blocks/homepage-articles/view.php
@@ -108,9 +108,14 @@ function newspack_blocks_render_block_homepage_articles( $attributes ) {
 				}
 				$newspack_blocks_post_id[ get_the_ID() ] = true;
 				$post_counter++;
+
+				$styles = '';
+				if ( 'behind' === $attributes['mediaPosition'] && $attributes['showImage'] && has_post_thumbnail() ) {
+					$styles = 'padding-top: ' . $attributes['paddingTop'] . 'vw';
+				}
 				?>
 
-				<article <?php echo has_post_thumbnail() ? 'class="post-has-image"' : ''; ?>>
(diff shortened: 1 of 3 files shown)


src/blocks/homepage-articles/view.php
```diff
@@ -108,9 +108,14 @@ function newspack_blocks_render_block_homepage_articles( $attributes ) {
 				}
 				$newspack_blocks_post_id[ get_the_ID() ] = true;
 				$post_counter++;
+
+				$styles = '';
+				if ( 'behind' === $attributes['mediaPosition'] && $attributes['showImage'] && has_post_thumbnail() ) {
+					$styles = 'padding-top: ' . $attributes['paddingTop'] . 'vw';
```
thomasguillot: Should it be `vh` instead?

thomasguillot (changes requested):
I really like this option and overall it works as expected.

src/blocks/homepage-articles/view.php
laurelfulford: Ah, thanks Thomas! I was experimenting with the `vw` but `vh` in this case really makes more sense. Updated in aeaea27.

laurelfulford:
Closing this PR in favour of going forward with #199.

PR #199, opened 2019-11-01
Article Block Height: Add range control for minimum height + top padding (option D)

laurelfulford:
### All Submissions:

This PR is kind of a combination of #197 and #198 -- it adds a 'min-height' control to the article block when you have the image set to display behind the article. But rather than just adding a minimum height, it also adds 20% of that value as a top padding. 

This is to help make sure the blocks don't lose all top spacing on smaller screens -- I was finding when you added a moderate min-height to an article with an excerpt, as it scaled down for smaller screens, it would often be 'shorter' than the article's new height, making it seem like it wasn't applied at all. 

The padding won't add anything when the min-height is taller than the article would normally be, but adds a bit of top padding when the min-height is shorter than the content. 

See #153 .

4. Confirm you now have a field in the Featured Image Settings labelled 'Minimum height':

image

6. Add a block with a taller minimum height (60-70). On the front-end, try adjusting the height of the browser window and confirm that the minimum height on the block visually adjusts too.
10. In an article block showing the excerpt, set the minimum height just tall enough that it adds a bit of space on the top. Scale it down on the front-end, and make sure that there's always a bit more padding on the top than bottom:

image

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/edit.js
+++ b/src/blocks/homepage-articles/edit.js
@@ -84,7 +84,8 @@ class Edit extends Component {
 		const {
 			showImage,
 			imageShape,
-			imageFileSize,
+			mediaPosition,
+			minHeight,
 			showCaption,
 			showExcerpt,
 			showAuthor,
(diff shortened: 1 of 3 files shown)


laurelfulford:
Now that I got this working, I'm wondering it this will now cause more problems, if we ever decide to add a 'align to top, middle, bottom' option for the article content 😆 

Anyway: feedback on this approach and others is appreciated!

thomasguillot:
Overall, it's a nice combination of options B and C.

> Now that I got this working, I'm wondering it this will now cause more problems, if we ever decide to add a 'align to top, middle, bottom' option for the article content 😆

Yeah, this is my concern with the padding approach. BUT we could always:
* Padding-Bottom, if align top
* Padding-Top, if align bottom
* 1/2 Padding-Bottom and 1/2 Padding-Top, if align middle

laurelfulford:
Thanks Thomas!

That's a good point re: moving the padding around. It seems like the easiest way to 'fix' the problem of seeming to 'lose' the minimum height, and it would be pretty straight-forward to adjust where the padding was added if we added alignment options.

I'm going to close the other PRs in favour of this one. 

@jeffersonrabb I know you had some reservations about the very early, CSS-heavy approach to this (multiple CSS classes, and a good chunk of CSS to go with it). Does this seem like a better approach to you?

laurelfulford:
> My only suggestion would be to add some help text to the range control explaining what it does and how it should be used.

Oh, good call Jeff! I'm going to commit this one as is, and address that in a second PR.

laurelfulford:
PR with help text has been added in #205.

PR #202, opened 2019-11-08
Article Block: Add Co-Author Plus plugin support

### LINKED ISSUE #200: Article Block: Add support for CoAuthor Plus plugin
Is it possible to add support for the Co Author Plus plugin to this block?

This could be:

* Displaying multiple authors on the front-end/in the editor.
* Being able to search for posts by Co Author Plus Guest authors (instead of just the standard WordPress users). 

I'm not sure how complex this will be, and whether it's going to require a lot of really specific code. Would be interested in getting a second opinion!

laurelfulford:
### All Submissions:

This PR adds support for the Co-Author Plus plugin to the Article block.

I also removed the link to the author archives from what we're making available to the API, since we don't actually use it in the editor anymore. 

Closes #200.

2. If you haven't already, install and activate the [Co-Authors Plus](https://wordpress.org/plugins/co-authors-plus/) plugin.

3. Create one 'Guest Author'.

4. On one article, add a second author; on a second article, add two. Make sure one is the Guest author and the other is a standard WordPress user. This gives us a good mix of different author types and numbers.

5. On a page, add the article block, and adjust the settings to make sure the above two posts are displayed, plus one with one regular author.
6. View in the editor and confirm:
     a. Each author's display name and avatar are correct.
     b. Two authors are separated by 'and'; three plus authors are separated by commas, with the last two authors separated by 'and'
image
image
     c. Scale down the block's Type Scale, and confirm the avatars successfully scale down.
image
     d. Hide the author avatar and confirm it works as expected.
     e. Hide the author completely and confirm it works as expected.
7. View the block on the front end, and run through the same a-e checks from step 6.
8. Deactivate the Co-Author Plus plugin and confirm things display the same as they did originally in the editor and on the front-end.

### Other information:

### DIFF
--- a/class-newspack-blocks-api.php
+++ b/class-newspack-blocks-api.php
@@ -130,11 +130,23 @@ class Newspack_Blocks_API {
 			$authors = get_coauthors();
 
 			foreach ( $authors as $author ) {
+				// Check if this is a guest author post type.
+				if ( 'guest-author' === get_post_type( $author->ID ) ) {
+					// If yes, make sure the author actually has an avatar set; otherwise, coauthors_get_avatar returns a featured image.
+					if ( get_post_thumbnail_id( $author->ID ) ) {
+						$author_avatar = coauthors_get_avatar( $author, 48 );
+					} else {
+						// If there is no avatar, force it to return the current fallback image.
+						$author_avatar = get_avatar( ' ' );
+					}
+				} else {
+					$author_avatar = coauthors_get_avatar( $author, 48 );
+				}
 				$author_data[] = array(
 					/* Get the author name */
 					'display_name' => esc_html( $author->display_name ),
 					/* Get the author avatar */
-					'avatar'       => wp_kses_post( coauthors_get_avatar( $author, 48 ) ),
+					'avatar'       => wp_kses_post( $author_avatar ),
 				);
 			}
 		else :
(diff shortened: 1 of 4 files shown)


laurelfulford:
> Also, seems to me the Avatars should link to the Author pages?

Good call! I created a separate issue for this in #206, since it's a wider issue than just with the co-author plus code.

laurelfulford:
> The only issue I found occurs when you create a guest author with no Avatar and add it to a post.

It looks like this is happening on the single posts too, due to the theme code 😕 -- it's trying to pick up the featured image from the current post. 

I made an issue for the theme here: https://github.com/Automattic/newspack-theme/issues/567

laurelfulford:
> The author shows with an Avatar (randomly selected?) but the image is un-cropped so the avatar doesn't necessarily fill the full space. Not sure if this is a Co-Authors issue or something we can address, but we probably shouldn't show an Avatar for a guest author that doesn't have one set.

I reworked the code to address this in 2e3e6a7. It's a little bit hacky: I added a check to make sure that, if the current author is a Guest Author, it has a featured image associated with it, and if it doesn't, to use whatever is set as WordPress's default avatar. 

We could also switch it to return nothing, but I think in a list of 3+ authors, it won't really be clear which of the two images go with which of the three names (unless we pair them image + name?)

laurelfulford:
Thanks for all the great code feedback, @jeffersonrabb! I've made the updates, though the code I added to fix the random images may also need a bit of polish.

I did retest the code with these changes and everything worked well for me, but I would appreciate your eyes on this a second time. Thanks!

laurelfulford:
Thanks!

PR #214, opened 2019-11-15
Add autocomplete search to article block story select

claudiulodro:
### All Submissions:

I wanted to get this in before the work to bring the articles block to wpcom gets too far along, because this introduces a backwards-incompatible change.

This PR changes up the way the specific stories selection works. Instead of picking one post from a select dropdown of the last 50 posts, the user can pick up to 100 posts using autocomplete search:

Screen Shot 2019-11-15 at 6 56 28 AM

1. Add an articles block, toggle "Choose specific stories" on, and select some stories. The order of the stories in the block should match the order of the stories selected. Save and view on frontend. Verify things look the same, and the stories you requested are available in the correct order.
2. Verify tag/author/category search stuff still works on homepage blocks, and carousel block still works.

### Other information:


src/blocks/homepage-articles/view.php
```diff
@@ -17,21 +17,23 @@ function newspack_blocks_render_block_homepage_articles( $attributes ) {
...
+	$posts_to_show  = intval( $attributes['postsToShow'] );
+	$specific_mode  = intval( $attributes['specificMode'] );
+	$args           = array(
 		'post_status'         => 'publish',
 		'suppress_filters'    => false,
 		'ignore_sticky_posts' => true,
 	);
-	if ( $single_mode ) {
-		$args['p'] = $single;
+	if ( $specific_mode && $specific_posts ) {
```
jeffersonrabb: Wondering if we should check that `$specific_posts` has items to avoid making the query if empty?

jeffersonrabb (approved):
This is great! In addition to improving the UI and not limiting to the 50 most recent, the option to pick multiple stories is a powerful addition, plus the code reorganization sets up `QueryControls` nicely for re-use. Everything I have is minor and non-blocking so I'm approving, but these might be nice improvements.

1) As hinted at in the PR description, this work will cause existing blocks using the single story selection tool to revert to showing a group of recent posts. I'm not sure how many we have in production, but we should discuss how this breakage will be  handled. If the number if very low (as I suspect it is) we could simply fix them up manually. If higher, we  could add deprecation logic to handle the upgrade (https://developer.wordpress.org/block-editor/developers/block-api/block-deprecation/), but this probably just adds unnecessary complexity and code weight.

2) The next three notes relate to the empty state of the posts UI. When `Choose specific stories` is first switched on, the block remains unchanged with no loading state. 

3) If posts are added and then removed, leaving the Posts field empty, the block gets stuck in a loading state and the console shows  this error `Uncaught (in promise) {code: "rest_orderby_include_missing_include", message: "You need to define an include parameter to order by include.", data: {…}}`

4) After adding posts, if `Choose specific stories` is switched off, the block returns to initial state (as expected). If switched on again it goes into the perpetual loading state, which differs from the behavior in item 2. 

5) In the 'give a mouse a cookie' category, being able to pick multiple stories made me want the option to reorder by drag and drop. It's easy enough to change the order by deleting and re-entering the posts so this feels very inessential, but I'm just noting the thought as a future nice-to-have.

src/blocks/homepage-articles/view.php
claudiulodro: For sure, that should also solve points 3 and 4 from above I believe. https://github.com/Automattic/newspack-blocks/pull/214/commits/99a163df0632ca36487fb4a297df009f4a8eb76a

src/blocks/carousel/edit.js
```diff
@@ -4,7 +4,6 @@
 import QueryControls from '../homepage-articles/query-controls';
 import createSwiper from './create-swiper';
 import classnames from 'classnames';
-import AutocompleteTokenField from '../homepage-articles/components/autocomplete-tokenfield.js';
```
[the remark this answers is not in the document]
claudiulodro: Good call. That's nicer. https://github.com/Automattic/newspack-blocks/pull/214/commits/54f43df068acbe35d9f1834a23832188db6a1c19

claudiulodro:
> As hinted at in the PR description, this work will cause existing blocks using the single story selection tool to revert to showing a group of recent posts. I'm not sure how many we have in production, but we should discuss how this breakage will be handled. If the number if very low (as I suspect it is) we could simply fix them up manually.

Looking through the launched sites' homepages, there is only one site that uses a specific story for the featured post. Let's just manually update it when this is deployed, and we should be fine.

claudiulodro:
I've created an issue for the remaining nice-to-have things you've mentioned: https://github.com/Automattic/newspack-blocks/issues/218

Thanks for the review!

PR #215, opened 2019-11-18
Only display the Featured Image on the Right or Left on devices with larger screens

thomasguillot:
### All Submissions:

Instead of splitting the screen in 2 on small screens, we only do this when breakpoint is >= 600px that way we avoid having an image with just a little bit of next next to it.

_Attempt to close https://github.com/Automattic/newspack-issues/issues/103_

__Before:__

1 before-iphone

__After:__

2 after-iphone

1. On a page, add the Articles Block and add a bunch of posts with the featured image either right or left. Check on mobile/small screen.
2. Switch to this branch and run `npm run build:webpack`
3. Now check the same page on mobile/small screen -- Does it look better?

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/view.scss
+++ b/src/blocks/homepage-articles/view.scss
@@ -90,60 +90,62 @@
 		}
 	}
 
-	&.image-alignleft,
-	&.image-alignright {
-		.post-has-image {
-			display: flex;
+	@include media( mobile ) {
+		&.image-alignleft,
+		&.image-alignright {
+			.post-has-image {
+				display: flex;
 
-			.post-thumbnail,
-			.entry-wrapper {
-				flex-basis: 50%;
+				.post-thumbnail,
+				.entry-wrapper {
+					flex-basis: 50%;
+				}
+			}
+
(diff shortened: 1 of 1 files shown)


laurelfulford:
I made a proof of concept of the 'changing the image/text balance' suggestion above in #216.

Looking at Automattic/newspack-issues#103, we may also be able to improve this by increasing the smaller font sizes on mobile.

My hope is to minimize the amount of changes we have to make to improve the appearance of the block in this case, but I understand we may have to go with something more drastic :)

thomasguillot:
@laurelfulford I like what you did in #216 and that's definitely a good solution. How about we go ahead with #216 but add an extra option like the one I did here for mobile devices and have a "stack on mobile" setting (turned off by default)

e.g. Media & Text:

Screenshot 2019-11-19 at 10 57 42

laurelfulford:
@thomasguillot that's a great idea -- best of both worlds! :) I'll add that now.

thomasguillot:
Closing this. #216 is much, much better 😊

PR #217, opened 2019-11-19
WIP: Bring back More button

### LINKED ISSUE #154: (3P) Article Block: Bring back the more link
We originally removed the optional 'More' button from the Article block in #94 because of the block's increasing complexity: the button was added when you could just pick articles by category, and it either linked to that category's archive, or, if you were showing all posts, to the blog posts page. 

As we added more query option (tags, authors, a mix of those plus categories, or a single specific post), it made predicting where the more button should go more complicated.

Charter publications also brought up the fact that the button showed you posts that you likely already saw, and that it would be better if it could somehow offset that, so you would link to articles that you hadn't seen (even if it was just an anchor). 

Ultimately, we decided to remove the button rather than have it there and working illogically. As a short-term fix, the button block is our recommended work-around. 

Despite its removal, the button is still valuable. This issue is to hammer out a good way to bring it back. We may not be able to solve all of the issues above, but as long as we can make sure it's linking to a logical spot, I think that's a solid, helpful start.

obenland:
First pass, essentially reverting #102

### All Submissions:

Closes #154.

1.
2.
3.

### Other information:


obenland:
I preliminarily tried to reuse Jetpack's infinite scroll module for this. While I don't believe using the feature directly will help us, I do think it would be work exploring the possibility of lifting some of that code and condense it down to fit our purposes here.

@WunderBart Is that something you think you could investigate further?

/cc @marekhrabe

marekhrabe:
That sounds good! An infinite pagination (based on a button to paginate) sounds like a very nice addition to the block. 

What I see us doing is to:

1. Check the Homepage Articles block code from editor and see if we can use the same code or api for rendering posts for the frontend. Considering post listing already works in the editor, I assume the code is finished and we will be able to reuse it for free 🤞
2. Look at Jetpack infinite scroll but I wouldn't recommend to reuse it verbatim. It's meant to work only on a single instance in a page (while we can have multiple block) and is written a single monolith file. We have the advantage of build system here so I imagine we can make a clean and simple version of the post loading just for this block.

WunderBart:
Thanks for starting this draft @obenland and for the tips, @marekhrabe. Starting to investigate right now!

WunderBart:
I think we should always validate if there actually are more posts available and update button accordingly, e.g. hide it or change its title.

There are 2 situations in which we should perform that check:

1. On initial render
2. On each button click

Validation scenario:
   1. Retrieved number of posts is less than the defined value (no need for an additional call).
   2. If the above is not true (the retrieved number is equal to the target number), perform an additional call and update the button if necessary.

IMO, we should just hide the button rather than update its title and let the user know that it's the desired behavior, e.g. by adding a note by the `Show "more" button` option toggle. Something like:



Does that make sense? What do you think?

getdave:
## Scope

Looking at this I feel the first pass should be aiming at the MVP solution. For me this is:

> As a user on the front of the website, I should see a "more button" in order that I can load more posts on demand

As a result, I would vote that we avoid complexities such as infinite scroll (which [can be bad for UX anyway](https://www.nngroup.com/articles/infinite-scrolling/) - how do you reach the footer on an infinite scroll site?) at least for a first pass.

I'm really not sure where the "Infinite Scroll" requirement has come from. If anyone can point to where this change of direction occurred I'd be grateful as I feel I might have missed something. If not then the goal is "Bring back More button" so I feel we should proceed with this.

## Technical Implementation

### AMP
We need to consider AMP here. Some quick digging reveals AMP has a ["Show More" component ](https://amp.dev/documentation/examples/interactivity-dynamic-content/show_more_button/?referrer=ampbyexample.com)for this.

From the AMP website:

> "Show more" is common design pattern used on e-commerce category pages to lazy load more content triggered by an user interaction.

You define a template inline and that gets used to render the additionally loaded posts:

```html
<template type="amp-mustache">
    <div class="list-item">
      <amp-img src="{{imageUrl}}"
        width="100"
        height="100">
      </amp-img>
      <p>{{title}}</p>
    </div>
  </template>
```

You also tell AMP where to get the "next" items.

### Standards Approach

TBC




Here's an alternative technique I've used in the past for handling "Load More".

WunderBart:
Thanks for helping with this, @getdave.

> The main benefit of this approach is that it keeps the rendering of an article consistent. 

Agree. Using the same PHP handler for view side as you described sounds like best option here.

Just wanted to point out that this component already has 2 sources of truth for post rendering - one from the PHP handler and another one for the [edit block](https://github.com/Automattic/newspack-blocks/blob/master/src/blocks/homepage-articles/edit.js#L111-L172). I'm not saying we should address it in this PR, just something to keep in mind. 

In [my first commit](https://github.com/Automattic/newspack-blocks/pull/217/commits/70e6042a1ddf95ac518345418abfc91323c43b7a) here I aimed to use the post template from the editor, but I'm gonna revert it as we're not gonna be using it.

package.json
```diff
@@ -27,7 +27,9 @@
     "swiper": "4.5.1"
   },
   "scripts": {
-    "build:webpack": "calypso-build --config='./webpack.config.js'",
+    "start": "npm run dev:webpack",
+    "dev:webpack": "calypso-build --watch",
+    "build:webpack": "calypso-build --watch --config='./webpack.config.js'",
```
WunderBart: Do you think we should move it to a separate PR?

getdave: Yes we should.

getdave:
> Just wanted to point out that this component already has 2 sources of truth for post rendering - one from the PHP handler and another one for the edit block. I

@WunderBart I understand and thanks for reminding me.

I think what we have is

1. Frontend rendering of Article on initial load.
2. Frontend rendering of Article on "More" button click
3. Editor rendering of Article.

At this point, I'm most concerned with keeping 1 and 2 in sync and have a single source of truth. 

It is very important to keep 3 in sync but less so at this point given that it is already not a single source of truth with the front end (as with most Blocks).

getdave:
Note: i've removed the build script changes and put them in https://github.com/Automattic/newspack-blocks/pull/224

WunderBart:
At this point, we're able to render HTML from `/wp-json/wp/v2/newspack-articles-block/articles` endpoint into the [`amp-list` template](https://github.com/Automattic/newspack-blocks/blob/0e1a8d8f882f5200b664f74ae88853b283c97686/src/blocks/homepage-articles/view.php#L315) which should allow us to utilize existing PHP render function.

Testing steps:
- Add the Homepage Articles Block
- Turn the `Show "More" Button` option on in block options
- Go to preview
- Scroll down to the very bottom
- Click "More" button repeatedly to see more items loaded

⚠️ If you can't see the `More` button at all, you'll probably need to manually replace `http` with `https` in the preview URL and refresh the page. It should _just_ work when using `localhost`. If you're using Flywheel you can trust the certificate which will enable `https` for your local server:

getdave:
I rebased this to get the latest dev scripts updates.

getdave:
Worked has move to https://github.com/Automattic/newspack-blocks/pull/226

PR #226, opened 2019-11-25
Restore the "Load more" button - no AMP support

### LINKED ISSUE #154: (3P) Article Block: Bring back the more link

WunderBart:
Closes https://github.com/Automattic/newspack-blocks/issues/154

This is a rebased version of https://github.com/Automattic/newspack-blocks/pull/217 with removed commits that introduced the Mustache template. We've decided to continue the approach of reusing existing PHP render function and compile post's HTML into the newly created endpoint response, as planned originally.


WunderBart:
I'm having a hard time trying to use the correct layout for the [`amp-list`](https://amp.dev/documentation/examples/components/amp-list/?format=websites) element so the initial content is always rendered at full height - no fixed height, no overflow, etc. The only layout that supports that is `container`, but it's not supported as an initial layout (?). It can be switched to via the `is-layout-container` attribute, but it has to use [`amp-bind`](https://amp.dev/documentation/components/amp-bind/) to do it. AFAIK, value binding can happen only by manual trigger (like a button click), so it won't work on initial load, unfortunately.

I've found an alternative component that seems like a more suitable choice for rendering the latest posts list - the [`amp-live-list`](https://amp.dev/documentation/components/amp-live-list/). Unfortunately, it doesn't support the `Load more` functionality, only pagination 😄 

Our use-case seems to be a mix between what the `amp-list` and the `amp-live-list` are designed for. I will continue the research tomorrow but would love to get any feedback on this. Would switching to `amp-live-list` even make sense if there's no support for the "more" button?

obenland:
If we were to make all attributes first-class args to the API, we'd be able to shed the validation/saitization callbacks. We could then just pass the json file as the arguments schema and let the API handle all validation & sanitization. It would also make the `page` argument a little less special and have it just be part of the other query args. Yay/nay?

WunderBart:
After spending some time on the documentation and trying out various configurations for the `amp-list` component, it turns out that it doesn't support dynamic resizing on page load. What that means, is that it requires a fixed height value to be provided for every layout it supports.

I'm considering it a blocker at this point as the content we want to render has variable height, and we'd always want to show it as is, without the `overflow` button to uncover the remaining part.

For more context on this issue, see [this huge thread](https://github.com/ampproject/amphtml/issues/10230) (and [this comment](https://github.com/ampproject/amphtml/issues/10230#issuecomment-376968125)).
_FYI, this [has been supported](https://github.com/ampproject/amphtml/pull/16564) at some point in the past, but it [got rid of due to some issues it caused](https://github.com/ampproject/amphtml/issues/18849)._

Even if we'd be able to use the `amp-list` component (e.g. when the `container` layout gets supported on page-load), there's another thing that would need addressing - The `list` wrapper `div` doesn't accept extra class-names and cannot be opted-out (it's being rendered even if the item's template is empty). This is how it looks: `<div role="list">{items}</div>`. This basically means that we can't use any shared/imported class-names and would have to write/copy additional CSS to keep currently applied styling.

src/blocks/homepage-articles/view.js
```diff
@@ -5,3 +5,36 @@
  */
 
 import './view.scss';
+
+const loadMoreButton = document.getElementById( 'load-more-button' );
```
getdave: Using IDs here could be problematic if we have multiple blocks on the page. I would recommend selecting by a data attribute.

src/blocks/homepage-articles/view.php
```diff
@@ -92,291 +92,124 @@ function newspack_blocks_render_block_homepage_articles( $attributes ) {
...
-				$authors = newspack_blocks_prepare_authors();
 
-				$post_counter++;
-
-				$styles = '';
-				if ( 'behind' === $attributes['mediaPosition'] && $attributes['showImage'] && has_post_thumbnail() ) {
-					$styles = 'min-height: ' . $attributes['minHeight'] . 'vh; padding-top: ' . ( $attributes['minHeight'] / 5 ) . 'vh;';
-				}
+			?>
+			<div id="posts-container" class="<?php echo esc_attr( $classes ); ?>" style="<?php echo esc_attr( $styles ); ?>">
```
getdave: Not sure we can use an `id` attr here because it should be unique in the DOM and if we have multiple Blocks on the page if won't be.

See also comment about `getElementByID()`.

src/blocks/homepage-articles/view.js
```diff
@@ -5,3 +5,36 @@
...
+const loadMoreButtonURLAttribute = 'data-load-more-url';
+const postsContainer = document.getElementById( 'posts-container');
+
+if ( loadMoreButton ) {
+  loadMoreButton.addEventListener( 'click', handleLoadMoreButtonClick );
+}
+
+function handleLoadMoreButtonClick() {
+  const loadMoreURL = this.getAttribute( loadMoreButtonURLAttribute );
+  const request = new XMLHttpRequest();
```
getdave: We should be able to use `@wordpress/api-fetch` here which wraps the `fetch` API in browsers.

src/blocks/homepage-articles/view.js
```diff
@@ -5,3 +5,36 @@
...
+
+  request.onload = function () {
+    if ( request.status === 200 ) {
+      const payload = JSON.parse( request.responseText );
+
+      renderPosts( payload.items );
+      loadMoreButton.setAttribute( loadMoreButtonURLAttribute, payload.next );
+    } else {
+      console.log( 'Something went wrong!' );
```
getdave: Next step: let's display something to the user if something went wrong. Probably by displaying a human friendly notice and allowing them to try again by re-clicking the button.

src/blocks/homepage-articles/view.js
```diff
@@ -5,3 +5,36 @@
...
+  loadMoreButton.addEventListener( 'click', handleLoadMoreButtonClick );
+}
+
+function handleLoadMoreButtonClick() {
+  const loadMoreURL = this.getAttribute( loadMoreButtonURLAttribute );
+  const request = new XMLHttpRequest();
+
+  request.onload = function () {
+    if ( request.status === 200 ) {
+      const payload = JSON.parse( request.responseText );
```
getdave: This will throw a SyntaxError exception if it's not valid JSON. We should wrap in `try/catch` and use a error handler to show an notice to the user.

src/blocks/homepage-articles/view.js
```diff
@@ -5,3 +5,36 @@
...
+      console.log( 'Something went wrong!' );
+    }
+  };
+
+  request.open( 'GET', loadMoreURL );
+  request.send();
+};
+
+function renderPosts( items ) {
+  const postsHTML = items.map( item => item.html );
```
getdave: I'm confused about this code because `postsHTML` would be an **array** of HTML strings. `insertAdjacentHTML ` expects a string of text as it's second argument.

How is this even working? 😄

src/blocks/homepage-articles/view.js
```diff
@@ -5,3 +5,36 @@
...
+    } else {
+      console.log( 'Something went wrong!' );
+    }
+  };
+
+  request.open( 'GET', loadMoreURL );
+  request.send();
+};
+
+function renderPosts( items ) {
```
getdave: Can we check that when this JS is "built" it is properly "rolled up" inside a closure or something that is protecting the scope of the vars and function declarations from bleeding into the global scope?

src/blocks/homepage-articles/view.js
```diff
@@ -5,3 +5,36 @@
...
+
+function handleLoadMoreButtonClick() {
+  const loadMoreURL = this.getAttribute( loadMoreButtonURLAttribute );
+  const request = new XMLHttpRequest();
+
+  request.onload = function () {
+    if ( request.status === 200 ) {
+      const payload = JSON.parse( request.responseText );
+
+      renderPosts( payload.items );
```
getdave: Could we

1. Show a loading indicator?
2. Disable the button from repeated clicks?
3. Only allow a single request at a time (via closure in code). This builds on 2 above.

getdave (changes requested):
Nice start to the `view.js` for the "Load More" @WunderBart 👍 

A few refinements would be great 🥇

src/blocks/homepage-articles/view.js
WunderBart: Good catch! Yeah, it stringifies it and leaves the `,` between every div. Will update 👍

src/blocks/homepage-articles/view.js
WunderBart: It is wrapped as a module, so nothing gets out :)

src/blocks/homepage-articles/view.js
WunderBart: Ad 1. Should we:
   - use some existing loading component,
   - replace button's text to `Loading...` or
   - hide button and render `Loading...` text (would address no. 2 and _almost_ 3, too)?

src/blocks/homepage-articles/view.js
WunderBart: Can we use some notice component here?

getdave: Don't think so. Just a paragraph probably. We don't have any components available like this.

src/blocks/homepage-articles/view.js
getdave: One of the later two.

src/blocks/homepage-articles/view.js
```diff
@@ -1,7 +1,54 @@
...
+ */
 import './view.scss';
+
+/**
+ * Load More Button Handling
+ */
+const loadMoreBtn = document.querySelector( '[data-load-more-btn]' );
+const loadMoreBtnURLAttr = 'data-load-more-url';
+const loadMoreLoadingText = document.querySelector( '[data-load-more-loading-text' );
+const postsContainer = document.querySelector( '[data-posts-container]');
```
getdave: These selectors should be scoped from the `loadMoreBtn`.

```
const loadMoreLoadingText = loadMoreBtn.querySelector( '[data-load-more-loading-text' );
```

If you don't do this you may select the wrong element.

src/blocks/homepage-articles/view.js
```diff
@@ -1,7 +1,54 @@
...
+
+if ( loadMoreBtn ) {
+  loadMoreBtn.addEventListener( 'click', handleLoadMoreBtnClick );
+}
+
+function handleLoadMoreBtnClick() {
+  const loadMoreURL = this.getAttribute( loadMoreBtnURLAttr );
```
getdave: We should probably avoid `this` here if possible. Try and get a ref to the target.

getdave (changes requested):
Some tweaks.

WunderBart:
At this moment, the `view.js` script supports:
  - multiple instances of the `Load more` button by switching block scope,
  - basic, but user-friendly loading and error messages,
  - hiding the `Load more` button if there are no more posts to load (`next` is _falsy_).

src/blocks/homepage-articles/view.js
```diff
@@ -1,7 +1,89 @@
...
+  const postsContainer = blockWrapper.querySelector( '[data-posts-container]' );
+
+  hideEl(loadMoreBtn);
+  hideEl(loadMoreErrorText);
+  showEl(loadMoreLoadingText);
+
+  isFetching = true;
+
+  apiFetch( { url: loadMoreURL } )
+    .then(( data ) => {
```
WunderBart: @getdave should we validate here if the `data` is a JSON?

src/blocks/homepage-articles/view.php
```diff
@@ -92,291 +92,127 @@ function newspack_blocks_render_block_homepage_articles( $attributes ) {
...
+			 * @see https://wp.me/paYJgx-jW
+			 */
+			$page = $article_query->paged ?? 1;
+			$has_more_pages = (++$page) <= $article_query->max_num_pages;
+
+			if ( ! Newspack_Blocks::is_amp() && $has_more_pages ) : ?>
+				<button type="button" data-load-more-url="<?php echo $articles_rest_url ?>"><?php _e( 'Load more articles' ); ?></button>
+				<p data-load-more-loading-text hidden><?php _e( 'Loading...' ); ?></p>
+				<p data-load-more-error-text hidden><?php _e( 'Something went wrong. Please refresh the page and/or try again.'); ?></p>
+				<p data-load-more-eod-text hidden><?php _e( 'No more posts to load!' ); ?></p>
```
obenland: Let's not show that. The lack of a load more button should be indicator enough.

getdave: Agreed 👍

src/blocks/homepage-articles/view.js
getdave: Yes we should. 

Additional: We could also check it conforms roughly to the schema we're expecting, namely objects containing `items` containing a `String`.

src/blocks/homepage-articles/view.js
```diff
@@ -1,7 +1,103 @@
...
+
+/**
+ * Load More Button Handling
+ */
+const page = document.getElementById( 'page' );
+const loadMoreBtnURLAttr = 'data-load-more-url';
+let isFetching = false;
+let isEndOfData = false;
+
+if ( page ) {
```
getdave: I like the event delegation approach here 👍 

Could/should we not simply delegate to the document as that is guaranteed to exist whereas the `id="page"` isn't?

src/blocks/homepage-articles/view.js
```diff
@@ -1,7 +1,103 @@
...
+	/**
+	 * Listen for clicks on the parent "page" element.
+	 */
+	page.addEventListener( 'click', handlePageClick );
+}
+
+function handlePageClick( event ) {
+	const isLoadMoreBtnClick = !!event.target.hasAttribute( loadMoreBtnURLAttr );
```
getdave: Do we need to check for the existence of `event.target` first?

```js
!! event.target && event.target.hasAttribute( loadMoreBtnURLAttr );
```

src/blocks/homepage-articles/view.js
```diff
@@ -1,7 +1,103 @@
...
+
+	if ( isLoadMoreBtnClick ) {
+		handleLoadMoreBtnClick( event.target );
+	}
+}
+
+function handleLoadMoreBtnClick( loadMoreBtn ) {
+	if ( !loadMoreBtn || isFetching || isEndOfData ) {
```
getdave: My guess is we should have already guarded against `null` values of `loadMoreBtn` but I guess I make sense to be doubly cautious!

getdave: Good to have `isEndOfData` check as well 👍

src/blocks/homepage-articles/view.js
```diff
@@ -1,7 +1,103 @@
...
+};
+
+function renderPosts( targetEl, items ) {
+	const postsHTML = items.map( item => item.html ).join( '' );
+
+	targetEl.insertAdjacentHTML( 'beforeend', postsHTML );
+};
+
+function hideEl( el ) {
+	console.log( el.outerHTML );
```
getdave: Let's ditch the `console` here.

src/blocks/homepage-articles/view.js
```diff
@@ -1,7 +1,103 @@
...
+	apiFetch( { url: loadMoreURL } )
+		.then( ( data ) => {
+			renderPosts( postsContainer, data.items );
+
+			if ( data.next ) {
+				loadMoreBtn.setAttribute( loadMoreBtnURLAttr, data.next );
+				showEl( loadMoreBtn );
+			} else {
+				hideEl( loadMoreBtn );
+				showEl( loadMoreEODText );
```
getdave: I think we can remove this as @obenland suggested elsewhere in his review.

src/blocks/homepage-articles/view.js
```diff
@@ -1,7 +1,103 @@
...
+	hideEl( loadMoreErrorText );
+	showEl( loadMoreLoadingText );
+
+	isFetching = true;
+
+	apiFetch( { url: loadMoreURL } )
+		.then( ( data ) => {
+			renderPosts( postsContainer, data.items );
+
+			if ( data.next ) {
```
getdave: Minor: should we _also_ check for the `.length` here? Just in case the endpoint updates and the value is an empty string?

src/blocks/homepage-articles/view.js
```diff
@@ -1,7 +1,103 @@
...
+
+	hideEl( loadMoreBtn );
+	hideEl( loadMoreErrorText );
+	showEl( loadMoreLoadingText );
+
+	isFetching = true;
+
+	apiFetch( { url: loadMoreURL } )
+		.then( ( data ) => {
+			renderPosts( postsContainer, data.items );
```
getdave: We should def validate the data is what we're expecting before we attempt to `renderPosts` here. Probably in a dedicated method.

src/blocks/homepage-articles/view.js
```diff
@@ -1,7 +1,103 @@
...
+
+				isEndOfData = true;
+			}
+
+			hideEl( loadMoreLoadingText );
+
+			isFetching = false;
+		} )
+		.catch( ( error ) => {
+			console.error( error );
```
getdave: Minor: should we only log this when in some sort of dev mode? Trying to recall if WP provides the state of `WP_DEBUG` to JS or not.

src/blocks/homepage-articles/view.js
```diff
@@ -1,7 +1,103 @@
...
+			hideEl( loadMoreLoadingText )
+			showEl( loadMoreErrorText );
+			showEl( loadMoreBtn );
+
+			isFetching = false;
+		} );
+};
+
+function renderPosts( targetEl, items ) {
+	const postsHTML = items.map( item => item.html ).join( '' );
```
getdave: Do you think its worth having some sort of guard here against `item.html` being undefined?

getdave (changes requested):
Looking much better. Nice job! A few tweaks and questions. Let's discuss.

Also we need to be testing and developing against a Post containing multiple instances of this Block.

src/blocks/homepage-articles/view.js
```diff
@@ -1,7 +1,103 @@
...
+ */
 import './view.scss';
+
+/**
+ * Load More Button Handling
+ */
+const page = document.getElementById( 'page' );
+const loadMoreBtnURLAttr = 'data-load-more-url';
+let isFetching = false;
+let isEndOfData = false;
```
getdave: **Major**: i've just realised this is a global state var. This means that if you have multiple Blocks on the page and your exhaust the posts in one Block (ie: you've loaded all the posts) then this var will Block all the other Blocks from loading more posts.

getdave (changes requested):
Found a major bug. Essentially [one Block can stop others from loading more posts](https://cld.wthms.co/woHrMq)!

src/blocks/homepage-articles/view.js
getdave: Also if one Block is fetching that probably shouldn't stop other Blocks from fetching.

src/blocks/homepage-articles/view.js
WunderBart: Good point. I assumed `page` is a convention and it will always be there. If there's any possibility that it will not, we should absolutely move to something higher, but I think `body` would be better than the whole document to attach the listener to in this case?

src/blocks/homepage-articles/view.js
getdave: Let's use closure to contain the state to a per load more button "instance" basis.

src/blocks/homepage-articles/view.js
getdave: I've always used `document` on delegation. What do others think @marekhrabe @retrofox @frontdevde @obenland?

src/blocks/homepage-articles/view.js
WunderBart: It should be fine - `""` (empty string) is treated as false, too 👍

src/blocks/homepage-articles/view.js
WunderBart: If the data object isn't valid it will be caught by the promise and the error message will be rendered. I think that would be the correct behaviour in such a situation?

src/blocks/homepage-articles/view.js
WunderBart: Great catch! Will address it in the next commit. Thanks!

src/blocks/homepage-articles/view.js
marekhrabe: I think it’s smart to use more specific wrapping element for delegation but in wp world, there is no guarantee of a theme markup so I think document might be appropriate

src/blocks/homepage-articles/view.js
```diff
@@ -16,88 +16,93 @@ import './view.scss';
...
-			showEl( loadMoreBtn );
-
-			isFetching = false;
-		} );
+	const btnURLAttr = 'data-load-more-url';
+	let isFetching = false;
+	let isEndOfData = false;
+
+	btnEl.addEventListener( 'click', function() {
+		if ( isFetching || isEndOfData ) {
```
getdave: We should be able to extract the handler `function` definition outside of the event listener. That way we don't redeclare it onclick each time.

src/blocks/homepage-articles/view.js
```diff
@@ -16,88 +16,93 @@ import './view.scss';
...
+				showEl( btnEl );
+
+				isFetching = false;
+			} );
+	} );
 };
 
 function renderPosts( targetEl, items ) {
-	const postsHTML = items.map( item => item.html ).join( '' );
+	if ( !targetEl || !items || items.length === 0 ) {
```
getdave: Nit: `!items.length` will have the same effect no?

src/blocks/homepage-articles/view.js
```diff
@@ -16,88 +16,93 @@ import './view.scss';
...
-function handlePageClick( event ) {
-	const isLoadMoreBtnClick = !!event.target.hasAttribute( loadMoreBtnURLAttr );
-
-	if ( isLoadMoreBtnClick ) {
-		handleLoadMoreBtnClick( event.target );
-	}
-}
+document
+	.querySelectorAll( '[data-load-more-btn]' )
+	.forEach( makeLoadMoreHandler )
```
getdave: Missing `;`. Not sure how this getting past the linter...

src/blocks/homepage-articles/view.php
```diff
@@ -136,10 +136,9 @@ function newspack_blocks_render_block_homepage_articles( $attributes ) {
 			$has_more_pages = (++$page) <= $article_query->max_num_pages;
 
 			if ( ! Newspack_Blocks::is_amp() && $has_more_pages ) : ?>
-				<button type="button" data-load-more-url="<?php echo $articles_rest_url ?>"><?php _e( 'Load more articles' ); ?></button>
+				<button type="button" data-load-more-btn data-load-more-url="<?php echo $articles_rest_url ?>"><?php _e( 'Load more articles' ); ?></button>
```
getdave: `data-load-more-url="<?php echo $articles_rest_url ?>` needs escaping using the appropriate `esc_*` function.

src/blocks/homepage-articles/view.js
```diff
@@ -16,88 +16,93 @@ import './view.scss';
...
+
+					isEndOfData = true;
+				}
+
+				hideEl( loadingEl );
+
+				isFetching = false;
+			} )
+			.catch( ( error ) => {
+				// console.error( error );
```
getdave: Can we remove this comment?

getdave (changes requested):
I've tested this and it now works functionality wise. Great 👍 

Some code points to tidy up.

src/blocks/homepage-articles/view.js
getdave: It will catch invalid JSON, but it won't necessarily match the schema (shape) of the data we're expecting.

I was just thinking we should check for a collection of `items` with an `html` prop which is a `String`.

getdave:
Just noticed a bug with https://github.com/Automattic/newspack-blocks/blob/a1a11c41ba2018ac81860192c88975390bc571f6/src/blocks/homepage-articles/edit.js#L146.

`post.title.rendered` may or may not exist. We need to test.

getdave:
## Status of this PR

* May require splitting into two PRs: one for refactor and one for the REST API work. Possibly overkill? 
* Fix some minor code hygiene points.

src/blocks/homepage-articles/edit.js
```diff
@@ -267,6 +269,14 @@ class Edit extends Component {
 							required
 						/>
 					) }
+					{ ! singleMode && (
+						<ToggleControl
+							label={ __( 'Show "More" Button' ) }
+							checked={ moreButton }
+							onChange={ () => setAttributes( { moreButton: ! moreButton } ) }
+							help={ __('Only available for non-AMP requests.', 'newspack-blocks') }
```
obenland: I worry this help text would be more confusing than helpful to the average user

WunderBart: I agree. I guess making the text less confusing is one option, but wouldn't it be even better if we just hide the whole toggle for non-AMP requests?

jeffersonrabb: This text appears in the editor and the concept of being an AMP request only really applies to the view. It's conceivable the text could be shown only if the AMP plugin is activated but that seems complex and error-prone.

WunderBart: Thanks, @jeffersonrabb - I've totally missed that fact! Keeping the text to make editors aware makes total sense.

jeffersonrabb:
Closing, as this work will continue in https://github.com/Automattic/newspack-blocks/pull/259.

PR #232, opened 2019-11-27
Update Calypso Build to 5.1.0

jeffersonrabb:
### All Submissions:

This PR updates `calypso-build` to the latest version, 5.1.0. 

The update removes a few Sass Muriel color variables that the Donate block uses. I didn't want to introduce any style changes in this PR so I created a temporary Sass file with these variables (https://github.com/Automattic/newspack-blocks/compare/update/calypso-build-5.1.0?expand=1#diff-d7cc79577ac5a9dd72abcdbc89faa614). In a future PR we should replace the Muriel color values and remove this file, cc: @thomasguillot. 

The approach to dependency extraction changes in this update to use https://developer.wordpress.org/block-editor/packages/packages-dependency-extraction-webpack-plugin/. 

1. `npm ci`. Verify `calypso-build` updates to version 5.1.0.
2. `npm run start`. Verify all blocks build with no errors.
3. Test normal operation of all three blocks in editor and view. Should be identical to `master`.
4. Bonus ⭐️: None of the blocks have view-side scripts. Add some Javascript to one  of  the `view.js` files and verify that it executes on the  front end. To verify the dependency enqueuing import a library. E.G.:

```
import classNames from 'classnames';
import { Component } from '@wordpress/element';
console.log( classNames( 'hello', 'classnames' ) );
```

### Other information:

### DIFF
--- a/class-newspack-blocks.php
+++ b/class-newspack-blocks.php
@@ -8,14 +8,22 @@ class Newspack_Blocks {
 	 * Enqueue block scripts and styles for editor.
 	 */
 	public static function enqueue_block_editor_assets() {
+		$asset_path = NEWSPACK_BLOCKS__PLUGIN_DIR . NEWSPACK_BLOCKS__BLOCKS_DIRECTORY . 'editor.asset.php';
+		$asset      = file_exists( $asset_path )
+			? require $asset_path
+			: array(
+				'dependencies' => array(),
+				'version'      => '1.0',
+			);
+
 		$editor_script = plugins_url( NEWSPACK_BLOCKS__BLOCKS_DIRECTORY . 'editor.js', __FILE__ );
 		$editor_style  = plugins_url( NEWSPACK_BLOCKS__BLOCKS_DIRECTORY . 'editor.css', __FILE__ );
-		$dependencies  = self::dependencies_from_path( NEWSPACK_BLOCKS__PLUGIN_DIR . 'dist/editor.deps.json' );
+
 		wp_enqueue_script(
 			'newspack-blocks-editor',
 			$editor_script,
-			$dependencies,
-			NEWSPACK_BLOCKS__VERSION,
+			$asset['dependencies'],
+			$asset['version'],
 			true
 		);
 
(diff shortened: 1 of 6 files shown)


class-newspack-blocks.php
```diff
@@ -1,21 +1,34 @@
...
 	/**
 	 * Enqueue block scripts and styles for editor.
 	 */
 	public static function enqueue_block_editor_assets() {
+		$asset_path = NEWSPACK_BLOCKS__PLUGIN_DIR . NEWSPACK_BLOCKS__BLOCKS_DIRECTORY . 'editor.asset.php';
+		$asset      = file_exists( $asset_path )
+			? require $asset_path
+			: array(
+				'dependencies' => array(),
+				'version'      => '1.0',
```
retrofox: sorry for this question, but does it mean this version?

class-newspack-blocks.php
```diff
@@ -108,38 +121,29 @@ public static function enqueue_view_assets( $type ) {
...
 		}
 		if ( file_exists( NEWSPACK_BLOCKS__PLUGIN_DIR . $script_path ) ) {
-			$dependencies = self::dependencies_from_path( NEWSPACK_BLOCKS__PLUGIN_DIR . "dist/{$type}/view.deps.json" );
+			$asset_path = NEWSPACK_BLOCKS__PLUGIN_DIR . "dist/{$type}/view.asset.php";
+			$asset      = file_exists( $asset_path )
+				? require $asset_path
+				: array(
+					'dependencies' => array(),
+					'version'      => '1.0',
+				);
```
retrofox: I feel that the removed `dependencies_from_path` method in some way was avoiding duplicated this code. Do you think to keep a method to unify it is a bad idea?

class-newspack-blocks.php
jeffersonrabb: This is a fallback that theoretically should never be called. In the [documentation example](https://github.com/Automattic/newspack-blocks/pull/232/commits/11c6df0bcac6e6335a7a956e0ed9cd3e4f94ab59) the fallback version uses the script's `filemtime` but  I went for 1.0 just to simplify and make it clear this was a fallback. I could easily be talked out of this 😄

retrofox: oh right. I think the `filemtime` was to force to the browser to load the last version of the script

retrofox: Oh, the version is automatically defined bu the compiled file.

class-newspack-blocks.php
jeffersonrabb: This is done in https://github.com/Automattic/newspack-blocks/pull/232/commits/157e179dca52d262d52f20450ec9ad5a9e781224

class-newspack-blocks.php
jeffersonrabb: Matched the approach shown in the example in https://github.com/Automattic/newspack-blocks/pull/232/commits/157e179dca52d262d52f20450ec9ad5a9e781224

jeffersonrabb:
Thanks for the reviews!

PR #234, opened 2019-11-27
Use `static` instead of `self` in order to make the method extensible

retrofox:
### All Submissions:

It replaces the how the static methods are internally called with `self::` by `static::`. It allows re-write them from the inherited class if it's needed.

```php
class MyCustomBlocks extends Newspack_Blocks {
    public static script_enqueue_helper() {
        // it will work because parent method is called with `static::` instead of `self::`
    }
}
```

Apply these changes. Everything should work as expected.

### DIFF
--- a/class-newspack-blocks.php
+++ b/class-newspack-blocks.php
@@ -40,7 +40,7 @@ class Newspack_Blocks {
 	 * Enqueue block scripts and styles for editor.
 	 */
 	public static function enqueue_block_editor_assets() {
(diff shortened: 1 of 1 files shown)


claudiulodro:
Is this PR still relevant after the filter was added?

retrofox:
> Is this PR still relevant after the filter was added?

I think it is. It allows overwriting the method, so guessing an improvement. Let me update the doc.

jeffersonrabb (approved):
LGTM

PR #242, opened 2019-12-02
Add REST endpoint from #226

obenland:

Adds API endpoint from #226 (incl. `block.json`).

Navigate to `/wp-json/newspack-blocks/v1` on your test site and make sure you get a valid response.

### DIFF
--- a/src/blocks/homepage-articles/class-wp-rest-newspack-articles-controller.php
+++ b/src/blocks/homepage-articles/class-wp-rest-newspack-articles-controller.php
@@ -1,0 +1,116 @@
+<?php
+/**
+ * WP_REST_Newspack_Articles_Controller file.
+ */
+
+/**
+ * Class WP_REST_Newspack_Articles_Controller.
+ */
+class WP_REST_Newspack_Articles_Controller extends WP_REST_Controller {
+
+	/**
+	 * Attribute schema.
+	 *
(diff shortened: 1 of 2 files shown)


src/blocks/homepage-articles/class-wp-rest-newspack-articles-controller.php
```diff
@@ -0,0 +1,116 @@
...
+
+	/**
+	 * Sets up and returns attribute schema.
+	 *
+	 * @return array
+	 */
+	public function get_attribute_schema() {
+		if ( empty( $this->attribute_schema ) ) {
+			$block_json = json_decode(
+				file_get_contents( __DIR__ . '/block.json' ),
```
retrofox: `block.json` seems to be duplicated in the `view.php` file. Could we use the same technique there as we do here, in order to avoid duplication code?

retrofox: oh, it's already handled [here](https://github.com/Automattic/newspack-blocks/pull/248). Props to @jeffersonrabb!

obenland:
@retrofox I rebased, this is ready for another round of reviews.

jeffersonrabb:
Rebased, pushed a small commit to fix up function names and template paths.

PR #255, opened 2019-12-03
Extract To Templates

jeffersonrabb:
### All Submissions:

This PR extracts the templates from https://github.com/Automattic/newspack-blocks/pull/226, and copies over a the template include function (`newspack_blocks_template_inc` in 226, `Newspack_Blocks::template_inc` here). On a great tip from @claudiulodro the templates are wrapped in  `call_user_func(function( $data ) { template_contents }, $data)` so variables used within are scoped and don't need to be prefixed. I did not carry over the `READ MORE` button markup from 226, but I did leave the wrapper `<div>` and new placement of the heading to help us see if either of these changes will cause visual regressions.

This is the last of three PRs that deal with significant code refactoring related to the Read More button.

Output should be visually and functionally identical to `master` branch. 

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/templates/article.php
+++ b/src/blocks/homepage-articles/templates/article.php
@@ -99,7 +99,7 @@ call_user_func(
 							$time_string = '<time class="entry-date published" datetime="%1$s">%2$s</time><time class="updated" datetime="%3$s">%4$s</time>';
 						endif;
 						printf(
-							esc_attr( $time_string ),
+							wp_kses_post( $time_string ),
 							esc_attr( get_the_date( DATE_W3C ) ),
 							esc_html( get_the_date() ),
 							esc_attr( get_the_modified_date( DATE_W3C ) ),
--- a/src/blocks/homepage-articles/templates/article.php
+++ b/src/blocks/homepage-articles/templates/article.php
@@ -99,7 +99,15 @@ call_user_func(
 							$time_string = '<time class="entry-date published" datetime="%1$s">%2$s</time><time class="updated" datetime="%3$s">%4$s</time>';
 						endif;
 						printf(
-							wp_kses_post( $time_string ),
+							wp_kses(
+								$time_string,
+								array(
+									'time' => array(
+										'class'    => true,
+										'datetime' => true,
+									),
+								)
+							),
 							esc_attr( get_the_date( DATE_W3C ) ),
 							esc_html( get_the_date() ),
 							esc_attr( get_the_modified_date( DATE_W3C ) ),
--- a/src/blocks/homepage-articles/templates/article.php
+++ b/src/blocks/homepage-articles/templates/article.php
@@ -3,97 +3,116 @@
  * Article template.
  *
  * @global array $attributes Block attributes.
+ * @package WordPress
  */
 
-$newspack_blocks_authors = Newspack_Blocks::prepare_authors();
-$styles = '';
-if ( 'behind' === $attributes['mediaPosition'] && $attributes['showImage'] && has_post_thumbnail() ) {
-	$styles = 'min-height: ' . $attributes['minHeight'] . 'vh; padding-top: ' . ( $attributes['minHeight'] / 5 ) . 'vh;';
-}
-$image_size = 'newspack-article-block-uncropped';
-if ( 'uncropped' !== $attributes['imageShape'] ) {
-	$image_size = Newspack_Blocks::image_size_for_orientation( $attributes['imageShape'] );
-}
-$thumbnail_args = '';
-// If the image position is behind, pass the object-fit setting to maintain styles with AMP.
-if ( 'behind' === $attributes['mediaPosition'] ) {
-	$thumbnail_args = array( 'object-fit' => 'cover' );
-}
-$category = false;
-// Use Yoast primary category if set.
-if ( class_exists( 'WPSEO_Primary_Term' ) ) {
-	$primary_term = new WPSEO_Primary_Term( 'category', get_the_ID() );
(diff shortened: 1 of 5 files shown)


src/blocks/homepage-articles/templates/articles-list.php
```diff
@@ -0,0 +1,23 @@
...
+ * Article list template.
+ *
+ * @global WP_Query $article_query Article query.
+ * @global array    $attributes
+ * @package WordPress
+ */
+
+call_user_func(
+	function( $data ) {
+		echo wp_kses_post(
```
obenland: This feels unnecessary, given that we have full control over the output of `article.php` and have all user data properly escaped within. Additionally we're running things through `wp_kses_post()` that has already gone through it in `articles-loop.php`?

jeffersonrabb: That's true. I often over-escape for two reasons—to avoid aPHPCS warnings for unescaped output (`WordPress.Security.EscapeOutput.OutputNotEscaped`) to make the code a tiny bit easier to review.

obenland: `wp_kses_post()` is not free, especially once the posts get long and the user choses to display 10+ of them at a time. Is there a reason we need to capture the template output at all and can't just have `article.php` display it?

jeffersonrabb:
Just set this back to Needs Review so  we don't accidentally merge before @laurelfulford has verified.

src/blocks/homepage-articles/templates/article.php
```diff
@@ -0,0 +1,118 @@
...
+		class="post-has-image"
+		<?php endif; ?>
+		<?php if ( $styles ) : ?>
+		style="<?php echo esc_attr( $styles ); ?>"
+		<?php endif; ?>
+		>
+		<?php if ( has_post_thumbnail() && $attributes['showImage'] && $attributes['imageShape'] ) : ?>
+			<figure class="post-thumbnail">
+				<a href="<?php the_permalink(); ?>" rel="bookmark">
+					<?php the_post_thumbnail( $image_size, $thumbnail_args ); ?>
```
laurelfulford: For some reason, the `$thumbnail_args` don't seem to be applied -- when I test the 'behind' image position with AMP enabled, the images don't fill the available space:

image

src/blocks/homepage-articles/view.php
```diff
@@ -69,155 +69,32 @@ function newspack_blocks_render_block_homepage_articles( $attributes ) {
 
 	if ( $article_query->have_posts() ) :
 		?>
-		<div class="<?php echo esc_attr( $classes ); ?>" style="<?php echo esc_attr( $styles ); ?>">
+		<div>
 
 			<?php if ( '' !== $attributes['sectionHeader'] ) : ?>
 				<h2 class="article-section-title">
```
laurelfulford: The `.article-section-title` now lives "outside" of the block's containing class, so the styles aren't being applied correctly. My test site is running Style 3 with the 'all caps' option on for the custom typography -- I get:

image

When I should get:

image

I think if we want to move it outside of the element with the block-specific class, we might want to give it a more unique classname to prevent clashes -- something like `.wpnbha-header`? We'll also need to make a separate issue to make this update in the theme. 

Edited to add: Actually, the header now also doesn't pick up the block's custom colours:

image

This is going to be an issue when a site has, say, nested this block in a Group block with a dark background and needs to change the text colour (something happening on at least one live site now).

src/blocks/homepage-articles/templates/article.php
```diff
@@ -0,0 +1,118 @@
...
+						</span><!-- .author-name -->
+						<?php
+					endif;
+					if ( $attributes['showDate'] ) :
+						$time_string = '<time class="entry-date published updated" datetime="%1$s">%2$s</time>';
+						if ( get_the_time( 'U' ) !== get_the_modified_time( 'U' ) ) :
+							$time_string = '<time class="entry-date published" datetime="%1$s">%2$s</time><time class="updated" datetime="%3$s">%4$s</time>';
+						endif;
+						printf(
+							wp_kses_post( $time_string ),
```
laurelfulford: The `wp_kses_post()` here is stripping out the `<time>` tags, so we're losing some styling -- when there's just a published date, it's really close to the author:

image

... and if there's an 'updated' date, it's showing both:

image

Since we know what markup should be there, what about a `wp_kses()` function passing the `time` tag and the attributes we have above (`class` and `datetime`)?

jeffersonrabb: Resolved in https://github.com/Automattic/newspack-blocks/pull/255/commits/b2588ebfc7c73b84a9cbc5c05483ab0dec91ffcc

src/blocks/homepage-articles/templates/article.php
jeffersonrabb: I think that the eased escaping in https://github.com/Automattic/newspack-blocks/pull/255/commits/b2588ebfc7c73b84a9cbc5c05483ab0dec91ffcc fixes this one too, unintentionally. Could you check it again?

src/blocks/homepage-articles/view.php
jeffersonrabb: Should be resolved in https://github.com/Automattic/newspack-blocks/pull/255/commits/fd812316b70ebe6c80e11343ec2f48e30bd955f6. This change _might_ cause trouble for the read more button work, but we can address that as it comes up.

jeffersonrabb:
Thanks for looking, @laurelfulford! I believe I got them all.

jeffersonrabb:
Thank you all for taking a look!

PR #263, opened 2019-12-06
Add rendered posts' IDs as "Load More" query param

WunderBart:
### All Submissions:

Add `exclude_ids` query param with currently rendered posts IDs to the "More" request URL. This should be used to address the post duplication issue (paYJgx-jW-p2#comment-418, paYJgx-jW-p2#comment-433).

1. Add `Homepage Posts` block & make sure there are any posts to render
2. Go to preview and open network tab in browser's devtools (filter by XHR)
3. Click the "Load more articles" button and check if `exclude_ids` param has been set to the request URL with correct posts' IDs.

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/class-wp-rest-newspack-articles-controller.php
+++ b/src/blocks/homepage-articles/class-wp-rest-newspack-articles-controller.php
@@ -64,13 +64,10 @@ class WP_REST_Newspack_Articles_Controller extends WP_REST_Controller {
 
 		$article_query_args = Newspack_Blocks::build_articles_query( $attributes );
 
-		// Append custom pagination arg for REST API endpoint.
-		$article_query_args['paged'] = $page;
-
 		$query = array_merge(
 			$article_query_args,
 			[
-				'posts_per_page' => $article_query_args['posts_per_page'] + count( $exclude_ids ),
+				'post__not_in' => $exclude_ids,
 			]
 		);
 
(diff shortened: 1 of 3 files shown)


src/blocks/homepage-articles/view.js
```diff
@@ -94,13 +94,28 @@ function buildLoadMoreHandler( btnEl ) {
...
+
+		fetchWithRetry( { url: requestURL.toString(), onSuccess, onError }, fetchRetryCount );
 	};
 }
 
+/**
+ * Returns unique IDs for posts that are currently in the DOM.
+ */
+function getRenderedPostsIds() {
+	const postEls = document.querySelectorAll( 'article[data-post-id]' );
```
getdave: Is this `data-` attr unique enough to allow for a document wide query? I would be tempted to namespace it to this block (eg: `data-a8c-block-posts-post-id`).

Also do you _need_ the `article` qualifier?

WunderBart: Good call! I don't think we need `article` there. Do we have a pattern for naming the `data-*` selectors in such cases? If not, to make it distinct from others maybe sth like `a8c-<block-name>-<element-name>` would be better? E.g. `data-a8c-homepage-articles-post-id`. The downside of that is that it would have to be updated when the block name changes (paYJgx-li-p2). Naming is hard.

src/blocks/homepage-articles/class-wp-rest-newspack-articles-controller.php
```diff
@@ -85,12 +82,6 @@ public function get_items( $request ) {
 		$post_counter = 0;
```
claudiulodro: `$post_counter` is unused now.

jeffersonrabb: Good catch! https://github.com/Automattic/newspack-blocks/pull/263/commits/4abaa7b62d9d642b22fe18729a2d56a0a789d9a9

PR #264, opened 2019-12-06
Replace apiFetch

obenland:

Replaces apiFetch with XMLHttpRequest. (I wasn't sure how serious we take IE11 support).
This fixes a bug with apiFetch on wp.com, where it always sends an `X-wp-nonce` header that wp.com rejects. See https://github.com/Automattic/jetpack/pull/12823/commits/c8029269a4ee6bc126fd072d1e00043f7871d6cd

I open to other fixes. Jetpack seems to have just gone with `fetch`, fine with that, too. I'd just like it to work on wp.com :) 

1. Set up a block with more button and make sure it still loads new posts.

### DIFF
--- a/src/blocks/homepage-articles/view.js
+++ b/src/blocks/homepage-articles/view.js
@@ -112,7 +112,7 @@ function buildLoadMoreHandler( btnEl ) {
 			showEl( btnEl );
 		};
 
-		apiFetchWithRetry(
+		fetchWithRetry(
 			{ url: btnEl.getAttribute( btnURLAttr ), onSuccess, onError },
 			fetchRetryCount
 		);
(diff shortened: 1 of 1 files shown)


src/blocks/homepage-articles/view.js
```diff
@@ -107,35 +98,56 @@ function buildLoadMoreHandler( btnEl ) {
...
+ * @param {Object} options XMLHttpRequest options
  * @param {Number} n retry count before throwing
  */
 function apiFetchWithRetry( options, n ) {
-	return apiFetch( options ).catch( error => {
-		if ( n === 1 ) {
-			throw error;
+	const xhr = new XMLHttpRequest();
+	xhr.onreadystatechange = () => {
+		if ( xhr.readyState !== 4 || n === 0 ) {
```
WunderBart: `n === 0` here will fire the request one time too many. We need to do `n === 1`.

src/blocks/homepage-articles/view.js
```diff
@@ -107,35 +98,56 @@ function buildLoadMoreHandler( btnEl ) {
...
 /**
- * Wrapper for api-fetch that performs given number of retries when error
+ * Wrapper for XMLHttpRequest that performs given number of retries when error
  * occurs.
  *
- * @param {Object} options api-fetch options
+ * @param {Object} options XMLHttpRequest options
  * @param {Number} n retry count before throwing
  */
 function apiFetchWithRetry( options, n ) {
```
WunderBart: I think we should rename it to just `fetchWithRetry` as it's no longer related to the `apiFetch`.

src/blocks/homepage-articles/view.js
WunderBart: Would be good to have a comment about what's happening here.

WunderBart: Ah, my bad - this is the **retry** count, not the call count. All good 👍

src/blocks/homepage-articles/view.js
```diff
@@ -126,26 +126,39 @@ function buildLoadMoreHandler( btnEl ) {
...
  */
-function apiFetchWithRetry( options, n ) {
+function fetchWithRetry( options, n ) {
 	const xhr = new XMLHttpRequest();
+
 	xhr.onreadystatechange = () => {
-		if ( xhr.readyState !== 4 || n === 0 ) {
+		/**
+		 * Return if the request is completed.
+		 */
```
obenland: @WunderBart Would you mind updating the comment markup to use `// …` for single line comments?
See https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/javascript/#single-line-comments

WunderBart: Thanks for the link! I was wondering which type I should use... Fixing it now!

PR #265, opened 2019-12-08
Load More Button: AMP Compatibility

jeffersonrabb:
### All Submissions:

Adds AMP compatibility for the Homepage Posts bock Load More button. 

1. Begin by verifying the block behaves exactly as it does on `master` in non-AMP requests.
2. Install AMP plugin, use Transitional mode.
3. View page with Homepage Posts block on it, switch to AMP request, verify identical behavior.
4. Test in pages with multiple Homepage Posts blocks. There should be no duplicate posts displayed after clicking Load More buttons.
5. Test in a variety of browsers.

### Other information:

### DIFF
--- a/amp/homepage-articles/view.js
+++ b/amp/homepage-articles/view.js
@@ -1,0 +1,217 @@
+const btnURLAttr = 'data-load-more-url';
+const fetchRetryCount = 3;
+
+/**
+ * Load More Button Handling
+ */
+
+attachLoadMoreHandler( document.querySelector( '[data-load-more-btn]' ) );
+
+/**
+ * Attaches an event handler to the Load more button.
+ * @param {DOMElement} btnEl the button that was clicked
+ */
+function attachLoadMoreHandler( btnEl ) {
+	if ( ! btnEl ) {
+		return null;
+	}
+
+	const handler = buildLoadMoreHandler( btnEl );
+
+	btnEl.addEventListener( 'click', handler );
+}
+
+/**
+ * Builds a function to handle clicks on the load more button.
+ * Creates internal state via closure to ensure all state is
+ * isolated to a single Block + button instance.
+ *
+ * @param {DOMElement} btnEl the button that was clicked
+ */
+function buildLoadMoreHandler( btnEl ) {
+	/**
+	 * Set elements from scope determined by the clicked "Load more" button.
+	 */
+	const blockWrapperEl = btnEl.parentNode; // scope root element
+	const postsContainerEl = blockWrapperEl.querySelector( '[data-posts-container]' );
+	const loadingEl = blockWrapperEl.querySelector( '[data-load-more-loading-text]' );
+	const errorEl = blockWrapperEl.querySelector( '[data-load-more-error-text]' );
+
+	/**
+	 * Set initial state flags.
+	 */
+	let isFetching = false;
+	let isEndOfData = false;
+
+	return () => {
+		/**
+		 * Early return if still fetching or no more posts to render.
+		 */
+		if ( isFetching || isEndOfData ) {
+			return false;
+		}
+
+		isFetching = true;
+
+		/**
+		 * Set elements visibility for fetching state.
+		 */
+		hideEl( btnEl );
+		hideEl( errorEl );
+		showEl( loadingEl );
+
+		const onSuccess = data => {
+			console.log( data );
+			/**
+			 * Validate received data.
+			 */
+			if ( isPostsDataValid( data ) ) {
+				/**
+				 * Render posts' HTML from string.
+				 */
+				const postsHTML = data.items.map( item => item.html ).join( '' );
+				const ampLayout = document.createElement( 'amp-layout' );
+				ampLayout.innerHTML = postsHTML;
+				blockWrapperEl.insertBefore( ampLayout, btnEl );
+
+				if ( data.next ) {
+					/**
+					 * Save next URL as button's attribute.
+					 */
+					btnEl.setAttribute( btnURLAttr, data.next );
+
+					/**
+					 * Unhide button since there are more posts available.
+					 */
+					showEl( btnEl );
+				} else {
+					isEndOfData = true;
+				}
+
+				isFetching = false;
+
+				hideEl( loadingEl );
+			}
+		};
+
+		const onError = () => {
+			isFetching = false;
+
+			/**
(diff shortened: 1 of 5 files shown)


src/blocks/homepage-articles/view.php
```diff
@@ -69,14 +69,20 @@ function( $attribute ) {
...
+			]
 		),
 		rest_url( '/newspack-blocks/v1/articles' )
 	);
 
 	ob_start();
 
 	if ( $article_query->have_posts() ) : ?>
+		<?php if ( Newspack_Blocks::is_amp() ) : ?>
+			<amp-script layout="container" src="https://newspack1.ngrok.io/wp-content/plugins/newspack-blocks/amp/homepage-articles/view.js">
```
westonruter: Note: The `src` here will need to be on the same origin as the site. If not, then you'll need to add add a `meta[name=amp-script-src]` tag to the `head`. See https://amp.dev/documentation/components/amp-script/#security-features

The AMP plugin includes a `amp_generate_script_hash()` function for doing this, which can also be used if you want to inline the script rather than reference an external file.

amp/homepage-articles/view.js
```diff
@@ -0,0 +1,217 @@
...
+
+/**
+ * Unhides given DOM element.
+ *
+ * @param {DOMElement} el
+ */
+function showEl( el ) {
+	el.style.display = '';
+	el.removeAttribute( 'hidden' );
+}
```
westonruter: As opposed to setting the `style`, why not just toggle the `hidden` property? I don't think you actually even need `hideEl()` and `showEl()`. You can just do:

```js
btnEl.hidden = true;
```

and

```js
loadingEl.hidden = false;
```

This will avoid issues with inline styles.

All browsers, including IE11, natively support `hidden` in their user agent styles: https://caniuse.com/#feat=hidden

And `amp-script` supports toggling `hidden`. Ref: https://github.com/ampproject/amphtml/issues/23156

westonruter:
> * [ ]  Missing markup (inline styles?) for posts loaded using the endpoint. Note font sizing.

Is this referring to style rules? Are there missing style rules due to tree shaking? Note an easy way to prevent tree-shaking style rules that target elements that aren't in the initial DOM rendered by PHP is to explicitly mention `amp-script` in the selector. Anything below the `amp-script` will be ignored. For example:

```css
.sidebar amp-script .post-12345 { color: red; } 
```

src/blocks/homepage-articles/class-wp-rest-newspack-articles-controller.php
```diff
@@ -114,4 +129,28 @@ public function get_attribute_schema() {
...
+			$dom,
+			amp_get_content_sanitizers(),
+			[
+				'use_document_element' => false,
+			]
+		);
+		foreach ( iterator_to_array( $dom->getElementsByTagName( 'noscript' ) ) as $noscript ) {
+			$noscript->parentNode->removeChild( $noscript ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.NotSnakeCaseMemberVar
+		}
+		$markup = AMP_DOM_Utils::get_content_from_dom( $dom );
```
westonruter: A workaround to persist `style` attributes to avoid them being extracted into style rules and converted into CSS classes would be to do something like this:

```php
$xpath = new DOMXPath( $dom );

foreach ( iterator_to_array( $xpath->query( '//@style' ) ) as $style_attr ) {
	if ( false !== strpos( $style_attr->nodeValue, '!important' ) ) {
		continue;
	}
	$element = $style_attr->parentNode;
	$element->removeAttributeNode( $style_attr );
	$element->setAttribute( 'data-temp-style', $style_attr->nodeValue );
}

AMP_Content_Sanitizer::sanitize_document( /* ... */ );

foreach ( iterator_to_array( $xpath->query( '//@data-temp-style' ) ) as $style_attr ) {
	$element = $style_attr->parentNode;
	$element->removeAttributeNode( $style_attr );
	$element->setAttribute( 'style', $style_attr->nodeValue );
}
```

AMP allows inline `style` attributes as long as they do not contain `!important` properties. 

Nevertheless, the AMP plugin's style sanitizer doesn't pass them through specifically in order to ensure `!important` properties can be reliably be converted into high-specificity style rules and preserve the cascade. For more info on that, see https://github.com/ampproject/amp-wp/issues/1313

src/blocks/homepage-articles/class-wp-rest-newspack-articles-controller.php
```diff
@@ -114,4 +129,28 @@ public function get_attribute_schema() {
...
+		$markup = preg_replace( '/<!--(.|\s)*?-->/', '', $markup );
```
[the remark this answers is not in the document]
jeffersonrabb: Nice: https://github.com/Automattic/newspack-blocks/pull/265/commits/808941591de0cb285f42c69d6841c55436ec056f. Seems like everything we collaborate on these days involves XPath.

src/blocks/homepage-articles/view.php
jeffersonrabb: This was an oversight—the script will always be on the same origin: https://github.com/Automattic/newspack-blocks/pull/265/commits/5910493dd4b28f03c6abe26b237b4a48882e6b33

amp/homepage-articles/view.js
jeffersonrabb: I'm trying to mirror the [non-AMP script](https://github.com/Automattic/newspack-blocks/blob/master/src/blocks/homepage-articles/view.js) as closely as possible in the hope that eventually we may be able to use the same script in both environments. The comment may be valid, but for the purposes of this PR my goal is to alter the script as little as possible.

src/blocks/homepage-articles/class-wp-rest-newspack-articles-controller.php
jeffersonrabb: False alarm—the styling issue was due to a hierarchy error: https://github.com/Automattic/newspack-blocks/pull/265/commits/e64302bdaecf5ab7673fb2e75169da8473dfa000. This may come in handy elsewhere, but doesn't seem to be necessary now.

jeffersonrabb:
The feature doesn't work in IE11 (neither does the non-AMP implementation) but this will be filed as an issue and addressed in a future PR.

PR #270, opened 2019-12-09
Exclude  IDs In Load More Posts Endpoint

jeffersonrabb:
### All Submissions:

Adds support  for an  `exclude_ids` parameter. As with the normal rendering logic, this avoids using `post__not_in` logic and instead queries for more posts than were requested and sorts out which to return in PHP logic. This is related to  https://github.com/Automattic/newspack-blocks/pull/263, which tacks on an array of all visible posts so they will be excluded.

This one is a little tough to test without https://github.com/Automattic/newspack-blocks/pull/263. Here's one approach:

1. Branch switch to `update/load-more-with-post-ids`.
2. Add several `Homepage Posts`blocks, and switch on `Show "More" Button`.
3. Publish and view the post.
4. With web inspector open to Network->XHR tab (Chrome), click the `Load More Posts` button.
5. You should see requests that look something like this: `https://newspack1.ngrok.io/wp-json/newspack-blocks/v1/articles?showExcerpt=0&postsToShow=2&typeScale=1&sectionHeader=Latest%20News&textColor=secondary-variation&className=&showDate=1&showImage=1&showCaption=0&imageShape=landscape&minHeight=0&moreButton=0&showAuthor=1&showAvatar=1&showCategory=0&postLayout=list&columns=3&mediaPosition=top&imageScale=3&mobileStack=0&specificMode=0&customTextColor=&singleMode=0&page=2&exclude_ids=5466%2C5464%2C5462%2C5460&_locale=user`. Double-click to open the endpoint in a new window.
6. Branch switch to `add/load-more-endpoint-exclude-ids` and reload the request. 
7. The results should now exclude any posts that are visible on  the page.

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/class-wp-rest-newspack-articles-controller.php
+++ b/src/blocks/homepage-articles/class-wp-rest-newspack-articles-controller.php
@@ -1,6 +1,8 @@
 <?php
 /**
  * WP_REST_Newspack_Articles_Controller file.
+ *
+ * @package WordPress
  */
 
 /**
@@ -101,17 +103,24 @@ class WP_REST_Newspack_Articles_Controller extends WP_REST_Controller {
 		if ( $next_page <= $article_query->max_num_pages ) {
 			$next_url = add_query_arg(
 				array_merge(
-					array_map( function( $attribute ) { return $attribute === false ? '0' : $attribute; }, $attributes ),
+					array_map(
+						function( $attribute ) {
+							return false === $attribute ? '0' : $attribute;
+						},
+						$attributes
+					),
 					[ 'page' => $next_page ] // phpcs:ignore PHPCompatibility.Syntax.NewShortArray.Found
 				),
 				rest_url( '/newspack-blocks/v1/articles' )
 			);
 		}
 
-		return rest_ensure_response( [
-			'items' => $items,
-			'next'  => $next_url,
-		] );
+		return rest_ensure_response(
+			[
+				'items' => $items,
+				'next'  => $next_url,
+			]
+		);
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


claudiulodro:
> Block A:
> should display 1 post at a time (set by Number of items option)
> can display posts with IDs [ 1, 4, 5 ]
>
> Block B:
> should display 1 post at a time
> can display posts with IDs [ 3, 4 ]

Are we sure this is a behavior that makes sense to readers? I don't think many readers (or potentially site owners) are going to understand that posts were previously loaded by a "More" button in a different section they were browsing, so now they won't show up in the new section they're browsing. 

IMHO the page-wide exclusion-synchronizing seems confusing to me as a reader. The clearest behavior would be to load `[4, 5]` in block A and also `[4]` in block B.

jeffersonrabb:
> To make a "proactive" solution, we probably need to re-determine the visibility of all the `More` buttons every time any of them is clicked. The other, "passive" solution would be accepting an empty response as valid and then hiding the button, without rendering any more posts. The latter one is much easier to implement, but IMO it also introduces bad UX. The first one I don't really know how to approach as it would probably involve some complicated DB queries?

This is a tough one! On our testing sites with relatively small numbers of posts it's quite easy to exhaust the pool of posts and experience the suboptimal UI of clicking the button and getting nothing. In a real-world site (especially a news site) the pool is probably enormous—hundreds or thousands of posts—so you would have to click the Load More button many times to exhaust. Because of this, I'd be inclined to call this an edge case and implement your "passive" solution.

jeffersonrabb:
> IMHO the page-wide exclusion-synchronizing seems confusing to me as a reader. The clearest behavior would be to load `[4, 5]` in block A and also `[4]` in block B.

Here's an example of how this might seem like an error:

Page load:
Screen Shot 2019-12-10 at 6 04 49 PM

After clicking Load More on the right:
Screen Shot 2019-12-10 at 6 06 13 PM

PR #271, opened 2019-12-09
Fixes for TC's phpcs

obenland:
Fixes phpcs errors when trying to merge NHA to wp.com

See D36505-code

### DIFF
--- a/class-newspack-blocks-api.php
+++ b/class-newspack-blocks-api.php
@@ -61,13 +61,13 @@ class Newspack_Blocks_API {
 	 * Get thumbnail featured image source for the rest field.
 	 *
 	 * @param array $object The object info.
-	 * @return array Featured image.
+	 * @return array | bool Featured image if available, false if not.
 	 */
 	public static function newspack_blocks_get_image_src( $object ) {
 		$featured_image_set = [];
 
 		if ( 0 === $object['featured_media'] ) {
-			return $featured_image_set;
+			return false;
 		}
 
 		// Landscape image.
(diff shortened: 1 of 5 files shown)


class-newspack-blocks-api.php
```diff
@@ -60,11 +60,14 @@ public static function register_rest_fields() {
...
-	 * @param Array $object  The object info.
+	 * @param array $object The object info.
+	 * @return array Featured image.
 	 */
 	public static function newspack_blocks_get_image_src( $object ) {
+		$featured_image_set = [];
+
 		if ( 0 === $object['featured_media'] ) {
-			return;
+			return $featured_image_set;
```
jeryj: Will changing this from returning `null` to an empty array have any unintended consequences? I searched the repo for  `newspack_blocks_get_image_src` but only found it referenced here.

jeffersonrabb: Good  catch, I think you may be right. This value is handled here  https://github.com/Automattic/newspack-blocks/blob/master/src/blocks/homepage-articles/edit.js#L102-L131, and it looks like an empty array will in fact bork the logic. I'll make the change.

jeryj: Thanks for referencing that code. I wasn't sure where it ended up being used.

jeffersonrabb: Addressed in https://github.com/Automattic/newspack-blocks/pull/271/commits/c4f77090636fab49f1c7f4458c0dde64c0d22be9. Could you verify this causes no new TC issues?

obenland: It'll be fine, the original complained was that the variable was undefined, which is still fixed. Thanks for catching that, it looked to me like it was just a filter callback.

PR #276, opened 2019-12-11
NHA: `More…` button improvements

### LINKED ISSUE #275: Blog Posts Block: "More" button consistence
`Continue reading` link and the `More` button look different between the editor and the frontend. Can we make them the same?

retrofox:
### All Submissions:

This PR implements several improvements to the `More…` button:

* Adds the needed CSS classes in order to try to show the button visually equivalent in both sides (Frontend and Editor canvas)

* Set the same button-text in both sides

* Default text: `Load more articles`. It's debatable and easily changeable.

* Be able to set a custom text for the button.

Closes #275 

## Testing the Styles
Check styles on both sides.

**BEFORE**

**Editor Canvas (Twenty Twenty)**

**Frontend (Twenty Twenty)**

**AFTER**

**Editor Canvas (Twenty Nineteen)**

**Frontend (Twenty Twenty)**
Same as **before**

**Editor Canvas (Twenty Nineteen)**

**Frontend (Twenty Nineteen)**

**Editor Canvas (Maywood)**

**Frontend (Twenty Maywood)**

## Testing button custom text

* Confirm that an additional field appears when it enables the `Show "More" Button` 

* Editing the `Button description` filed should update the button text on the fly.

* Save and confirm that the custom text is propagated in the frontend page.

* Finally, confirm that an empty value sets the default text

### DIFF
--- a/src/blocks/homepage-articles/block.json
+++ b/src/blocks/homepage-articles/block.json
@@ -35,8 +35,8 @@
       "default": false
     },
     "moreButtonText": {
-	  "type": "string",
-	  "default": ""
+      "type": "string",
+      "default": ""
     },
     "showAuthor": {
       "type": "boolean",
--- a/src/blocks/homepage-articles/block.json
+++ b/src/blocks/homepage-articles/block.json
@@ -34,6 +34,10 @@
       "type": "boolean",
       "default": false
     },
+    "moreButtonText": {
+	  "type": "string",
+	  "default": ""
+    },
     "showAuthor": {
       "type": "boolean",
       "default": true
--- a/src/blocks/homepage-articles/edit.js
+++ b/src/blocks/homepage-articles/edit.js
@@ -314,23 +314,12 @@ class Edit extends Component {
 						/>
 					) }
 					{ ! specificMode && (
-						<Fragment>
-							<ToggleControl
-								label={ __( 'Show "More" Button', 'newspack-blocks' ) }
-								checked={ moreButton }
-								onChange={ () => setAttributes( { moreButton: ! moreButton } ) }
-								help={ __('Only available for non-AMP requests.', 'newspack-blocks') }
-							/>
-
-							{ moreButton && (
-								<TextControl
-									value={ moreButtonText }
-									label={ __( 'Button text', 'newspack-blocks' ) }
-									onChange={ ( moreButtonText ) => setAttributes( { moreButtonText } ) }
-									placeholder={ __( 'Load more posts', 'newspack-blocks' ) }
-								/>
-							) }
-						</Fragment>
+						<ToggleControl
+							label={ __( 'Show "More" Button', 'newspack-blocks' ) }
+							checked={ moreButton }
+							onChange={ () => setAttributes( { moreButton: ! moreButton } ) }
+							help={ __( 'Only available for non-AMP requests.', 'newspack-blocks' ) }
+						/>
 					) }
 				</PanelBody>
 				<PanelBody title={ __( 'Featured Image Settings', 'newspack-blocks' ) }>
@@ -617,12 +606,14 @@ class Edit extends Component {
 				{ ! specificMode && latestPosts && moreButton && (
 					<div className="editor-styles-wrapper">
 						<div className="wp-block-button">
-							<button className="wp-block-button__link" type="button">
-								{ moreButtonText && moreButtonText.length
-									? moreButtonText
-									: __( 'Load more posts', 'newspack-blocks' )
-								}
-							</button>
+							<RichText
+								placeholder={ __( 'Load more posts', 'newspack-blocks' ) }
+								value={ moreButtonText }
+								onChange={ value => setAttributes( { moreButtonText: value } ) }
+								className="wp-block-button__link"
+								keepPlaceholderOnFocus
+								allowedFormats={ [] }
+							/>
 						</div>
 					</div>
 				) }
--- a/src/blocks/homepage-articles/edit.js
+++ b/src/blocks/homepage-articles/edit.js
@@ -601,9 +601,14 @@ class Edit extends Component {
 				</div>
 
 				{ ! specificMode && latestPosts && moreButton && (
-					<button className="button" type="button">
-						{ __( 'More…', 'newspack-blocks' ) }
-					</button>
+					<div className="editor-styles-wrapper">
+						<button
+							className="wp-block-button__link"
+							type="button"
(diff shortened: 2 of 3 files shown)


src/blocks/homepage-articles/block.json
```diff
@@ -34,6 +34,10 @@
       "type": "boolean",
       "default": false
     },
+    "moreButtonText": {
+	  "type": "string",
+	  "default": ""
```
obenland: Looks like indentation is off? json files use spaces rather than tabs, is that maybe it?

src/blocks/homepage-articles/edit.js
```diff
@@ -294,12 +296,23 @@ class Edit extends Component {
...
+								label={ __( 'Show "More" Button', 'newspack-blocks' ) }
+								checked={ moreButton }
+								onChange={ () => setAttributes( { moreButton: ! moreButton } ) }
+								help={ __('Only available for non-AMP requests.', 'newspack-blocks') }
+							/>
+
+							{ moreButton && (
+								<TextControl
+									value={ moreButtonText }
+									label={ __( 'Button description', 'newspack-blocks' ) }
```
obenland: "Button text" maybe?

src/blocks/homepage-articles/edit.js
```diff
@@ -294,12 +296,23 @@ class Edit extends Component {
...
+									onChange={ ( moreButtonText ) => setAttributes( { moreButtonText } ) }
+									placeholder={ __( 'Load more articles', 'newspack-blocks' ) }
```
obenland: Maybe an opportunity to switch to "posts"?

src/blocks/homepage-articles/view.php
```diff
@@ -119,7 +119,13 @@ function( $attribute ) {
 			if ( ! Newspack_Blocks::is_amp() && $has_more_pages && boolval( $attributes['moreButton'] ) ) :
 				?>
 				<button type="button" data-load-more-btn data-load-more-url="<?php echo esc_url( $articles_rest_url ); ?>">
-					<?php _e( 'Load more articles', 'newspack-blocks' ); ?>
+				<?php
+				if ( isset( $attributes['moreButtonText'] ) && ! empty( $attributes['moreButtonText'] ) ) {
```
obenland: The `! empty()` check is enough here. If it's not empty, it's also set.

src/blocks/homepage-articles/edit.js
```diff
@@ -312,12 +314,23 @@ class Edit extends Component {
...
+						<div>
+							<ToggleControl
+								label={ __( 'Show "More" Button', 'newspack-blocks' ) }
+								checked={ moreButton }
+								onChange={ () => setAttributes( { moreButton: ! moreButton } ) }
+								help={ __('Only available for non-AMP requests.', 'newspack-blocks') }
+							/>
+
+							{ moreButton && (
+								<TextControl
```
jeffersonrabb: If we're emulating the `Button` block, the UI would be more Gutenberg-esque if the text were editable in place via a `RichText` component rather than in the sidebar.

src/blocks/homepage-articles/edit.js
```diff
@@ -312,12 +314,23 @@ class Edit extends Component {
 						/>
 					) }
 					{ ! specificMode && (
-						<ToggleControl
-							label={ __( 'Show "More" Button', 'newspack-blocks' ) }
-							checked={ moreButton }
-							onChange={ () => setAttributes( { moreButton: ! moreButton } ) }
-							help={ __( 'Only available for non-AMP requests.', 'newspack-blocks' ) }
-						/>
+						<div>
```
jeffersonrabb: What's the reason for this wrapper div?

src/blocks/homepage-articles/block.json
```diff
@@ -34,6 +34,10 @@
       "type": "boolean",
       "default": false
     },
+    "moreButtonText": {
+      "type": "string",
+      "default": ""
```
jeffersonrabb: Any reason not to use "Load more posts" as the proper default vs. implementing default handling logic in code?

retrofox: It's about the way that I've implemented to translate, or not, the text.
When it's empty, the component sets and translate the default value:

```es6
								{ moreButtonText && moreButtonText.length
									? moreButtonText
									: __( 'Load more posts', 'newspack-blocks' )
								}
```

But when the text is defined, it is not translated.

src/blocks/homepage-articles/edit.js
retrofox: because it's possible to get two sibling components. In that case, we need a wrapper for them. Maybe I'm missing something.

src/blocks/homepage-articles/edit.js
retrofox: Are you suggesting remove this implementation re-implement the button using the `core/button`? Sounds good to me.
If so, maybe we will have to deal with the link part of it, as well as the rest of the block options.
Does it fit well in the posts block? format, alignment, etc.

src/blocks/homepage-articles/edit.js
jeffersonrabb: Did you mean "impossible"? (If so, you could use `<Fragment>` instead, which allows you to have  sibling components without additional DOM elements.)

src/blocks/homepage-articles/block.json
jeffersonrabb: Ah, that makes sense. I wonder if this reveals a shortcoming in extracting the attributes to a JSON file — lack of translatable defaults? Definitely out of this PR's scope though :-)

src/blocks/homepage-articles/edit.js
retrofox: I mean to `possible` because the second one depends on the `moreButton` var:

```es6
{ moreButton && (
    <TextControl
        // ...
```

`<Fragment>` sounds correct. Thanks.

src/blocks/homepage-articles/edit.js
jeffersonrabb: Proposed fix: https://github.com/Automattic/newspack-blocks/pull/276/commits/30348515c4bc02742180b5d790c6a4a6ef0f25eb

src/blocks/homepage-articles/edit.js
jeffersonrabb: Ah  gotcha, sorry I misunderstood!

Issue #277, opened 2019-12-11
NHA: Edit posts title of post breaks the app


retrofox:
**Describe the bug**
1) Create a post
2) create an NHA block, which will include the post which contains the block 😱 
3) Edit post title
4) Save the post

![image](https://user-images.githubusercontent.com/77539/70658914-f901b980-1c2c-11ea-986b-a5862009e340.png)

![image](https://user-images.githubusercontent.com/77539/70658921-fc954080-1c2c-11ea-81c8-faecacd13a4d.png)

PR #278, opened 2019-12-11
Handle Post Title As String

jeffersonrabb:
### All Submissions:

As reported in https://github.com/Automattic/newspack-blocks/issues/277, if a Post contains an instance of a block which displays the  current Post, editing the Post title causes the block to break. Normally `post.title` is an object, but after an edit to the current post it comes back as a simple string. The change in this PR accounts for this situation.

Closes https://github.com/Automattic/newspack-blocks/issues/277.

1. Create a post, save and publish.
2. Add Homepage Posts block. Observe the top Post shown is the  current one.
3. Edit the title of the current Post  and save. 
4. Observe the block reflects the title change and doesn't crash.

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/edit.js
+++ b/src/blocks/homepage-articles/edit.js
@@ -109,7 +109,7 @@ class Edit extends Component {
 		};
 
 		const authorNumber = post.newspack_author_info.length;
-
+		const postTitle =  this.titleForPost( post );
 		return (
 			<article
 				className={ post.newspack_featured_image_src ? 'post-has-image' : null }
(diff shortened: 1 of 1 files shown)


src/blocks/homepage-articles/edit.js
```diff
@@ -109,7 +109,7 @@ class Edit extends Component {
 		};
 
 		const authorNumber = post.newspack_author_info.length;
-
+		const postTitle =  this.titleForPost( post );
```
retrofox: minor: extra space there

jeffersonrabb: Oops, ran the whole file through Prettier: https://github.com/Automattic/newspack-blocks/pull/278/commits/f39623eabe40a70ca16a49692f9e28b846de72bd

retrofox:
It shows the title, but a few seconds after it gets back to the previous one. Hard reset shows the right one.

jeffersonrabb:
> It shows the title, but a few seconds after it gets back to the previous one. Hard reset shows the right one.

I'm finding the title reverts while updating is in flight, then corrects once it's finished. Video example: https://cloudup.com/cPDHdx2iFNn

It seems possible this is a problem that goes beyond the block, maybe related to `getEntityRecords`. We could explore 1) excluding the current post from the block (probably a good idea anyway), 2) Locking updates during save operations 3) Exploring if something weird is going on with `getEntityRecords`. I suggest we land this PR to resolve the serious problem—the crash—and come back to this in a later PR.

PR #285, opened 2019-12-12
NHA: Add styles to hide `.updated` class

### LINKED ISSUE #283: Homepage Blocks: Assumes the theme is hiding `.updated`
Right now the Homepage/Posts block assumes that the theme is hiding `.updated` dates, but if it isn't, they incorrectly display:

image

The block should include styles to hide this.

laurelfulford:
### All Submissions:

Many themes output both a 'published' and 'updated' time for posts, and hide the latter.

The NHA block does this too, but it has relied on the theme hiding the updated time. Because of this, it's visible with themes that don't (like Twenty Twenty). This PR adds styles to hide it to the block.

Closes #283.

1. Switch to a theme that doesn't hide the `.updated` class (like Twenty Twenty).
2. Add a Homepage Block to a page.
3. Pick a post displayed by the block that was published earlier than today, and edit it. Save and publish.
4. View the block on the front end; you should see two dates, like in the first post below:

image

6. Confirm that the second date no longer displays, and only the one representing when the post was originally published is shown:

image

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/view.scss
+++ b/src/blocks/homepage-articles/view.scss
@@ -190,6 +190,10 @@
 		.byline:not(:last-child) {
 			margin-right: 1.5em;
 		}
+
+		.updated:not(.published) {
+			display: none;
+		}
 	}
 
 	.avatar {


laurelfulford:
Thanks!

PR #289, opened 2019-12-12
Homepage Posts: Deduplicate posts in editor when there are multiple blocks 

georgeh:
### All Submissions:

This PR changes the editor to match the de-duplication logic on the front-end.

This changes the Homepage Articles block to use a `wp.data` store to query instead of `getEntityRecords`. The store takes requests and returns a list of posts that are unique to the block. If the same post would show up in multiple blocks, the first block on the page will get it and the rest of the blocks will not.

1. Add multiple Homepage Articles blocks to the page
2. Configure them to return the same post
3. Confirm that the same post does not show up multiple times.
4. Check that in Specific Post Mode, the posts are returning correctly


src/blocks/homepage-articles/edit.js
```diff
@@ -78,7 +76,32 @@ const coverIcon = (
...
+					author: authors,
+					tags,
+			  },
+		value => ! isUndefined( value )
+	);
+	return latestPostsQuery;
+};
+
 class Edit extends Component {
+	constructor( props ) {
```
jeffersonrabb: Wondering if the initial `updateCriteria` should be in `componentDidMount` rather than the constructor. I've never seen it fail, but I'm curious if there could ever be a situation where the constructor is  too early?

src/blocks/homepage-articles/edit.js
```diff
@@ -287,24 +299,24 @@ class Edit extends Component {
 					{ postsToShow && (
 						<QueryControls
 							numberOfItems={ postsToShow }
-							onNumberOfItemsChange={ value => setAttributes( { postsToShow: value } ) }
+							onNumberOfItemsChange={ postsToShow => setAttributes( { postsToShow } ) }
```
jeffersonrabb: I'm having trouble with the `Number of  Items`  setting. 

- After adding a block I can successfully lower the count but raising it  has no effect.
- After lowering it, if I raise it back  up the highest I can get is the original count (3)
- There seems to  be a problem returning  to a value that has been selected earlier.

All of these examples are shown in this video: https://cloudup.com/cihTdL9Q0rQ

src/blocks/homepage-articles/edit.js
```diff
@@ -24,16 +24,14 @@ import {
 	RangeControl,
 	Toolbar,
 	ToggleControl,
-	Dashicon,
 	Placeholder,
 	Spinner,
```
jeffersonrabb: The `Placeholder` with the `Spinner` component is supposed  to  show before the block has retrieved any posts. Spinner state  indicates loading while `Sorry, no posts were found` indicates the retrieve is complete and no posts  are available. On this branch, only the "Sorry" state is ever seen.

src/blocks/homepage-articles/edit.js
```diff
@@ -287,24 +299,24 @@ class Edit extends Component {
...
+							onSpecificModeChange={ specificMode => setAttributes( { specificMode } ) }
 							specificPosts={ specificPosts }
-							onSpecificPostsChange={ value => setAttributes( { specificPosts: value } ) }
+							onSpecificPostsChange={ specificPosts => setAttributes( { specificPosts } ) }
 							authors={ authors }
-							onAuthorsChange={ value => setAttributes( { authors: value } ) }
+							onAuthorsChange={ authors => setAttributes( { authors } ) }
 							categories={ categories }
-							onCategoriesChange={ value => setAttributes( { categories: value } ) }
+							onCategoriesChange={ categories => setAttributes( { categories } ) }
```
jeffersonrabb: When Categories, Tags, or Author changes  the block should switch  to  Loading  state until the new results are available. This video shows the correct behavior on `master`: https://cloudup.com/c2D5XhhGfmm

src/blocks/homepage-articles/store.js
```diff
@@ -0,0 +1,266 @@
+import apiFetch from '@wordpress/api-fetch';
```
jeffersonrabb: Is this store file identical to the one in development for Query block? If so, it might be good to move  it  out of the block's directory so it can be shared between the two blocks. Maybe `/src/store/index.js`?

src/blocks/homepage-articles/store.js
```diff
@@ -0,0 +1,266 @@
...
+		};
+	},
+	receivePosts( clientId, posts ) {
+		return {
+			type: RECEIVE_POSTS,
+			clientId,
+			posts,
+		};
+	},
+	updateBlocks( blocks ) {
```
jeffersonrabb: Is the `updateBlocks` applicable  to Homepage Articles or Query only?

src/blocks/homepage-articles/store.js
```diff
@@ -0,0 +1,266 @@
...
+			type: UPDATE_BLOCKS,
+			blocks,
+		};
+	},
+};
+
+/**
+ * Returns the Query blocks that appear before the current one on the page
+ *
+ * @param {Block[]} orderedBlocks
```
jeffersonrabb: Reordering  blocks doesn't seem to trigger a requery. In this video it would be expected that the content  of the two blocks change after the reorder: https://cloudup.com/cHiGAV5XpDi

src/blocks/homepage-articles/editor.js
```diff
@@ -3,5 +3,7 @@
  */
 import { registerBlockType } from '@wordpress/blocks';
 import { name, settings } from '.';
+import { registerQueryStore } from './store';
 
 registerBlockType( `newspack-blocks/${ name }`, settings );
+registerQueryStore();
```
jeffersonrabb: What's the rationale for registering the store in this file vs. the block's `edit.js`? Is this done to ensure registration occurs only once per page load?

src/blocks/homepage-articles/edit.js
georgeh: This seems tied to the way that `wp.data` memoizing selectors with resolvers. It works the first time you hit a number that you were looking for but stops updating if you try it again. I'm working on the right way to skip the memoization with some extra criteria params.

src/blocks/homepage-articles/edit.js
georgeh: This is tougher now that the loading state is managed by the store's resolvers, because that gets in the way of resolving the query immediately (vs having no results). Once we get a fix in, that should also apply to changing the criteria per your next comment.

src/blocks/homepage-articles/store.js
georgeh: It's been modified a bit but once this is solid we should be able to get both blocks to be able to use the same store.

src/blocks/homepage-articles/store.js
georgeh: It's block specific right now, it filters for the block named on line 10. If I have a Homepage Articles block and a Query block on the same page, should they both display the same posts or should they deduplicate with each other?

src/blocks/homepage-articles/store.js
georgeh: After some changes, it's triggering half the time now. I think it is related to the memoization issue above.

src/blocks/homepage-articles/editor.js
georgeh: Yes, I more or less assume that we only want to register a store once. We also only want one `subscribe()` call to `wp.data` to watch the block order for when the blocks change

src/blocks/homepage-articles/store.js
jeffersonrabb: That's a great question. In this particular case there probably won't be a use case for Query and Homepage Articles to co-exist. However for other blocks like Carousel the question is definitely valid. I think all blocks that use de-duplication logic should draw from the same  pool.

src/blocks/homepage-articles/edit.js
georgeh: This is working with the latest updates.

src/blocks/homepage-articles/edit.js
georgeh: Fixed in the latest commits!

src/blocks/homepage-articles/edit.js
georgeh: I tried it with `componentDid Mount` and it worked just as well, but I was able to remove it entirely. I still have a `constructor` but that's to set the initial state now, instead.

src/blocks/homepage-articles/store.js
georgeh: This is fixed in the latest commits.

laurelfulford:
One other thing I forgot to note in my original review:

I noticed the 'No Posts Found' message coming up when posts were still being loaded. On my localhost, the post loading is really slow, so it's very noticeable. Using Jurassic Ninja, it's a lot faster, so it is really a flash -- but it can be confusing:

Shot of a page after the first block loads -- the spinner seems to show for all until the first posts load, then the 'No Posts' message show as the rest of the blocks update:

image

The lack of spinner can be pretty confusing; and on slower connections, it can give the impression that things are moving even slower than they are, just because it's not clear anything is happening.

Ideally the spinner would be showing for a block any time it's updating, though I can see that being tough to do. 

If it's too much, maybe we can do something else to improve the experience -- like changing the 'No Posts' message to something a bit more ambiguous? (This kind of shifts the bad experience to folks who are building blocks that are actually empty, but on the flip side, this will happen a lot less often than blocks loading). I'm happy to brainstorm around this!

georgeh:
@laurelfulford @jeffersonrabb could you take the latest version for a spin? I was able to simplify a lot and now the blocks seem to be working well together.

src/blocks/homepage-articles/edit.js
georgeh: Tested and seems to be working

src/blocks/homepage-articles/edit.js
```diff
@@ -229,6 +230,7 @@ class Edit extends Component {
 	);
 
 	renderInspectorControls = () => {
+		// const { latestPosts } = this.state;
```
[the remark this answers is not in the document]
claudiulodro: Looks like it was removed before I finished reviewing! Now that's service! :P

PR #228, opened 2019-11-25
homepage-articles: rename title and descriptions

retrofox:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-theme/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

This PR renames the `title` and `descriptions` of the Homepage Articles block. 

Screen Shot 2019-11-25 at 11 57 08 AM

it's a simpler approach compared with https://github.com/Automattic/newspack-blocks/pull/227.

1. Apply this brown changes
3. Tests with blocks already created before these changes.
4. Confirm that everything is working as expected.
5. Create a new `Homepage posts` block.
6. Check that `article` has been replaced by `post`: inserter, block preview, etc.

PR #290, opened 2019-12-12
Columns styles: Make sure nested columns don't inherit border styles

laurelfulford:
### All Submissions:

**Note:** there will be a complementary theme PR to make sure this update doesn't break the border styles in style packs 3 and 5. This one should not be merged until that one is also ready to merge.

Edited to add: this is the related theme PR: https://github.com/Automattic/newspack-theme/pull/632

When you create a column block, give it the border styles, and nest a second column block inside of it, that second block incorrectly inherits those styles. This PR fixes that.

Closes #228.

1. On a page, add a column block and apply the border styles. Inside one of the cells, nest a second column block.
2. For testing regressions, also add a second column block with the border styles, but that's not nested.
3. Use any style pack but 3 and 5 (since they have different border styles, which this PR breaks).
4. Note in the first blocks that the nested columns block is also inheriting the border styles on the front-end and in the editor:

image

6. Confirm that the nested columns block is no longer inheriting the styles:

image

7. Shrink down the browser window to make sure the borders under each column display where expected:

image

### Other information:

### DIFF
--- a/src/block-styles/core/columns/editor.scss
+++ b/src/block-styles/core/columns/editor.scss
@@ -22,7 +22,7 @@
 	}
 
 	&.is-style-borders {
-		[data-type="core/column"] {
+		& > .editor-inner-blocks > .editor-block-list__layout > [data-type="core/column"] {
 			border-bottom: 1px solid $color__border;
 			position: relative;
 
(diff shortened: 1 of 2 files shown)


laurelfulford:
Thanks!

PR #298, opened 2019-12-17
Handle Post Subtitle

adekbadek:
This adds **post subtitle** (Automattic/newspack-theme#635) handling for Homepage Articles block. 

Post subtitle can now appear in the block:

image

And this can be toggled in the block settings:

image

### DIFF
--- a/src/blocks/homepage-articles/templates/article.php
+++ b/src/blocks/homepage-articles/templates/article.php
@@ -71,11 +71,18 @@ call_user_func(
 				</div>
 				<?php
 			endif;
-			if ( '' === $attributes['sectionHeader'] ) :
+
+			$is_section_header = '' === $attributes['sectionHeader'];
+			if ( $is_section_header ) :
 				the_title( '<h2 class="entry-title"><a href="' . esc_url( get_permalink() ) . '" rel="bookmark">', '</a></h2>' );
 			else :
 				the_title( '<h3 class="entry-title"><a href="' . esc_url( get_permalink() ) . '" rel="bookmark">', '</a></h3>' );
 			endif;
+			?>
+			<div class="newspack-post-subtitle newspack-post-subtitle--in-homepage-block <?php echo $is_section_header ? 'newspack-post-subtitle--small' : '' ?>">
+				<?php echo get_post_meta(get_the_ID(), 'newspack_post_subtitle', true); ?>
+			</div>
+			<?php
 			if ( $attributes['showExcerpt'] ) :
 				the_excerpt();
 			endif;
--- a/src/blocks/homepage-articles/templates/article.php
+++ b/src/blocks/homepage-articles/templates/article.php
@@ -79,9 +79,13 @@ call_user_func(
 				the_title( '<h3 class="entry-title"><a href="' . esc_url( get_permalink() ) . '" rel="bookmark">', '</a></h3>' );
 			endif;
 			?>
-			<div class="newspack-post-subtitle newspack-post-subtitle--in-homepage-block <?php echo $is_section_header ? 'newspack-post-subtitle--small' : '' ?>">
-				<?php echo get_post_meta(get_the_ID(), 'newspack_post_subtitle', true); ?>
-			</div>
+			<?php
+			if ( $attributes['showSubtitle'] ) :
+				?>
+				<div class="newspack-post-subtitle newspack-post-subtitle--in-homepage-block <?php echo esc_html( $is_section_header ? 'newspack-post-subtitle--small' : '' ); ?>">
+					<?php echo esc_html( get_post_meta( get_the_ID(), 'newspack_post_subtitle', true ) ); ?>
+				</div>
+			<?php endif; ?>
 			<?php
 			if ( $attributes['showExcerpt'] ) :
 				the_excerpt();
@@ -95,7 +99,7 @@ call_user_func(
 							echo wp_kses(
 								newspack_blocks_format_avatars( $authors ),
 								array(
-									'img' => array(
+									'img'      => array(
 										'class'  => true,
 										'src'    => true,
 										'alt'    => true,
@@ -105,8 +109,8 @@ call_user_func(
 										'srcset' => true,
 									),
 									'noscript' => array(),
-									'a' => array(
-										'href'  => true,
+									'a'        => array(
+										'href' => true,
 									),
 								)
 							);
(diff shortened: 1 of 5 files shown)


laurelfulford:
Thanks for diving into this, @adekbadek! So excited to see this added :) A couple things I noticed while playing around with this:

* With both this PR and Automattic/newspack-theme#635 on my test site, I noticed the subtitles were showing on the front-end, even though they weren't set in the backend, and I can turn them on in the backend, but the option doesn't save. I think it's because showSubtitle needs a default value in: https://github.com/Automattic/newspack-blocks/blob/master/src/blocks/homepage-articles/block.json
* The 'Type Scale' option for the block should also affect the Subtitle text. Let me know if you'd like me to take a swing at this bit! 

image

* Total nitpick, but what do you think about moving the 'Show Subtitle' above the 'Show Excerpt', in the sidebar, since it appears above it in the block? 

Also just a general question for everyone: this option is visible whether or not the post meta field is available in the theme, and whether or not you're using a different theme all together. This feels... not right. I don't have a great alternative in mind -- I don't think we want to be checking for the theme in the plugin before showing it either, and pulling out the subtitle functionality into a separate plugin might be overkill (especially since it's kind of a temporary solution as we work towards doing that kind of thing with a block). So just flagging in case someone has any ideas how to improve this that are better than mine :)

src/blocks/homepage-articles/templates/article.php
```diff
@@ -71,11 +71,18 @@ class="post-has-image"
...
+
+			$is_section_header = '' === $attributes['sectionHeader'];
+			if ( $is_section_header ) :
 				the_title( '<h2 class="entry-title"><a href="' . esc_url( get_permalink() ) . '" rel="bookmark">', '</a></h2>' );
 			else :
 				the_title( '<h3 class="entry-title"><a href="' . esc_url( get_permalink() ) . '" rel="bookmark">', '</a></h3>' );
 			endif;
+			?>
+			<div class="newspack-post-subtitle newspack-post-subtitle--in-homepage-block <?php echo $is_section_header ? 'newspack-post-subtitle--small' : '' ?>">
+				<?php echo get_post_meta(get_the_ID(), 'newspack_post_subtitle', true); ?>
```
claudiulodro:
```suggestion
				<?php echo esc_html( get_post_meta( get_the_ID(), 'newspack_post_subtitle', true ) ); ?>
```

[Same as the other PR](https://github.com/Automattic/newspack-theme/pull/635/files#r359095544), this output needs some escaping. 

This bit here also needs some space to follow PHPCS rules. It was mentioned at the other PR also: you can just do `composer install` to get PHPCS automatically installed and set up, and it'll run and analyze the code you're about to commit whenever you try and commit.

src/blocks/homepage-articles/templates/article.php
```diff
@@ -71,11 +71,18 @@ class="post-has-image"
...
-			if ( '' === $attributes['sectionHeader'] ) :
+
+			$is_section_header = '' === $attributes['sectionHeader'];
+			if ( $is_section_header ) :
 				the_title( '<h2 class="entry-title"><a href="' . esc_url( get_permalink() ) . '" rel="bookmark">', '</a></h2>' );
 			else :
 				the_title( '<h3 class="entry-title"><a href="' . esc_url( get_permalink() ) . '" rel="bookmark">', '</a></h3>' );
 			endif;
+			?>
+			<div class="newspack-post-subtitle newspack-post-subtitle--in-homepage-block <?php echo $is_section_header ? 'newspack-post-subtitle--small' : '' ?>">
```
claudiulodro: This should check whether subtitles are toggled on before outputting the subtitle. (e.g. `if ( $attributes['showSubtitle'] ) { /* output subtitle */ }`, otherwise the subtitle toggle won't control anything.

[block.json](https://github.com/Automattic/newspack-blocks/blob/fix/author-id/src/blocks/homepage-articles/block.json) also needs to be updated whenever a new block attribute is added to handle defaults, etc.

adekbadek: that's right, I've only handled it the editor 🤦‍♂
fixed now

src/blocks/homepage-articles/templates/article.php
```diff
@@ -71,11 +71,22 @@ class="post-has-image"
...
+			if ( $is_section_header ) :
 				the_title( '<h2 class="entry-title"><a href="' . esc_url( get_permalink() ) . '" rel="bookmark">', '</a></h2>' );
 			else :
 				the_title( '<h3 class="entry-title"><a href="' . esc_url( get_permalink() ) . '" rel="bookmark">', '</a></h3>' );
 			endif;
+			?>
+			<?php
+			if ( $attributes['showSubtitle'] ) :
+				?>
+				<div class="newspack-post-subtitle newspack-post-subtitle--in-homepage-block <?php echo esc_html( $is_section_header ? 'newspack-post-subtitle--small' : '' ); ?>">
```
claudiulodro: For content within an HTML attribute (`class`, `value`, etc.) you want to use [`esc_attr`](https://developer.wordpress.org/reference/functions/esc_attr/).

claudiulodro: Actually, since this is a static string that doesn't ever get translated, you don't even need escaping. My note above about `esc_attr` is still true, but you don't need to escape this output since users can't affect it in any way.

laurelfulford:
Thanks for adjusting the font sizes, @adekbadek! 

I noticed a couple things in reviewing:

For the larger font sizes, the dek seems pretty big; I think it could probably max out at the size it is on single posts (similar to the excerpts, which get as large as the standard font size):

image

The line height/vertical spacing seems a bit off in some cases -- I think tightening up the line height of the dek could help: 

image

As the font sizes get smaller, the dek gets smaller than the except; I think it should probably stop at the same size, or a hair larger (even if it's roughly the size of the headline, the bold/italics will help differentiate them, and it'll be easier to read):

image

Lastly, the dek sizes don't always seem proportionate on smaller screens:

image

I know I'm getting a bit into the weeds here, so let me know if you'd like me to mess around with the styles myself. Thanks!

adekbadek:
@laurelfulford Thanks for spotting these, should be better now!
@claudiulodro [f60ae9](https://github.com/Automattic/newspack-blocks/pull/298/commits/f6d0ae981d4697ca83265df18233339636693385) adds `post-subtitles` theme support detection

laurelfulford:
The font size updates look great 👍 Thanks @adekbadek!

PR #307, opened 2019-12-19
Homepage Posts: Fix issues with translation

laurelfulford:
### All Submissions:

This PR fixes a handful of issues with translations in the Homepage Posts block:

* It restores the front-end translation for the block (for the words 'by' and 'and' in the byline). I think this was lost in the recent restructuring.
* It adds a translation for the strings 'by' and 'and' in Spanish to the editor (which were missing previously).
* It updates the date in the editor to display in the correct language. 
* It updates the POT file for the whole block.

Please note: I think the current approach to the translations for the editor are not the correct approach, but I haven't managed to get that to work yet.

The editor should either load a json file of translations with the naming convention `{slug}-{language}-{script handle}.json` (which currently works -- in the plugin, it's `newspack-blocks-es_ES-newspack-blocks-editor.json`), or one with an md5 hash, `{slug}-{language}-{md5}.json` (like `newspack-blocks-es_ES-f03802d9ac1ae8f3b85caee4347b93a7.json`). That hash is dynamically generated as part of the .json file name when you run `wp i18n make-json [translation file]`; one .json file is generated per JavaScript file, and the md5 hash is created from the JavaScript file's original name.

I haven't managed to get `wp_set_script_translations()` to pick up the `{slug}-{language}-{md5}.json` files, though it does work fine for `{slug}-{language}-{script handle}.json`. I'm not sure if it's something to do with how the blocks are structured, or I'm doing something wrong. But to get the translation working in the editor for both the donate and homepage posts blocks, I've copy-pasted what's generated in their individual `{slug}-{language}-{md5}.json` files into `newspack-blocks-es_ES-newspack-blocks-editor.json`. 

This works for now, but definitely won't scale as we add more translations and languages. We might be able to get around it by creating a script to create and combine the files, but I'm really curious why it's not working.

More info here: https://developer.wordpress.org/block-editor/developers/internationalization/

1. Set up a Homepage Posts block, and a Donate block on a page.
2. Navigate to Settings > General, and switch your site language to Spanish. 
3. View the blocks on the front end; note that the Donate block is translated, but the Homepage Posts block is not (specifically, the author byline):

image

4. View in the editor, and note a similar issue (except here, the date on the posts block is also not translated):

image

6. View the page on the front-end again; confirm that both the Donate block and Homepage Posts block are now translated (the Donate block helps confirm there are no regressions):

image

7. View the page in the editor, and confirm that the Donate block is still translated, and that the string 'by' and the date are also both in Spanish:

image

### Other information:


jeffersonrabb:
> I haven't managed to get `wp_set_script_translations()` to pick up the `{slug}-{language}-{md5}.json` files

Here's an explanation, but no solution yet 😄. The MD5 is generated from the path of the source file, but the editor is loading the transpiled file, hence the MD5 mismatch and failure to pick up the translations. To illustrate:

- The path of Homepage Posts block's edit file is `src/blocks/homepage-articles/edit.js` and the MD5 is `727631509ad8eec8e2cbe740c6dad199` (using https://www.md5hashgenerator.com/)

- The path of `newspack-block`'s transpiled editor file is `dist/editor.js` —  `53e2a1d5945b8d2b1c35e81ae1e532f3`

To see how this works try this flow:

- Check out this branch.
- Delete or rename `languages/newspack-blocks-es_ES-newspack-blocks-editor.json`
- Set Site Language to Spanish.
- Edit a page with Homepage Articles, observe there is no Spanish text.
- Run `wp i18n make-json languages/newspack-blocks-es_ES.po --no-purge` to generate new JSON files.
- Reload and observe there is still no Spanish text.
- Find the JSON file with the source MD5 (`languages/newspack-blocks-es_ES-727631509ad8eec8e2cbe740c6dad199.json`) and rename to `newspack-blocks-es_ES-53e2a1d5945b8d2b1c35e81ae1e532f3.json`
- Reload the editor and you should see the translated text.

I don't have a solution yet, this is just an explanation of the issue. We may need to cook up a script that combines the generated .json for all editor files into a single file with the correct name. It looks like Jetpack resolved a similar situation in this PR: https://github.com/Automattic/jetpack/pull/11225/files

laurelfulford:
Thanks!

akirk:
@jeffersonrabb have you created the issue yet? The missing `wp_set_script_translations()` causes problems with loading the translations inside the Full Site Editing plugin.

The solution for your problem with the MD5 sums above is to run `wp i18n make-pot` after the transpiled files have been generated. While it means that the `*-727631509ad8eec8e2cbe740c6dad199.json` (source URL) will be created in vain by make-json, it'll also create the `*-53e2a1d5945b8d2b1c35e81ae1e532f3` file.

akirk:
See https://github.com/Automattic/wp-calypso/pull/39554

PR #314, opened 2019-12-23
Homepage Posts: Remove dynamic columns limit

### LINKED ISSUE #305: Blog Posts Block: Columns setting controlled by number of published posts
When adding a Blog Posts Block, the number of columns is restrained by the number of blog published posts there are on the site. For example, if someone wants to display their posts in 3 columns, but only has 1 published post when starting, the number of columns setting does not allow them to make that change. It seems like there has to be at least the number of published posts as desired columns to make that setting save.

**To Reproduce**

1. Count how many published posts there are on the site.
2. Create or edit a page.
3. Add the Blog Posts Block, set the display to be grid (not list), and change the columns setting to be more than the number of published posts.
4. Move the focus away from the columns setting field or publish/update the page and notice that the value entered is not saved.

**Screencast**

https://d.pr/v/4IlXIK/FoEr4COhEX

**Expected behavior**

For the columns value to be saved and respected, regardless of how many published posts there are.

laurelfulford:
### All Submissions:

Right now, if you set the Homepage Posts block to display as grid, you have an option to pick the number of columns you want to show. The maximum number of columns is six, or, if you have fewer posts than that (either based on the number of posts you've chosen to show, or the number of posts your query is returning), the maximum number of columns will be that. 

By default, the number of posts loaded is 3, so that's also your maximum number of columns.

This has been reported as confusing, and it also 'breaks' the ability to type in a column count value, since it gets "corrected" when you move focus away from the column field, and reset to the maximum.

This PR removes the restriction, so the maximum is always 6. Down the road, we should also improve the appearance of the columns when your column count is higher than your article count, but that should probably wait until #300 lands. (Right now, the columns are the same width, but spread out; they should probably all align left).

(There are some other code changes in this file, but they're coming from JS Prettifier).

Closes #305.

2. Add a Homepage Posts block to a page and switch it to 'grid' view.
3. Confirm that the default is still 3 columns.
4. Confirm you can increase the column count, even though you only have three articles:

image

5. Confirm other column settings work.

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/edit.js
+++ b/src/blocks/homepage-articles/edit.js
@@ -40,15 +40,14 @@ import { withSelect } from '@wordpress/data';
 import { compose } from '@wordpress/compose';
 import { decodeEntities } from '@wordpress/html-entities';
 
-let IS_SUBTITLE_SUPPORTED_IN_THEME
-if ( typeof window === 'object' && window.newspackIsPostSubtitleSupported && window.newspackIsPostSubtitleSupported.post_subtitle ) {
-	IS_SUBTITLE_SUPPORTED_IN_THEME = true
+let IS_SUBTITLE_SUPPORTED_IN_THEME;
+if (
+	typeof window === 'object' &&
+	window.newspackIsPostSubtitleSupported &&
+	window.newspackIsPostSubtitleSupported.post_subtitle
+) {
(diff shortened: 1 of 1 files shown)


laurelfulford:
Thanks Jeff!

PR #319, opened 2020-01-06
Add fsevents as a dependency to enable building on node v.12

adekbadek:
### All Submissions:

Add `fsevents` as an explicit dependency to avoid issues with installing node modules on node v. 12. The earlier version was incompatible with node v.12.
`fsevents` is used by webpack's file watching feature internally.

`.nvmrc` file was removed (added in #262), because locking node version to 10 is no longer needed.

1. Set node version to 12 (`nvm use 12`)
2. Install node dependencies (`npm i`)
3. Compile code (`npm run build`)
4. All should work without issues. You _might_ have to remove `node_modules` first

### Other information:


jeffersonrabb:
I ran into some trouble running `npm i`. Full log: 

[newspack-blocks-pr-319.log](https://github.com/Automattic/newspack-blocks/files/4025694/newspack-blocks-pr-319.log)

After reading https://github.com/golang/go/issues/25240 it sounds like this has to do with an issue that occurs when migrating from one Mac to another, where `/usr/local/include` is incorrectly left intact. After executing `sudo mv /usr/local/include /usr/local/include_old`, I was able to install and build successfully.

PR #325, opened 2020-01-09
Columns Block: Remove update block output

### LINKED ISSUE #321: Columns Block: `calc()` is stripped out of inline styles when user doesn't have `unfiltered_html`
In the block, we've reworked the Columns Block to use `calc()` in the inline CSS, as a temporary work-around for a bug.

However, if a user doesn't have `unfiltered_html` ability (anyone but Administrators), the `calc()` style is stripped out, causing the block to throw an error. This has to do with this safe CSS filter here: 

https://developer.wordpress.org/reference/functions/safecss_filter_attr/

... and only happens when you change the widths of the columns (eg. when they are all not the same width as each other). 

One work-around may be to change the columns block to use negative margins on the outside, and use margins/padding to add the gutters for the block, rather than using `calc()` to make each column narrower. A change like this will require existing column blocks to be updated as well, as they'll start throwing errors in the editor as soon as the expected `calc()` value is removed.

There also may be something we can do about the validation for the short-term. 

Related Core issue: https://core.trac.wordpress.org/ticket/46197

laurelfulford:
### All Submissions:

This PR removes the modified Columns block output, since `calc()` is stripped out on save if the current user's account does not have `unfiltered_html` (by default, only Administrators have this). 

It also adjusts the border block styles to continue to display correctly. 

This PR should be tested with the complementary theme PR, https://github.com/Automattic/newspack-theme/pull/685. 

Note: This update will break columns on existing sites, since it changes the output -- they'll display correctly, but the block will throw an error in the editor. Ideally before committing this update we'll have a script to go with it to correct the columns blocks, and roll it out to sites one at a time to make sure we're not introducing issues. 

Closes #321.

1. Apply this PR and the related PR https://github.com/Automattic/newspack-theme/pull/685 to the theme. Run `npm run build` for both.
2. Add columns to the editor with various settings (in column count, width, and using the available styles). In the columns, add paragraph blocks with backgrounds; it's the easiest way to see where the edges of the columns are.
3. On the front-end, test different screen sizes, and make sure the columns are behaving as expected (not wrapping until the window is less than 600px wide).
4. Specifically for the theme PR, navigate to Customizer > Style Packs and switch to Style 3; confirm that the columns block with the border uses a dark dotted line on the front-end and in the editor, rather than a light grey border.
5. Specifically for the theme PR, navigate to Customizer > Style Packs and switch to Style 5; confirm that the columns block with the border uses a dark border on the front-end and in the editor, rather than a light grey border.

### Other information:


jeffersonrabb:
I looked into using Gutenberg `deprecated` to protect  blocks from becoming invalidated, but alas no luck. `deprecated` doesn't seem to play nicely with the `blocks.registerBlockType` filter. It looks like we'll need to resolve broken blocks manually or via a custom script.

laurelfulford:
I did some testing, and searching for `wp-block-columns` or `wp:columns` in the post or page search in WP Admin successfully returns all of the posts/pages with this block. 

Checking live sites, the one with the 'most' columns uses them on less than ten posts and pages -- so it should be pretty quick to manually fix as we roll this out.

laurelfulford:
Thanks for the review, Jeff!

PR #291, opened 2019-12-13
Set up SCSS and JS linting

jeffersonrabb:
### Changes proposed in this Pull Request:

Match `wp-calypso` linting, so synced files no longer raise warnings. 

What I've done so far:

- Added `@wordpress/eslint-plugin`, `stylelint`, `stylelint-config-wordpress`.
- `.eslintrc` as described in https://github.com/WordPress/gutenberg/tree/4786c14803859a046b14e5eb324b1041b8fbfff3/packages/eslint-plugin
- Copied lint npm scripts from https://github.com/Automattic/jetpack/pull/11767/files
- Copied `.stylelintrc` from Calypso (https://github.com/Automattic/wp-calypso/blob/46a9eafeda1d2d74871b527b25765fd7a9f7ae1e/.stylelintrc)

`npm run lint` correctly reports linting issues but errors out.

PR #326, opened 2020-01-10
Fix JS after linting was set up

adekbadek:
### All Submissions:

After #291 set up linting configuration, now it's time to fix the  existing JS files. 
These changes are in most part things like removing unused variables or fixing spacing. Only code refactor was in `AutocompleteTokenField`, where `setState` was used in `componentDidMount`, which is a React antipattern. 

1. Look at the code
2. Look some more
3. If it's all innocuous refactoring, then we're cool

### Other information:


adekbadek:
1. Prettier is the next step in this journey - sorry for not communicating that earlier 🙈 I've opened a draft PR which marries prettier and eslint - #331
2. `react-in-jsx-scope` - you're right, since this code will always run in Gutenberg, then this rule can be switched off (as it's done in [Guntenberg's eslint config](https://github.com/WordPress/gutenberg/blob/master/packages/eslint-plugin/configs/react.js#L31))

PR #338, opened 2020-01-20
Make querySelectorAll loop IE11 compatible

### LINKED ISSUE #315: Load More Button Fails In IE11
**Describe the bug**
The Load More Button functionality fails silently in IE11 (AMP and non-AMP)

**To Reproduce**

Add Homepage Posts block to a page, enable Load More button, publish, view in IE11.

**Screenshots**
Javascript errors for AMP:
Screen Shot 2019-12-24 at 8 46 16 AM

Javascript errors for non-AMP:
Screen Shot 2019-12-24 at 8 53 30 AM

### LINKED ISSUE #332: Multiple Premium Themes: JS error on IE 11 causes site to load blank or breaks the mobile menu toggle
We have received multiple reports from happiness engineers about premium themes and IE11.

The following themes are loading broken or blank, with this error in the console log:

`Object doesn't support property or method 'forEach'`

Broken themes:
* Solar
* Huntt
* Goodz Magazine
* Ippo
* Eris
* Aperitive
* Veni

Same error but broken mobile menu toggle:
* Adaline
* Owari
* Vivaro
* Spatial
* Bari
* Novara

The theme shops pointed out that the issue is on our end.

Looks like the corresponding WordPress.com code can be found here: ​https://github.com/Automattic/wp-calypso/blob/fe44257b9e3bbf6aff1692ef5601eda1d7588d4d/apps/full-site-editing/full-site-editing-plugin/blog-posts-block/newspack-homepage-articles/blocks/homepage-articles/view.js#L18

They said that, it appears to be an easy fix, just have to add a polyfill: ​https://developer.mozilla.org/en-US/docs/Web/API/NodeList/forEach#Browser_Compatibility

obenland:
IE11 and other browsers don't implement `NodeList.forEach()`, so looping over `querySelectorAll()`'s return value causes errors there.

This PR attempts to fix it by using `Array.prototype.forEach` directly, as suggested in https://developer.mozilla.org/en-US/docs/Web/API/NodeList#Example

Fixes #332.
Fixes #315.

### DIFF
--- a/src/blocks/homepage-articles/view.js
+++ b/src/blocks/homepage-articles/view.js
@@ -12,11 +12,14 @@ const fetchRetryCount = 3;
 
 /**
  * Load More Button Handling
+ *
+ * Calls Array.prototype.forEach for IE11 compatibility.
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/NodeList
  */
-
-document
-	.querySelectorAll( '.wp-block-newspack-blocks-homepage-articles.has-more-button' )
-	.forEach( buildLoadMoreHandler );
+Array.prototype.forEach.call(
+	document.querySelectorAll( '.wp-block-newspack-blocks-homepage-articles.has-more-button' ),
+	buildLoadMoreHandler
+);
 
 /**
  * Builds a function to handle clicks on the load more button.


jeffersonrabb:
This resolves the `forEach` problem but now I'm seeing a second IE11 issue breaking Load More: limited support for the URL API https://caniuse.com/#search=URL(): https://github.com/Automattic/newspack-blocks/blob/2c03137ba5f08fd7e0a130caac742ae491fa7eef/src/blocks/homepage-articles/view.js#L53-L56. We'll need to Polyfill or replace.

obenland:
@jeffersonrabb Thanks for the test and feedback! I'll replace that implementation in a separate PR.

PR #340, opened 2020-01-20
HA: Dumb down handling of exclude_ids

obenland:
Fixes IE11 incompatibility with `URL` api.

UI will now always add all IDs of displayed posts, the endpoint will always remove it from its next url.
See https://github.com/Automattic/newspack-blocks/pull/338#issuecomment-576428130

1. Create a block instance with more button.
2. In the front-end, click the more button a few times, make sure the request URL looks sane and posts load as expected.
3. Make sure IE 11 works as expected.

### DIFF
--- a/src/blocks/homepage-articles/view.js
+++ b/src/blocks/homepage-articles/view.js
@@ -51,7 +51,7 @@ function buildLoadMoreHandler( blockWrapperEl ) {
 		const requestURL =
 			btnEl.getAttribute( 'data-next' ) + '&exclude_ids=' + getRenderedPostsIds().join( ',' );
 
-		fetchWithRetry( { url: requestURL, onSuccess, onError }, fetchRetryCount );
+		fetchWithRetry( { url: encodeURI( requestURL ), onSuccess, onError }, fetchRetryCount );
 
 		/**
 		 * @param {Object} data Post data
--- a/src/blocks/homepage-articles/view.js
+++ b/src/blocks/homepage-articles/view.js
@@ -47,12 +47,11 @@ function buildLoadMoreHandler( blockWrapperEl ) {
 		blockWrapperEl.classList.remove( 'is-error' );
 		blockWrapperEl.classList.add( 'is-loading' );
 
-		const requestURL = new URL( btnEl.getAttribute( 'data-next' ) );
+		// Set currently rendered posts' IDs as a query param (e.g. exclude_ids=1,2,3)
+		const requestURL =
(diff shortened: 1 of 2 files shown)


src/blocks/homepage-articles/view.js
```diff
@@ -47,12 +47,11 @@ function buildLoadMoreHandler( blockWrapperEl ) {
...
-		const requestURL = new URL( btnEl.getAttribute( 'data-next' ) );
+		// Set currently rendered posts' IDs as a query param (e.g. exclude_ids=1,2,3)
+		const requestURL =
+			btnEl.getAttribute( 'data-next' ) + '&exclude_ids=' + getRenderedPostsIds().join( ',' );
 
-		// Set currenty rendered posts' IDs as a query param (e.g. exclude_ids=1,2,3)
-		requestURL.searchParams.set( 'exclude_ids', getRenderedPostsIds().join( ',' ) );
-
-		fetchWithRetry( { url: requestURL.toString(), onSuccess, onError }, fetchRetryCount );
+		fetchWithRetry( { url: requestURL, onSuccess, onError }, fetchRetryCount );
```
WunderBart: Because we're dropping the URI constructor, let's encode the `requestURL` with [`encodeURI`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI) here.

src/blocks/homepage-articles/view.js
```diff
@@ -47,12 +47,11 @@ function buildLoadMoreHandler( blockWrapperEl ) {
 		blockWrapperEl.classList.remove( 'is-error' );
 		blockWrapperEl.classList.add( 'is-loading' );
 
-		const requestURL = new URL( btnEl.getAttribute( 'data-next' ) );
+		// Set currently rendered posts' IDs as a query param (e.g. exclude_ids=1,2,3)
+		const requestURL =
+			btnEl.getAttribute( 'data-next' ) + '&exclude_ids=' + getRenderedPostsIds().join( ',' );
```
WunderBart: Are we sure that the `?` separator will always be present in the `data-next` URL?

obenland: Yes, both [the initial value](https://github.com/Automattic/newspack-blocks/blob/b3cc58efa9e2c5b245444788b1d84e02390ab035/src/blocks/homepage-articles/view.php#L63-L77) and [the updated one](https://github.com/Automattic/newspack-blocks/blob/b3cc58efa9e2c5b245444788b1d84e02390ab035/src/blocks/homepage-articles/class-wp-rest-newspack-articles-controller.php#L101-L115) already comes with query parameters.

PR #347, opened 2020-01-24
Hide "More" button on private sites

### LINKED ISSUE #306: Load More button: does not handle private blog
**Describe the bug**
When you activate the Load More button on the block, it doesn't load the posts, it shows the message _"Something went wrong. Please refresh the page and/or try again"_.

**To Reproduce**
1. Go to Pages and create a new page
2. Add the Blog Posts block
3. Activate the option _"Show "More" Button"_ on the block
4. Publish the page

**Screenshots**
When you click on the button to load more posts, this is what shows up, including the errors on the console:

BlogPostsBlock

**Expected behavior**
To load more posts.

WunderBart:
### All Submissions:

Closes #306.

The "More" button request is returning 403 (Forbidden) when the site is set as `Private` in WP.com settings:

image

Because there's no straightforward way of authenticating the request from the view side of the blog (see https://github.com/Automattic/newspack-blocks/pull/347#issuecomment-585883729), we've decided to hide the button in both the editor and view sides.

This is a bit complicated because it depends on a WP.com blog property, so I recommend the following:

Editor testing:

1. Create page
2. In the JS console, run `window.wpcomGutenberg.blogPublic = '-1'`
3. Add Newspack Articles block
4. There should not be the `Show "More" button` option available

View testing:

1. Manually set [`$is_blog_private`](https://github.com/Automattic/newspack-blocks/blob/fix/handle-private-blog-load-more/src/blocks/homepage-articles/view.php#L90) to true
2. Publish & go to a page containing the Articles block with the "More" button enabled
3. There should not be the "More" button on the page

### Other information:


WunderBart:
@obenland @apeatling @jeffersonrabb I haven't tested manually this one yet (pretty sure the implementation is solid, tho), but I'm wondering if this would be the best direction - should we continue with suppressing the button for private sites or rather focus on getting the request authenticated? FYI, I've already tried setting the `X-WP-Nonce` to do the latter but got `Request header field x-wp-nonce is not allowed by Access-Control-Allow-Headers in preflight response`, so I'd appreciate any feedback here 🙏

claudiulodro:
AFAIK the change has been made so the button should be working on WPCOM now. This may not be needed any more?

WunderBart:
I've just tried using "Load More" on my private site and it's still not working. Could you point me to that change please, @claudiulodro?

src/blocks/homepage-articles/edit.js
```diff
@@ -336,6 +337,16 @@ class Edit extends Component {
...
+							help={
+								<>
+									{ __( 'The button will not be displayed on private sites. ' ) }
+									{
+										<ExternalLink href="https://en.support.wordpress.com/settings/#general">
+											{ __( 'More information' ) }
+										</ExternalLink>
+									}
+								</>
+							}
```
obenland: Rather than adding that help text, let's just not even show that option on private sites.

obenland:
This approach would be a bandaid over the bug but it doesn't fix the cause of it. Can we try finding a way to authenticate the API request?

WunderBart:
>This approach would be a bandaid over the bug but it doesn't fix the cause of it. Can we try finding a way to authenticate the API request?

@obenland, I couldn't find an obvious way of authenticating this request (see https://github.com/Automattic/newspack-blocks/pull/347#issuecomment-578224942, https://github.com/Automattic/newspack-blocks/issues/306#issuecomment-583372932 & https://github.com/Automattic/newspack-blocks/issues/306#issuecomment-585874324). The `admin-ajax.php` solution would be worth checking but it's a larger task and for now, I would resolve this edge case by hiding the option for private sites after all. How does that sound? cc @jeffersonrabb @claudiulodro

obenland:
Yeah, let's hide the option in the editor on private sites

src/blocks/homepage-articles/edit.js
WunderBart: > Rather than adding that help text, let's just not even show that option on private sites.

I couldn't find any `isBlogPublic/Private` flag available in the client (editor) and I can't really find a _non-invasive_ way of passing the `blog_public` variable from the back-end either. If we decide that hiding the feature in the editor is necessary, then we could probably consider creating a script tag and assigning the flag globally (the _invasive_ way). What do you think?

> That would also avoid the change in view.php.

Can we actually avoid making a change in `view.php`? If a page already has the block on with `Load More` enabled and then makes the site private, if we disable it in the editor it would still display in the view, wouldn't it?

/cc @jeffersonrabb

jeffersonrabb: > Can we actually avoid making a change in view.php? If a page already has the block on with Load More enabled and then makes the site private, if we disable it in the editor it would still display in the view, wouldn't it?

Yeah, I agree, I think you still need to suppress the feature in `view.php`.

src/blocks/homepage-articles/edit.js
```diff
@@ -5,6 +5,7 @@
  */
 import QueryControls from '../../components/query-controls';
 import { STORE_NAMESPACE } from './store';
+import { getIsBlogPrivate } from './helpers';
```
obenland: Instead of creating a helper function, what do you think of defining a variable in `render()` based on whether the global is available and has the desired value? It would decrease the overhead quite a bit.

src/blocks/homepage-articles/edit.js
```diff
@@ -331,7 +332,7 @@ class Edit extends Component {
 							required
 						/>
 					) }
-					{ ! specificMode && (
+					{ ! specificMode && getIsBlogPrivate() !== true && (
```
obenland: Since it doesn't matter whether it's a boolean or not, these checks can be just implicit

src/blocks/homepage-articles/view.php
```diff
@@ -13,6 +13,8 @@
  * @return string Returns the post content with latest posts added.
  */
 function newspack_blocks_render_block_homepage_articles( $attributes ) {
+	$is_blog_private = defined( 'IS_BLOG_PRIVATE' ) && IS_BLOG_PRIVATE;
```
obenland: Where is this constant being defined? Can we not use the option here?

WunderBart: The [constant will be defined in BPB](https://github.com/Automattic/wp-calypso/pull/39509/commits/22618082d437e13e06f91b6877c1ee9d47593a40), same as [the client-side one](https://github.com/Automattic/wp-calypso/pull/39509/commits/a231e26c1e2d8ba53a7c5a63d00e844373911ac6). I wasn't sure if it's OK to directly use `wpcom` options here, but if it is I think we can also drop the mapping of the client-side one. What do you think?

marekhrabe: I think it is safe to use the direct option instead of the constant. `blog_public` is actually a built-in site option of WordPress so it will be always present.

Documentation for `blog_public` https://developer.wordpress.org/apis/handbook/options/#privacy

The only part that is specific for wpcom is the addition of the third possible value `-1` which means "private".

WunderBart: Ah, that makes a lot of sense. Will use the `blog_public` instead then. Thanks, @marekhrabe!

src/blocks/homepage-articles/edit.js
WunderBart: The `private` flag is needed in 2 different `render`s - IMO it's best to have it in a separate function. I decided to move it to a separate file as this one is already quite heavy.

src/blocks/homepage-articles/edit.js
```diff
@@ -633,8 +641,9 @@ class Edit extends Component {
 						{ latestPosts && latestPosts.map( post => this.renderPost( post ) ) }
 					</div>
 				</div>
-
-				{ ! specificMode && latestPosts && moreButton && (
+				{ ! specificMode && latestPosts && moreButton && ! getIsBlogPrivate() && (
+					// The "More" button option is hidden for private sites, so we should
+					// also hide the button in case it was previously enabled.
```
WunderBart: Decided to go with inline comments because linter kept requiring a `@param` definition here for some reason when I used a block comment.

retrofox: I think you can do it removing one `*` from the first line.

```
/*
 * Use this structure for generic doc block
 */
```

```
/** <- use double for functions/methods/etc
 *
 */
```

src/blocks/homepage-articles/helpers.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Based global WP.com blog_public option, checks whether current blog is
+ * private or not.
+ *
+ * @return {boolean} a private WP.com blog flag
+ */
+export const getIsBlogPrivate = () =>
+	typeof window === 'object' &&
+	window.wpcomGutenberg &&
+	Number( window.wpcomGutenberg.blogPublic ) === -1;
```
retrofox: Are we agree that `Number( window.wpcomGutenberg.blogPublic )` will be `NaN` if blogPublig is not defined?
Maybe it worths check if blogPublic is defined, too:

```es6
export const getIsBlogPrivate = () =>
	typeof window === 'object' &&
	window.wpcomGutenberg &&
	window.wpcomGutenberg.blogPublic &&
	Number( window.wpcomGutenberg.blogPublic ) === -1;
```

src/blocks/homepage-articles/edit.js
WunderBart: TIL, thanks!

src/blocks/homepage-articles/helpers.js
WunderBart: IMO, it's not necessary - `Number()` will not throw error no matter what type of value we put there. `NaN === -1` is still a valid comparison and returns expected value.

retrofox:
Just a simple note: the value of the `blogPublic` property is a string, so in the testing instructions is should be `window.wpcomGutenberg.blogPublic = "-1"`

retrofox:
Just a simple note: the value of the `blogPublic` property is a string, so in the testing instructions is should be `window.wpcomGutenberg.blogPublic = "-1"`

obenland:
@jeffersonrabb Do you think this PR would be worth a new .26 release?

jeffersonrabb:
> @jeffersonrabb Do you think this PR would be worth a new .26 release?

Yeah, although I'm going to do a quick pass through recent issues to see if there's any other low hanging fruit. Either way we'll do a .26 release tomorrow.

PR #353, opened 2020-01-29
CI: semantic-release configuration tweak 

adekbadek:
### All Submissions:

1. [Configures](https://semantic-release.gitbook.io/semantic-release/usage/workflow-configuration#prerelease) `semantic-release` to use `release` branch.
A thorough overview of release/pre-release process can be found [here](https://github.com/semantic-release/semantic-release/blob/master/docs/recipes/pre-releases.md).
1. Updates the version of `semantic-release-version-bump`, so that
`NEWSPACK_BLOCKS__VERSION` will be updated on release, as done manually here: 3add1f003eac3822c576b2e6d132ad13e70e13ec
1. Change CI workflow name to `all` - `quality_assurance` was big ambiguous as it also handles releasing

### Other information:


package.json
```diff
@@ -77,6 +77,13 @@
     "*.scss": "npm run lint:scss:staged"
   },
   "release": {
+    "branches": [
+      "release",
+      {
+        "name": "pre-release",
+        "prerelease": true
```
claudiulodro: Sorry about all the weird requirements that weren't made super clear up front, but we actually need this to publish a regular non-prerelease release but with version `1.0.0-alpha.x`. This is because prereleases get ignored when determining the download for e.g. https://github.com/Automattic/newspack-blocks/releases/latest/download/newspack-blocks.zip and therefore the installers won't work correctly.

adekbadek: That's true - so what we need are regular releases with pre-release version numbers, which AFAIK is not configurable with semantic-release. 
Therefore we have to wait until we can do regular releases :( I've updated this PR so it just configures releases to be run from `release` branch, and we just won't use this branch until March

PR #371, opened 2020-02-15
Block Patterns

jeffersonrabb:
### All Submissions:

* [ ] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-theme/blob/master/.github/CONTRIBUTING.md)?
* [ ] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [ ] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Creates a framework for defining block patterns which can be inserted into posts.

Screen Shot 2020-02-15 at 11 32 27 AM

The patterns themselves are defined in `newspack-block-patterns.php`. By using a filter, other plugins and themes can add their own. It's also possible to add patterns only to certain post types. In the examples here the three are added to Posts and Pages.

### How to test the changes in this Pull Request:

1.
2.
3.

### Other information:

* [ ] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #377, opened 2020-02-20
Block Patterns Infrastructure

jeffersonrabb:
### All Submissions:

Introduces UI and infrastructure for "Block Patterns:" preset groupings of blocks that can be inserted into posts. Similar to https://github.com/Automattic/newspack-blocks/pull/371, but this PR contains no actual patterns, only the infrastructure. 

Block patterns are defined by using the `newspack_blocks_patterns` filter. In the editor, if there are any patterns defined, there will be an icon to the right of the Settings icons. Clicking reveals the Newspack Patterns sidebar, which contains all defined patterns. Clicking the pattern inserts the blocks into the editor.

Example implementation:

Updated: `icon` parameter renamed to `preview_image`.
```
add_filter(
	'newspack_blocks_patterns',
	function( $patterns, $post_type ) {
		if ( in_array( $post_type, [ 'post', 'page' ], true ) ) {
			$patterns[] = [
				'category' => __( 'Homepage Posts', 'newspack-blocks' ),
				'content'  => '<!-- wp:heading --><h2>Hello World</h2><!-- /wp:heading -->',
				'preview_image'     => '/path/to/icon/image',
				'title'    => __( 'Test Pattern', 'newspack-blocks' ),
			];
		}
		return $patterns;
	},
	10,
	2
);
```

The `$post_type` parameter can be used to register patterns only for that post type. `content` can be any valid block markup. To test, simply create some blocks, switch to Code Editor, and copy paste. Patterns will be grouped by Category. Categories will be listed alphabetically. `icon` is a path to an icon to use for the pattern.

1. Initially verify that the patterns icon is NOT shown in the editor. It should only be shown if there are registered patterns available.
2. Register one or more patterns using code like the example above. For real world examples, try the ones in https://github.com/Automattic/newspack-blocks/pull/371. Icons are available in that PR as well.
3. Return to the Editor. Verify the icon is present. Click the icon, verify the UI. Click on any of the patterns and verify the blocks are inserted into the post.

### Other information:


class-newspack-blocks-api.php
```diff
@@ -10,6 +10,14 @@
  */
 class Newspack_Blocks_API {
 
+	/**
+	 * Namespace for REST routes
+	 *
+	 * @var string
+	 */
+
+	public static $namespace = 'newspack-blocks/v1';
```
adekbadek: Question: where is this used? Also, there's an extra newline on line 18.

src/patterns/index.js
```diff
@@ -0,0 +1,107 @@
...
+				) }
+			</PluginSidebar>
+		);
+	}
+}
+
+const PatternsSidebarWithDispatch = compose( [
+	withSelect( select => {
+		const { getCurrentPostType } = select( 'core/editor' );
+		return { currentPostType: getCurrentPostType() };
```
adekbadek: `currentPostType` prop is not used - why is it here?

src/patterns/index.js
```diff
@@ -0,0 +1,107 @@
...
+												} }
+												role="button"
+												tabIndex="0"
+												aria-label={ pattern.title }
+											>
+												<div className="editor-block-styles__item-preview block-editor-block-styles__item-preview">
+													{ pattern.icon && (
+														<img src={ pattern.icon } alt={ __( 'Preview', 'newspack-block' ) } />
+													) }
+												</div>
```
adekbadek: name `icon` is a bit misleading – I'd think an icon name could be supplied here. How about `previewImage`?

src/patterns/style.scss
```diff
@@ -0,0 +1,14 @@
...
+ */
+
+.newspack-patterns-block-styles {
+	.block-editor-block-styles__item {
+		padding-top: 6px;
+	}
+
+	.block-editor-block-styles__item-preview {
+		margin: 0;
+		padding: 0;
```
adekbadek:
```suggestion
		padding: 0;
		width: 100%;
```

to prevent this:

image

src/patterns/index.js
```diff
@@ -0,0 +1,107 @@
...
+														event.preventDefault();
+														insertBlocks( parse( pattern.content ) );
+													}
+												} }
+												role="button"
+												tabIndex="0"
+												aria-label={ pattern.title }
+											>
+												<div className="editor-block-styles__item-preview block-editor-block-styles__item-preview">
+													{ pattern.icon && (
```
adekbadek: Maybe a generic pattern image could be supplied here (or at least a grey rectangle)?

class-newspack-blocks-api.php
claudiulodro: I think [this bit](https://github.com/Automattic/newspack-blocks/pull/371/files#diff-290d886f31daef68ffba027bd7cc99e8R26-R44) is missing from this PR?

claudiulodro: Or maybe that was supposed to be removed, since patterns are now in a global variable?

src/patterns/index.js
```diff
@@ -0,0 +1,107 @@
+/**
+ * WordPress dependencies.
+ */
+import apiFetch from '@wordpress/api-fetch';
```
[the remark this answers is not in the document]
jeffersonrabb: Fixed in https://github.com/Automattic/newspack-blocks/pull/377/commits/0a56f1a72620ff07d1317cc0ec6143c34e980784. Any idea why this didn't pop up when I committed locally but did in Circle CI?

class-newspack-blocks-api.php
jeffersonrabb: In the older PR patterns were retrieved by API but in this one I switched to a global variable. The primary reason for this change is to hide the patterns icon if there are no patterns registered. A secondary advantage is that the patterns UI is faster since they are immediately available without requiring an API response. Any reason to be concerned with this approach? The unused `$namespace` var is removed in https://github.com/Automattic/newspack-blocks/pull/377/commits/0a56f1a72620ff07d1317cc0ec6143c34e980784.

src/patterns/index.js
jeffersonrabb: Good catch. Removed in https://github.com/Automattic/newspack-blocks/pull/377/commits/0a56f1a72620ff07d1317cc0ec6143c34e980784, along with `withSelect` and `compose`.

src/patterns/index.js
jeffersonrabb: Good idea. Changed to `preview_image` in https://github.com/Automattic/newspack-blocks/pull/377/commits/0f9faa93427f272d6f495a6bd2fd2c093fa5992c

src/patterns/index.js
jeffersonrabb: @thomasguillot suggested simply suppressing any registered patterns that have no preview image, which is done in https://github.com/Automattic/newspack-blocks/pull/377/commits/0f9faa93427f272d6f495a6bd2fd2c093fa5992c.

jeffersonrabb:
> When a block is inserted it seems to get sanitized, which breaks the block:

Where did this block content come from?

jeffersonrabb:
> One UX note: I unpinned the icon from the top-right of the screen (Click on it and select "Unpin from toolbar"), and now it seems to be permanently gone and I don't know how to get it back. :)

Good call. In https://github.com/Automattic/newspack-blocks/pull/377/commits/6e02b6d006861e2d6ffd6bdddd0f9dec33b1e96e I've added it to the Plugin Sidebar More Menu. Click the three vertical dots, then you should see "Newspack Patterns" in the Plugins section of the menu. Clicking on this takes you back to the menu, and from there you can star it again to re-add to the top menu.

class-newspack-blocks-api.php
claudiulodro: No I think it's a good approach 👍

PR #378, opened 2020-02-20
Add a script to generate translations

akirk:
### All Submissions:

This standardizes the process to generate all files needed for a translated Gutenberg UI, improving on #307.

In #307 not all necessary JSON files were generated because the POT file didn't include references to the webpack generated files. This introduces a `npm run i18n` which
1. compiles the blocks to production ready versions, then
2. extracts the strings using `wp i18n make-pot`, then
3. updates the translations from previous PO files to maintain them (using `msgmerge`), and finally
4. generates the JSON files using `wp i18n make-json`.

1. Switch your wp-admin UI to Spanish or German
2. Add a new Homepage Posts block
3. Observe translations in the sidebar, for example "Authors", this should be translated.

### Other information:


akirk:
Correct. It is a best practice though since the source code positions for translatable texts show  generated and plain text file positions. Granted, this is a small project so not as relevant, but generally this makes it easier to find the code position where a text appears, and thus allow to gather more context for a better translation. 

The script is also future proof as it will pick up any new blocks as well.

Issue #436, opened 2020-04-23
Article Carousel: Can Tab through slides without using the interactive buttons


jeryj:
**Describe the bug**
When you're navigating the slides via keyboard, you can tab through the slides without using the buttons. 

Based on the `<amp-carousel>` examples, I _think_ it's possible to have this interaction work: https://amp.dev/documentation/examples/multimedia-animations/image_galleries_with_amp-carousel/?format=websites. Some of them work to bring you to the buttons instead of tabbing through the images, a some work like the screenshot I uploaded.

**To Reproduce**
1. Create an Article Carousel
2. View it on the Front-end site
3. Tab to the Article Carousel
4. Keep on Tabbing :)
5. See the slides change before you reach the buttons.

**Screenshots**
![article-carousel-tabbing](https://user-images.githubusercontent.com/967608/80148419-a3de9880-857a-11ea-8769-610240f40397.gif)

**Expected behavior**
When tabbing through a carousel, it should go from the content in the current slide and then move to the "next" carousel control button.

jeffersonrabb:
When the Jetpack Slideshow block was in testing, we did an accessibility review here: `p1HpG7-6jw-p2`. If I'm understanding correctly, the item titled `Using it with a screen reader, I can’t actually move into the individual slides.` is addressing the same question as this issue, but describes the opposite behavior as expected.

jeryj:
Thanks for looking into this, @jeffersonrabb! 🙌 It's awesome that an a11y review was done for this block. Major kudos there!

> Using it with a screen reader, I can’t actually move into the individual slides.

You should definitely still be able to tab into a slide, but only the current slide should be in the focus order. The rest should not be able to receive focus until they're the current slide.

Here's the current focus order:
- slide one post title
- slide one post author
- slide two post title
- slide two post author
- etc for the rest of the slides
- Previous button
- Next button
- Navigation buttons

I could see arguments for a couple different ways and I'm torn on what the best focus order would be. I _think_ this is the one I'd recommend though if I had to choose.

- Previous Button
- Current Slide Title
- Current Slide author
- Next Button
- Navigation buttons

I like that once you're on a slide, to continue to the next slide you don't have to tab through the previous button as well. However, when you first enter the slider, you'd see the Previous button on content you haven't even encountered. But, focusing the Previous button first would follow the visual focus order, which is a point in its favor.

[...]

PR #469, opened 2020-05-07
Articles Carousel: Improve accessibility

jeryj:
Improves Articles Carousel accessibility through keyboard focus management and screen reader interactions.

### All Submissions:

* [x] Add the `@wordpress/a11y` package to use `speak()`
* [x] Add the `@wordpress/escape-html` package to escape slide change announcement
* [x] Uses `<button>` elements instead of `<a role="button">` elements
* [x] Removes inactive slide links from the focus order, so you can't tab to a new slide (must use the arrow or pagination bullets)
* [x] Announce new slide content and "slide _ of _" (if not autoplaying) 
* [x] Sets inactive slides to `aria-hidden`
* [x] Announce Paused/Playing state after button press
* [x] Fix focus loss when pressing Pause / Play button
* [x] Moves Pause/Play buttons to _before_ the slider so it can be stopped sooner by assistive technologies like screen readers (AT)
* [x] Remove "This is the last slide" announcement (and all other not-so-helpful Swiper 
 announcements) that prevents the slide content from being announced.
Closes https://github.com/Automattic/newspack-blocks/issues/436
* [x] Unifies `createSwiper` implementation across frontend and editor (Thanks @WunderBart!)

**Editor with VoiceOver on Safari**
ezgif com-optimize

**Frontend with VoiceOver on Safari**
ezgif com-optimize (1)

1. Add an Articles Carousel to a post
2. Tab through the slider
3. It should not switch slides when tabbing
4. All Arrows and Navigation dots should be able to be used with the keyboard
5. Check that the current slide's title and author can be tabbed to
6. Pause/Play buttons should work and not lose focus when activating the button with a keyboard
7. Test the above on Editor side as well (you can move focus into the Editor Carousel by clicking initially into the carousel, then tabbing around)

### Other information:

### DIFF
--- a/src/blocks/carousel/create-swiper.js
+++ b/src/blocks/carousel/create-swiper.js
@@ -14,13 +14,16 @@ import 'swiper/dist/css/swiper.css';
  * @param {HTMLElement} slide Slide DOM element
  */
 export function activateSlide( slide ) {
-	slide.ariaHidden = 'false';
+	slide.setAttribute( 'aria-hidden', 'false' );
+
 	/**
 	 * Calls Array.prototype.forEach for IE11 compatibility.
 	 *
 	 * @see https://developer.mozilla.org/en-US/docs/Web/API/NodeList
 	 */
-	Array.prototype.forEach.call( slide.querySelectorAll( 'a' ), e => ( e.tabIndex = '0' ) );
+	Array.prototype.forEach.call( slide.querySelectorAll( 'a' ), el =>
+		el.removeAttribute( 'tabindex' )
+	);
 }
 
 /**
@@ -29,13 +32,15 @@ export function activateSlide( slide ) {
  * @param {HTMLElement} slide Slide DOM element
  */
 export function deactivateSlide( slide ) {
-	slide.ariaHidden = 'true';
+	slide.setAttribute( 'aria-hidden', 'true' );
 	/**
 	 * Calls Array.prototype.forEach for IE11 compatibility.
 	 *
 	 * @see https://developer.mozilla.org/en-US/docs/Web/API/NodeList
 	 */
-	Array.prototype.forEach.call( slide.querySelectorAll( 'a' ), e => ( e.tabIndex = '-1' ) );
+	Array.prototype.forEach.call( slide.querySelectorAll( 'a' ), el =>
+		el.setAttribute( 'tabindex', '-1' )
+	);
 }
 
 /**
--- a/src/blocks/carousel/create-swiper.js
+++ b/src/blocks/carousel/create-swiper.js
@@ -2,15 +2,26 @@
  * External dependencies
  */
 import { merge } from 'lodash';
+import { speak } from '@wordpress/a11y';
 import { __, sprintf } from '@wordpress/i18n';
 import Swiper from 'swiper';
 import 'swiper/dist/css/swiper.css';
+
+export const activateSlide = slide => {
+	slide.ariaHidden = 'false';
+	slide.querySelectorAll( 'a' ).forEach( e => ( e.tabIndex = '0' ) );
+};
+
+export const deactivateSlide = slide => {
+	slide.ariaHidden = 'true';
+	slide.querySelectorAll( 'a' ).forEach( e => ( e.tabIndex = '-1' ) );
+};
 
 /**
  * @param {string} container Selector
  * @param {Object} params Params passed to Swiper
  */
-export default function createSwiper( container = '.swiper-container', params = {} ) {
+export const createSwiper = ( container = '.swiper-container', params = {} ) => {
 	const defaultParams = {
 		/**
 		 * Remove the messages, as we're announcing the slide content and number. These messages are overwriting the slide announcement.
@@ -42,7 +53,36 @@ export default function createSwiper( container = '.swiper-container', params = 
 		releaseFormElements: false,
 		setWrapperSize: true,
 		touchStartPreventDefault: false,
+		on: {
+			init() {
+				this.wrapperEl
+					.querySelectorAll( '.swiper-slide' )
+					.forEach( slide => deactivateSlide( slide ) );
+
+				// Set-up our active slide.
+				activateSlide( this.slides[ this.activeIndex ] );
+			},
+			slideChange() {
+				deactivateSlide( this.slides[ this.previousIndex ] );
+
+				activateSlide( this.slides[ this.activeIndex ] );
+
+				// If we're autoplaying, don't announce the slide change, as that would be supremely annoying.
+				if ( ! this.autoplay.running ) {
+					// Announce the contents of the slide.
+					speak(
+						`${ this.slides[ this.activeIndex ].innerHTML }, ${ sprintf(
+							/* translators: current slide number and the total number of slides */
+							__( 'Slide %s of %s', 'newspack-blocks' ),
+							this.realIndex + 1,
+							this.pagination.bullets.length
+						) }`,
+						'assertive'
+					);
+				}
+			},
+		},
 	};
 
 	return new Swiper( container, merge( {}, defaultParams, params ) );
-}
+};
--- a/src/blocks/carousel/create-swiper.js
+++ b/src/blocks/carousel/create-swiper.js
@@ -44,7 +44,8 @@ export function deactivateSlide( slide ) {
 
 /**
  * @param {Object} els Swiper elements
- * @param {Element} els.container Container element
+ * @param {Element} els.block Block element
+ * @param {Element} els.container Swiper container element
  * @param {Element} els.next Next button element
  * @param {Element} els.prev Previous button element
  * @param {Element} els.play Play button element
@@ -98,7 +99,7 @@ export function createSwiper( els, config = {} ) {
 						}
 
 						this.autoplay.stop();
-						els.container.classList.remove( 'wp-block-newspack-blocks-carousel__autoplay-playing' );
+						els.block.classList.remove( 'wp-block-newspack-blocks-carousel__autoplay-playing' );
 						speak( __( 'Paused', 'newspack-blocks' ), 'assertive' );
 						// Move focus to the play button.
 						els.play.focus();
@@ -111,7 +112,7 @@ export function createSwiper( els, config = {} ) {
 						}
 
 						this.autoplay.start();
-						els.container.classList.add( 'wp-block-newspack-blocks-carousel__autoplay-playing' );
+						els.block.classList.add( 'wp-block-newspack-blocks-carousel__autoplay-playing' );
 						speak( __( 'Playing', 'newspack-blocks' ), 'assertive' );
 						// Move focus to the pause button.
 						els.pause.focus();
--- a/src/blocks/carousel/create-swiper.js
+++ b/src/blocks/carousel/create-swiper.js
@@ -7,21 +7,41 @@ import { __, sprintf } from '@wordpress/i18n';
 import Swiper from 'swiper';
 import 'swiper/dist/css/swiper.css';
 
-export const activateSlide = slide => {
+/**
+ * Modifies attributes on slide HTML to make it accessible.
+ *
+ * @param {HTMLElement} slide Slide DOM element
+ */
+export function activateSlide( slide ) {
 	slide.ariaHidden = 'false';
-	slide.querySelectorAll( 'a' ).forEach( e => ( e.tabIndex = '0' ) );
-};
+	/**
+	 * Calls Array.prototype.forEach for IE11 compatibility.
+	 *
+	 * @see https://developer.mozilla.org/en-US/docs/Web/API/NodeList
+	 */
+	Array.prototype.forEach.call( slide.querySelectorAll( 'a' ), e => ( e.tabIndex = '0' ) );
+}
 
-export const deactivateSlide = slide => {
+/**
+ * Modifies attributes on slide HTML to make it accessible.
+ *
+ * @param {HTMLElement} slide Slide DOM element
+ */
+export function deactivateSlide( slide ) {
 	slide.ariaHidden = 'true';
-	slide.querySelectorAll( 'a' ).forEach( e => ( e.tabIndex = '-1' ) );
-};
+	/**
+	 * Calls Array.prototype.forEach for IE11 compatibility.
+	 *
+	 * @see https://developer.mozilla.org/en-US/docs/Web/API/NodeList
+	 */
+	Array.prototype.forEach.call( slide.querySelectorAll( 'a' ), e => ( e.tabIndex = '-1' ) );
+}
 
 /**
  * @param {string} container Selector
  * @param {Object} params Params passed to Swiper
  */
-export const createSwiper = ( container = '.swiper-container', params = {} ) => {
+export function createSwiper( container = '.swiper-container', params = {} ) {
 	const defaultParams = {
 		/**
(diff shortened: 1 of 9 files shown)


src/blocks/carousel/create-swiper.js
```diff
@@ -24,12 +41,48 @@ export default function createSwiper( container = '.swiper-container', params =
...
+					index + 1
+				) }</span></button>`;
+			},
 		},
 		preventClicksPropagation: false /* Necessary for normal block interactions */,
 		releaseFormElements: false,
 		setWrapperSize: true,
 		touchStartPreventDefault: false,
+		on: {
+			init() {
```
jeffersonrabb, line 57: Looks like you can't tab into blocks in the editor? If this is true, would it make sense to move all of the a11y wrangling into `view.js` to simplify? There's already some duplicated code due to the competing `init` callbacks.

src/blocks/carousel/create-swiper.js
```diff
@@ -2,15 +2,32 @@
...
  */
 import { merge } from 'lodash';
+import { speak } from '@wordpress/a11y';
+import { __, sprintf } from '@wordpress/i18n';
 import Swiper from 'swiper';
 import 'swiper/dist/css/swiper.css';
 
+export const activateSlide = slide => {
+	slide.ariaHidden = 'false';
+	slide.querySelectorAll( 'a' ).forEach( e => ( e.tabIndex = '0' ) );
```
jeffersonrabb, line 12: This will be a problem in some older browsers (IE11) because `forEach` isn't universally supported in `NodeList` (https://caniuse.com/#feat=mdn-api_nodelist_foreach). There are already some other issues breaking the Carousel in IE11, but assuming those can be resolved it would be good to fix these too.

src/blocks/carousel/view.js
```diff
@@ -44,20 +47,34 @@ if ( typeof window !== 'undefined' ) {
...
 								block.classList.add( 'wp-block-newspack-blocks-carousel__autoplay-playing' );
+								speak( __( 'Playing', 'newspack-blocks' ), 'assertive' );
+								// Move focus to the pause button.
+								els.pause.focus();
 							} );
 						}
+
+						this.wrapperEl
+							.querySelectorAll( '.swiper-slide' )
+							.forEach( slide => deactivateSlide( slide ) );
```
jeffersonrabb, line 74: Same note about `NodeList`: `forEach`

src/blocks/carousel/create-swiper.js
WunderBart: Just for reference, this is how we recently resolved it in the Homepage Articles block: https://github.com/Automattic/newspack-blocks/commit/ec20af576874f5059a06d8979e49900de0458d35

src/blocks/carousel/create-swiper.js
```diff
@@ -24,12 +41,48 @@ export default function createSwiper( container = '.swiper-container', params =
...
+			slideChange() {
+				deactivateSlide( this.slides[ this.previousIndex ] );
+
+				activateSlide( this.slides[ this.activeIndex ] );
+
+				// If we're autoplaying, don't announce the slide change, as that would be supremely annoying.
+				if ( ! this.autoplay.running ) {
+					// Announce the contents of the slide.
+					speak(
+						`${ this.slides[ this.activeIndex ].innerHTML }, ${ sprintf(
```
WunderBart, line 74: If this is something to be spoken, should we use `innerText` instead of `innerHTML` to avoid announcing markup?

src/blocks/carousel/create-swiper.js
```diff
@@ -2,15 +2,32 @@
...
+	slide.ariaHidden = 'true';
+	slide.querySelectorAll( 'a' ).forEach( e => ( e.tabIndex = '-1' ) );
+};
+
 /**
  * @param {string} container Selector
  * @param {Object} params Params passed to Swiper
  */
-export default function createSwiper( container = '.swiper-container', params = {} ) {
+export const createSwiper = ( container = '.swiper-container', params = {} ) => {
```
retrofox, line 24: minor one: It's possible don't use arrow function, just in case. 

```suggestion
function createSwiper( container = '.swiper-container', params = {} ) {
```

jeryj: 👍 To make sure I understand, is it just for older browser compatibility?

src/blocks/carousel/create-swiper.js
jeryj: I was using innerHTML in order to have the image alt text announced as well, but I'm not sure if the alt text on images is even being used. May need to build that out manually, like `${ esc_html(img.alt)} ${ this.slides[ this.activeIndex ].innerText }`

src/blocks/carousel/create-swiper.js
jeryj: You _should_ be able to tab into the blocks in the editor. If the intention is for the carousel in the editor to not be functional, then we can remove these things from the editor side. However, if it's supposed to be functional on the editor side, then we should allow keyboard controls as well in order to maintain a similar experience for assistive technology users.

jeffersonrabb: > You should be able to tab into the blocks in the editor. 

Are you able to in your testing? I may have been doing it wrong, but in the editor I was only able to tab through the top-level blocks, never into the internals (or even `InnerBlocks`). This seems to be the case for any blocks, not just the Carousel.

jeryj: Screen Shot 2020-05-11 at 9 59 52 AM

Once you tab to it and see this state, press `Enter` to enter into the block and then you can access the slider using `tab` a few more times.

src/blocks/carousel/create-swiper.js
jeryj: Updated to use `Array.prototype.forEach.call`

src/blocks/carousel/create-swiper.js
jeryj: Updated this to use:
- `import { escapeHTML } from '@wordpress/escape-html';`
- Grab `alt` text if present on the image and add it to the spoken string
- Use slide innerText

jeryj:
I'm going to put a bow on this PR since it's grown so much 😬 

Things to follow-up on:
- AMP accessibility
- Adding alt text to editor slider

src/blocks/carousel/create-swiper.js
```diff
@@ -1,35 +1,166 @@
...
+						els.pause.focus();
+					} );
+				}
+
+				/**
+				 * Calls Array.prototype.forEach for IE11 compatibility.
+				 *
+				 * @see https://developer.mozilla.org/en-US/docs/Web/API/NodeList
+				 */
+				Array.prototype.forEach.call( this.wrapperEl.querySelectorAll( '.swiper-slide' ), slide =>
```
WunderBart, line 127: Can we make a helper from this to make things a bit DRYer? I.e.:

```js
function forEachNode( nodeList, cb ) {
  /**
   * Calls Array.prototype.forEach for IE11 compatibility.
   *
   * @see https://developer.mozilla.org/en-US/docs/Web/API/NodeList
   */
  Array.prototype.forEach.call( nodeList, cb );
}
```

src/blocks/carousel/view.scss
```diff
@@ -181,7 +185,7 @@
 		right: 1.5em;
 		top: 1.5em;
 		transform: none;
-		z-index: 1;
+		z-index: 9;
```
WunderBart, line 188: Why does this need to be a `9`?

jeryj: To show-up on top of the carousel. Without it, it's hidden behind it. It could be `2` though, if you'd prefer to lower it :)

src/blocks/carousel/create-swiper.js
```diff
@@ -1,35 +1,166 @@
...
+ * @param {Element} els.block Block element
+ * @param {Element} els.container Swiper container element
+ * @param {Element} els.next Next button element
+ * @param {Element} els.prev Previous button element
+ * @param {Element} els.play Play button element
+ * @param {Element} els.pause Pause button element
+ * @param {Element} els.pagination Pagination element
+ * @param {Object} config Swiper config
+ */
+export function createSwiper( els, config = {} ) {
```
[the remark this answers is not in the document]
jeryj: It was originally the `default`, but I switched it up when I added the others to the file. But now the others aren't even exported, so it makes a lot of sense to switch back to `default`.

jeryj:
Thanks, @WunderBart! All those changes look like good ideas to me 👍

WunderBart:
@jeffersonrabb @jeryj I noticed some additional issues while doing some heavy manual testing in the editor. I decided to push the fixes here (https://github.com/Automattic/newspack-blocks/pull/469/commits/8eae15378867942617c4b350c4423af6e0ef01fa) - could you give it another look before we merge?

src/blocks/carousel/create-swiper.js
```diff
@@ -1,35 +1,172 @@
...
 		releaseFormElements: false,
 		setWrapperSize: true,
 		touchStartPreventDefault: false,
-	};
+		on: {
+			init() {
+				if ( config.autoplay ) {
+					els.pause.addEventListener( 'click', () => {
+						if ( this.destroyed ) {
+							return;
```
jeryj, line 104: Rather than just returning, shouldn't we actually remove the event listener entirely? As in, when the slider is destroyed, remove the event listeners for these buttons.

src/blocks/carousel/edit.js
```diff
@@ -32,57 +25,64 @@ import { withSelect } from '@wordpress/data';
...
+			prevProps.attributes.autoplay !== autoplay ||
+			prevProps.attributes.delay !== delay
+		) {
+			let initialSlide = 0;
+
+			if ( this.swiperInstance ) {
+				if ( latestPosts && this.swiperInstance.realIndex < latestPosts.length ) {
+					initialSlide = this.swiperInstance.realIndex;
+				}
+				this.swiperInstance.destroy( true, true );
```
jeryj, line 62: Maybe this would be a good spot to remove the play / pause listeners?

WunderBart: > Why are we checking to set the `initialSlide` and then destroying it? Is it to attempt to preserve the current position in the slider when updating to new settings?

Exactly! 😄

WunderBart:
> Why are we checking to set the `initialSlide` and then destroying it? Is it to attempt to preserve the current position in the slider when updating to new settings?

Exactly! 😄

src/blocks/carousel/create-swiper.js
WunderBart: Implemented in https://github.com/Automattic/newspack-blocks/pull/469/commits/99464880ad8f27907dbb4d1bfa15944b73e57f72.

src/blocks/carousel/edit.js
WunderBart: Good spot, done in https://github.com/Automattic/newspack-blocks/pull/469/commits/99464880ad8f27907dbb4d1bfa15944b73e57f72!

src/blocks/carousel/create-swiper.js
```diff
@@ -1,35 +1,204 @@
...
+
+		swiper.on( 'autoplayStop', function() {
+			els.block.classList.remove( autoplayClassName ); // Hide pause & show play button.
+			speak( __( 'Paused', 'newspack-blocks' ), 'assertive' );
+		} );
+
+		swiper.on( 'beforeDestroy', function() {
+			els.play.removeEventListener( 'click', handlePlayButtonClick );
+			els.pause.removeEventListener( 'click', handlePauseButtonClick );
+		} );
```
jeryj, line 198: I think we should remove these handle functions, and just do:

```
   swiper.on( 'init', function() {
	els.play.addEventListener( 'click', this.autoplay.start );
	els.pause.addEventListener( 'click', this.autoplay.stop );
} );

swiper.on( 'autoplayStart', function() {
			els.block.classList.add( autoplayClassName ); // Hide play & show pause button.
			speak( __( 'Playing', 'newspack-blocks' ), 'assertive' );
            els.pause.focus(); // Move focus to the pause button.
		} );

		swiper.on( 'autoplayStop', function() {
			els.block.classList.remove( autoplayClassName ); // Hide pause & show play button.
			speak( __( 'Paused', 'newspack-blocks' ), 'assertive' );
            els.play.focus(); // Move focus to the play button.
		} );

       swiper.on( 'autoplayStart', function() {
			els.block.classList.add( autoplayClassName ); // Hide play & show pause button.
			speak( __( 'Playing', 'newspack-blocks' ), 'assertive' );
		} );

		swiper.on( 'autoplayStop', function() {
			els.block.classList.remove( autoplayClassName ); // Hide pause & show play button.
			speak( __( 'Paused', 'newspack-blocks' ), 'assertive' );
		} );

        swiper.on( 'beforeDestroy', function() {
			els.play.removeEventListener( 'click', handlePlayButtonClick );
			els.pause.removeEventListener( 'click', handlePauseButtonClick );
		} );
```

WunderBart: We cannot do that, unfortunately. The reason is the focus, that should be switched *only* when the button is clicked. Otherwise, when you change the block settings, the `autoplayStart` event would _steal_ your current focus, wherever it is, and move it to the play button. Does that make sense?

jeryj: Yes. Thanks for explaining it!

src/blocks/carousel/edit.js
```diff
@@ -62,22 +62,24 @@ class Edit extends Component {
...
-					pause: this.btnPauseRef.current,
-					pagination: this.paginationRef.current,
-				},
-				{
-					autoplay,
-					delay: delay * 1000,
-					initialSlide,
-				}
-			);
+			if ( latestPosts && latestPosts.length > 0 ) {
```
WunderBart, line 65: Can we pair this condition with [these](https://github.com/Automattic/newspack-blocks/pull/469/commits/da72cecadca5058367953a2c66310327fe0d80a5#diff-969286d51bb0afe10983b9aec317b5eeR52-R54)?

jeryj: Hmm... I'm not sure, as we still want to run the the destroy swiper stuff if we don't have any posts. Maybe I'm misunderstanding what you're proposing though.

jeryj:
@jeffersonrabb This is ready for review again :) We've done a few performance and resiliency improvements:
- Only add the autoplay handlers if the autoplay config is on
- Remove the autoplay handlers when the slider is destroyed, if the autoplay config is on (using Swiper `beforeDestroy` Event)
- Reinitialize the slider in the editor if:
  - Attributes have changed
  - Post length has changed (attribute changes should detect if the query has changed)
- Only attempt to initialize the slider if we have posts

jeryj:
Just tested this and I see what you mean. It looks like the new component is sharing the instance of the other swiper somehow. When you add the second swiper, the `componentDidUpdate` already has an instance of `this.swiperInstance` instead of it being `null`.

It looks like this is an issue on `master` as well though. Do you mind if we get this merged and then fix this in a subsequent PR?

jeffersonrabb:
> Just tested this and I see what you mean. It looks like the new component is sharing the instance of the other block somehow. When you add the second swiper, the `componentDidUpdate` already has an instance of `this.swiperInstance` instead of it being `null`.
> 
> It looks like this is an issue on `master` as well though. Do you mind if we get this merged and then fix this in a subsequent PR?

Yeah, we probably need to maintain `swiperInstance` in state instead. I'm fine looking at that in a separate PR, and this is looking good.

PR #473, opened 2020-05-11
Homepage Posts: Add image shape classes to the front end

### LINKED ISSUE #472: Homepage Posts: Add or move image size classes
Right now the homepage posts block will output the image's shape (landscape, portrait, etc), as a class on the image itself.

If this were moved -- ideally to the `.post-thumbnail` element -- it would increase the flexibility of what could be done with it. Specifically, we'd be able to change the aspect ratio of the images on a site by site basis, by using CSS like:

```
.is-landscape .post-thumbnail a {
	display: inline-block;
	padding-top: 56.25%;
	position: relative;
	overflow: hidden;
	width: 100%;
}
.is-landscape .post-thumbnail img {
	position: absolute;
	left: 50%;
	top: 50%;                        
  transform: translate(-50%, -50%)
}
```

laurelfulford:
### All Submissions:

This PR makes sure the 'image shape' classes are actually being output on the front-end of the block (right now they only work in the editor). It also simplifies the classes a bit, from image-shape[shapename] to just is-[shapename], to shorten them, and make them match other class formats.

These classes can be used to change the image aspect ratio a bit with CSS only -- so if someone wanted their landscape images to be 16:9 rather than 4:3 on the front end, they could use CSS like the example in #472 to adjust them.

Closes #472.

2. Add four homepage posts blocks to the editor; one with the default image shape, one with portrait, once with square and one with no crop.
3. Inspect each and confirm that a class for each shape is output on the main block element, with the `wpnbha` class.
4. Publish your page. 
5. Confirm that those classes are also showing up on the homepage.

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/edit.js
+++ b/src/blocks/homepage-articles/edit.js
@@ -508,7 +508,7 @@ class Edit extends Component {
 			[ `image-align${ mediaPosition }` ]: showImage,
 			[ `is-${ imageScale }` ]: imageScale !== '1' && showImage,
 			'mobile-stack': mobileStack,
-			[ `image-shape${ imageShape }` ]: imageShape !== 'landscape',
+			[ `is-${ imageShape }` ]: showImage,
 			'has-text-color': textColor.color !== '',
 			'show-caption': showCaption,
 			'show-category': showCategory,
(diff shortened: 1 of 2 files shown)


laurelfulford:
Thanks!

PR #476, opened 2020-05-14
Articles Carousel: Fix Swiper init for additional blocks

WunderBart:
Solves issue described in https://github.com/Automattic/newspack-blocks/pull/469#pullrequestreview-411362555.

Swiper wasn't initializing for additional blocks because `componentDidUpdate` (where we initialize Swiper) is not called for them. The reason that only the first block is calling `componentDidUpdate` is that the `latestPosts` prop is empty on mount and changes once the request is done, triggering the update. That is not the case for additional blocks because the `latestPosts` data is already there. Initializing `Swiper` on `componentDidMount` (if posts are available) fixes the issue.

### All Submissions:

### Other information:

### DIFF
--- a/src/blocks/carousel/edit.js
+++ b/src/blocks/carousel/edit.js
@@ -44,9 +44,12 @@ class Edit extends Component {
 		this.paginationRef = createRef();
 	}
 
+	componentDidMount() {
+		this.initializeSwiper( 0 );
+	}
+
 	componentDidUpdate( prevProps ) {
 		const { attributes, latestPosts } = this.props;
-		const { autoplay, delay } = attributes;
 
 		if (
 			prevProps.latestPosts !== latestPosts ||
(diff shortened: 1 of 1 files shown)


src/blocks/carousel/edit.js
```diff
@@ -64,24 +67,32 @@ class Edit extends Component {
...
-				);
-			}
+			this.initializeSwiper( initialSlide );
+		}
+	}
+
+	initializeSwiper( initialSlide ) {
+		const { latestPosts } = this.props;
+
+		if ( latestPosts && latestPosts.length ) {
```
jeryj, line 77: Do you think this check should be outside the `initializeSwiper` function? As in
```
if ( latestPosts && latestPosts.length ) {
  initialiazeSwiper( initialSlide );
}
```

Or is it better for this function to "protect itself" by never initializing without slides?

WunderBart: I think that it's good to have it here as it should never initialize if there are no slides available. Otherwise, we'd have to make the same check inside `componentDidMount` and `componentDidUpdate`.

Issue #477, opened 2020-05-14
Display Settings disappear if Number of Items field is empty


jeryj:
When editing the Articles Carousel or Homepage Posts block Display Settings in the sidebar, you can make the `<QueryControls>` component disappear if you delete the Number of Items value from the input.

**To Reproduce**
1. Select the Number of Posts field 
2. Delete the number that's there
3. This input and all the other `<QueryControls>`  fields will disappear. 

**Screenshots**
![posts-query-disappears](https://user-images.githubusercontent.com/967608/81940898-b8341500-95bd-11ea-93ca-a6d11d2c82d8.gif)

The `<QueryControls>` component is wrapped in a `{ postsToShow && (<QueryControls ...` statement, so when `postsToShow` is empty, it removes the controls. I'm not sure on the best fix, but to possibilities would be:
- `postsToShow` can't be empty, but a minimum of 1
- Remove the `postsToShow` if statement.

PR #479, opened 2020-05-15
fix: disallow 0 as post to show value

jeffersonrabb:
### All Submissions:

It is possible to break the block's UI by clearing the `Number of Items` control, as described in https://github.com/Automattic/newspack-blocks/issues/477. This branch fixes the issue by ensuring the value is 1 or greater.

If this approach is successful, I'll make the same fix for Homepage Posts.

Closes https://github.com/Automattic/newspack-blocks/issues/477

1. On `master`, observe the scenario described in the issue.
2. Switch to this branch and attempt to repeat the scenario.
3. Verify it is impossible to set the field to a value less than 1.

### Other information:

### DIFF
--- a/src/blocks/carousel/edit.js
+++ b/src/blocks/carousel/edit.js
@@ -187,7 +187,9 @@ class Edit extends Component {
 							<QueryControls
 								enableSpecific={ false }
 								numberOfItems={ postsToShow }
-								onNumberOfItemsChange={ value => setAttributes( { postsToShow: value } ) }
+								onNumberOfItemsChange={ value =>
+									setAttributes( { postsToShow: value ? value : 1 } )
+								}
 								authors={ authors }
 								onAuthorsChange={ value => setAttributes( { authors: value } ) }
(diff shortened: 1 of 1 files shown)


jeffersonrabb:
> The approach makes sense to me, but when I test it out by deleting the input, the slider renders to a weird spot.

I think this is an issue in `QueryControls`, or perhaps even `RangeControl`. You'll see the same behavior in the Latest Posts block, and if you deselect and reselect the Carousel after this behavior the value is correctly set to 1.

jeryj:
Where would an issue get created to address the root problem? That would be good to make sure it doesn't get lost.

jeffersonrabb:
> Where would an issue get created to address the root problem? That would be good to make sure it doesn't get lost.

I can reproduce the same behavior in the `RangeControl` for autoplay delay, so it would seem to be an issue in that component rather than `QueryControls`. I'd file it in the [Gutenberg repo](https://github.com/WordPress/gutenberg).

PR #487, opened 2020-05-20
Homepage Posts: add category, tag classes to each article

### LINKED ISSUE #363: Homepage Posts: Add classes for posts categories and tags
In WordPress, on single posts, CSS classes can be added for each of the tags and categories applied to the post, using `<?php post_class(); ?>`. It would be very handy to do something similar to the `<article>` classes in the Homepage Posts block, because then they could be used to add custom styles to specific tags/categories. 

Edited to add: one good use case for this is a 'Sponsored' or similar tag or category, to do something to make it clear that these posts are different.

laurelfulford:
### All Submissions:

This PR adds classes based on each post's assigned category and tag to the `<article>` tag, using the same format that is output by the `[post_class()](https://developer.wordpress.org/reference/functions/post_class/)` function. (I opted not to use that function as it's more likely to output classes -- like `type-post` or `has-post-thumbnail` -- that themes could already be using. The tag and category classes are most likely to be used on a site-by-site basis to customize specific taxonomies). 

This change will make it possible to add custom CSS to posts with specific categories or tags within a block, rather than assigning custom CSS to all posts in a block based on a class added to that block.

Closes #363.

2. Add a Homepage Posts block to a page and publish.
3. Inspect the block in the editor and confirm that each `<article>` tag has classes assigned based on the categories the post has (using the format `category-[slug]`), and classes assigned based on the tags the post has (using the format `tag-[slug]`).
4. Inspect the block on the frontend and confirm that each `<article>` tag has the same classes assigned based on each post's categories and tags. 

### Other information:

### DIFF
--- a/class-newspack-blocks.php
+++ b/class-newspack-blocks.php
@@ -412,6 +412,32 @@ class Newspack_Blocks {
 	}
 
 	/**
+	 * Prepare an array classes based on assigned tags.
+	 *
+	 * @param string $post_id Post ID.
+	 * @return array Array of CSS classes.
+	 */
+	public static function get_term_classes( $post_id ) {
+		$classes = [];
+
+		$tags = get_the_terms( $post_id, 'post_tag' );
+		if ( ! empty( $tags ) ) {
+			foreach ( $tags as $tag ) {
+				$classes[] = 'tag-' . $tag->slug;
+			}
+		}
+
+		$categories = get_the_terms( $post_id, 'category' );
+		if ( ! empty( $categories ) ) {
+			foreach ( $categories as $cat ) {
+				$classes[] = 'category-' . $cat->slug;
+			}
+		}
+
+		return implode( ' ', $classes );
+	}
+
+	/**
 	 * Get patterns for post type.
 	 *
 	 * @param string $post_type Post type.
--- a/class-newspack-blocks.php
+++ b/class-newspack-blocks.php
@@ -412,10 +412,10 @@ class Newspack_Blocks {
 	}
 
 	/**
-	 * Prepare an array classes based on assigned tags.
+	 * Prepare a list of classes based on assigned tags and categories.
 	 *
 	 * @param string $post_id Post ID.
-	 * @return array Array of CSS classes.
+	 * @return string CSS classes.
 	 */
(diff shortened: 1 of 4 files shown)


src/blocks/homepage-articles/templates/article.php
```diff
@@ -9,10 +9,29 @@
...
+		}
 
-		$styles = '';
+		// Add classes based on the post's assigned tags.
+		$tags = get_the_terms( get_the_ID(), 'post_tag' );
+		if ( ! empty( $tags ) ) {
+			foreach ( $tags as $tag ) {
+				$classes[] = 'tag-' . $tag->slug;
+			}
+		}
```
adekbadek, line 30: Could this logic of assigning tag and category classes be moved to `Newspack_Blocks` (like `prepare_authors` is) to avoid duplication?

laurelfulford: Ah, that's a good call, Adam! I've updated it to live in Newspack_Blocks in 04f01c3.

class-newspack-blocks.php
```diff
@@ -411,6 +411,32 @@ public static function prepare_authors() {
 		);
 	}
 
+	/**
+	 * Prepare an array classes based on assigned tags.
```
adekbadek, line 415: Nitpick: tags and categories ;)

class-newspack-blocks.php
```diff
@@ -411,6 +411,32 @@ public static function prepare_authors() {
...
+	 *
+	 * @param string $post_id Post ID.
+	 * @return array Array of CSS classes.
```
adekbadek, line 418:
```suggestion
	 * @return string CSS classes.
```

class-newspack-blocks.php
laurelfulford: Gah, thanks for catching that Adam! 🤦‍♀️ Fixed in 2bf0c3d.

class-newspack-blocks.php
laurelfulford: Thanks! Fixed in 2bf0c3d, too -- the comment cleanup commit 😄

laurelfulford:
Yay! Thanks for help hammer this into shape!

PR #499, opened 2020-05-28
Improve Homepage Posts block editor performance

### LINKED ISSUE #493: Homepage Posts block editor performance
With multiple Homepage Posts block on a page, there is a very suboptimal number of requests fired in order to prevent deduplication.

adekbadek:
### All Submissions:

Improve the performance of Homepage Blocks post in the editor. 

Closes #493 

1. Create a page with `n` of Homepage Posts blocks, each with a category assigned
2. On master, load the page in the editor. Observe a larger* than `n` number of requests to `/v2/posts` WP API endpoint**
3. Switch to this branch and rebuild
4. Observe that the number of `/v2/posts` requests on editor page load matches the number of blocks on the page
4. Observe that some of the block posts query controls are disabled during the update***
5. Simulate an error by blocking the first `/v2/posts` request****. Observe that the sabotaged block displays an error, but the rest are displayed as expected. Such an error on `master` should  cause an infinite loading state on the block.  

\* I went with 10, which triggered 55 requests
\** To see the no. of requests in Chrome's network tab:

image

\*** Not all `@wordpress/components` control components support the `disabled` prop - changing that is in progress at https://github.com/WordPress/gutenberg/pull/22701

\**** In Chrome network tab:
image

### Other information:

* [X] Have you successfully ran tests with your changes locally?

### DIFF
--- a/src/blocks/homepage-articles/editor.js
+++ b/src/blocks/homepage-articles/editor.js
@@ -3,7 +3,7 @@
  */
 import { registerBlockType } from '@wordpress/blocks';
 import { BLOCK_NAME, settings } from '.';
-import registerHomepagePostsBlockStore from './store';
+import { registerQueryStore } from './store';
 
 registerBlockType( BLOCK_NAME, settings );
-registerHomepagePostsBlockStore();
+registerQueryStore();
--- a/src/blocks/homepage-articles/editor.js
+++ b/src/blocks/homepage-articles/editor.js
@@ -2,8 +2,10 @@
  * Internal dependencies
  */
 import { registerBlockType } from '@wordpress/blocks';
-import { BLOCK_NAME, settings } from '.';
+import { settings, name } from '.';
 import { registerQueryStore } from './store';
 
+const BLOCK_NAME = `newspack-blocks/${ name }`;
+
 registerBlockType( BLOCK_NAME, settings );
-registerQueryStore();
+registerQueryStore( BLOCK_NAME );
--- a/src/blocks/homepage-articles/editor.js
+++ b/src/blocks/homepage-articles/editor.js
@@ -2,8 +2,8 @@
  * Internal dependencies
  */
 import { registerBlockType } from '@wordpress/blocks';
-import { name, settings } from '.';
-import { registerQueryStore } from './store';
+import { BLOCK_NAME, settings } from '.';
+import registerHomepagePostsBlockStore from './store';
 
-registerBlockType( `newspack-blocks/${ name }`, settings );
-registerQueryStore( `newspack-blocks/${ name }` );
+registerBlockType( BLOCK_NAME, settings );
+registerHomepagePostsBlockStore();
--- a/src/components/autocomplete-tokenfield.js
+++ b/src/components/autocomplete-tokenfield.js
@@ -161,7 +161,7 @@ class AutocompleteTokenField extends Component {
 	 * Render.
 	 */
 	render() {
-		const { help, label = '' } = this.props;
+		const { help, label = '', disabled } = this.props;
 		const { suggestions, loading } = this.state;
 
 		return (
@@ -172,6 +172,7 @@ class AutocompleteTokenField extends Component {
 					onChange={ tokens => this.handleOnChange( tokens ) }
 					onInputChange={ input => this.debouncedUpdateSuggestions( input ) }
 					label={ label }
+					disabled={ disabled }
 				/>
 				{ loading && <Spinner /> }
 				{ help && <p className="autocomplete-tokenfield__help">{ help }</p> }
--- a/src/components/autocomplete-tokenfield.js
+++ b/src/components/autocomplete-tokenfield.js
@@ -161,7 +161,7 @@ class AutocompleteTokenField extends Component {
 	 * Render.
 	 */
 	render() {
-		const { help, label = '', disabled } = this.props;
+		const { help, label = '' } = this.props;
 		const { suggestions, loading } = this.state;
 
 		return (
@@ -172,7 +172,6 @@ class AutocompleteTokenField extends Component {
 					onChange={ tokens => this.handleOnChange( tokens ) }
 					onInputChange={ input => this.debouncedUpdateSuggestions( input ) }
 					label={ label }
-					disabled={ disabled }
 				/>
 				{ loading && <Spinner /> }
 				{ help && <p className="autocomplete-tokenfield__help">{ help }</p> }
--- a/src/components/query-controls.js
+++ b/src/components/query-controls.js
@@ -159,6 +159,7 @@ class QueryControls extends Component {
 			tagExclusions,
 			onTagExclusionsChange,
 			enableSpecific,
+			disabled,
 		} = this.props;
 		const { showAdvancedFilters } = this.state;
 
@@ -169,6 +170,7 @@ class QueryControls extends Component {
 					checked={ specificMode }
 					onChange={ onSpecificModeChange }
 					label={ __( 'Choose Specific Posts', 'newspack-blocks' ) }
+					disabled={ disabled }
 				/>
 			),
 			specificMode && (
@@ -183,6 +185,7 @@ class QueryControls extends Component {
 						'Begin typing post title, click autocomplete result to select.',
 						'newspack-blocks'
 					) }
+					disabled={ disabled }
 				/>
 			),
 			! specificMode && <BaseControl key="queryControls" { ...this.props } />,
@@ -194,6 +197,7 @@ class QueryControls extends Component {
 					fetchSuggestions={ this.fetchAuthorSuggestions }
 					fetchSavedInfo={ this.fetchSavedAuthors }
 					label={ __( 'Authors', 'newspack-blocks' ) }
+					disabled={ disabled }
 				/>
 			),
 			! specificMode && onCategoriesChange && (
@@ -204,6 +208,7 @@ class QueryControls extends Component {
 					fetchSuggestions={ this.fetchCategorySuggestions }
 					fetchSavedInfo={ this.fetchSavedCategories }
 					label={ __( 'Categories', 'newspack-blocks' ) }
+					disabled={ disabled }
 				/>
 			),
 			! specificMode && onTagsChange && (
@@ -214,6 +219,7 @@ class QueryControls extends Component {
 					fetchSuggestions={ this.fetchTagSuggestions }
 					fetchSavedInfo={ this.fetchSavedTags }
 					label={ __( 'Tags', 'newspack-blocks' ) }
+					disabled={ disabled }
 				/>
 			),
 			! specificMode && onTagExclusionsChange && (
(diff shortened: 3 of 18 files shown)


laurelfulford:
I've tested a few cases in the editor, more with an eye to see if things were working the same way as before. Adding and removing blocks works really well, whether the blocks are added/removed at the top of the page, bottom or middle.

I did run into an issue when reordering the blocks -- it looks like the posts aren't updated at all when the order is changed, though they correct themselves when doing another action that updates them (like add/delete):

reordering

I also ran into something weird when changing the number of posts per block -- for the most part, this updated pretty well, but from time to time I managed to "trick" it, and could set the number of posts per block to something that would show under Display Settings, but wouldn't be reflected in the editor. The following clip happened after changing the counts of a few of the blocks; the third block is correctly showing 2 posts from the previous change:

https://cloudup.com/cAY3FIeblcF

Let me know if I can provide any info about either of these!

adekbadek:
Thanks for catching those, Laurel!

1. Update on block reordering added here: 634aba0ea46b971fddcc6c7d3a0995d97d8efabb
2. Wrong counts issue: I could not reproduce that, could you please provide a list of steps to reproduce?

laurelfulford:
Thanks Adam! Block reordering looks good now in my testing 👍 

> Wrong counts issue: I could not reproduce that, could you please provide a list of steps to reproduce?

I could recreate a couple times by doing the following:

1. Added 7 blocks to the editor, displaying 3 stories each.
2. Went up and down the page, picking a block and increasing the stories in a block up to 6, or back down to three if I'd already hit it. To change the count, I would click on the number text input, and use my up/down keyboard arrows. 
3. Each time I would wait for the posts to stop loading before editing the next block.

At some point, the count gets weird -- the first time I tried to recreate today, the second block I changed was set to 6 posts, but only displayed 5. The second time it took a bit more editing; after changing 4-5 blocks, I ended up with one that says it has a count of 3, but is actually displaying 7. 

My localhost can get pretty slow -- that could also be contributing, I'm not sure. 

Just let me know if it'd be more helpful to do a quick screenshare or something -- issues that are hard to recreate kind of drive me nuts, you never know if they're happening in their own weird bubble (like my slow localhost), or if there's just a really specific step that is hard to hit!

adekbadek:
I figured out a way to reproduce that:

1. insert a single block, set it to `3` posts (which is the default)
2. change it to `10`, then immediately back to `3`
4. observe that the block has `10` posts (after fetch completes), but input displays `3`

In pt. 2 the `3` posts request is cached, so it completes before `10` request. When `10` comes back, posts are replaced with the fetched amount.

The reason is twofold:
1. the input is not disabled (that's fixed in https://github.com/WordPress/gutenberg/pull/22701)
2. `redux-saga`s `debounce` effect does not cancel the effect "from before", which may result in a race condition

I've solved two by implementing a slightly different debouncing strategy: a1e63452ce1f4744a81afde0767f81a5ecba26a4 and it should work as expected now, even without the disabling of the input.

jeffersonrabb:
I think we should consider removing the disabled states. From a UI perspective it's nice to show the activity, but on pages with a large number of blocks this leads to significant periods of time with locked UI. Here's a video showing this. It was taken on a Jurassic Ninja instance with approximately 300 Posts, on a Page with 27 Homepage Posts blocks. Several Newspack sites have homepages of similar or greater density. 

https://cloudup.com/cTmN9CMSh9V

@adekbadek what do you think?

adekbadek:
Done in 1600925850b562eceab30487fb83c42c9ff607fc . Note that just the sidebar UI controls' disabling is removed, the blocks will still be dimmed until the reflow completes.

src/blocks/homepage-articles/editor.js
```diff
@@ -2,8 +2,8 @@
...
 import { registerBlockType } from '@wordpress/blocks';
-import { name, settings } from '.';
-import { registerQueryStore } from './store';
+import { BLOCK_NAME, settings } from '.';
+import registerHomepagePostsBlockStore from './store';
 
-registerBlockType( `newspack-blocks/${ name }`, settings );
-registerQueryStore( `newspack-blocks/${ name }` );
+registerBlockType( BLOCK_NAME, settings );
+registerHomepagePostsBlockStore();
```
jeffersonrabb, line 9: Renaming the store registration function may break the implementation in Full Site Editing, here: https://github.com/Automattic/wp-calypso/blob/master/apps/full-site-editing/full-site-editing-plugin/newspack-blocks/blog-posts-block-editor.js#L14

adekbadek:
> In FSE the block is registered with a different name

I think that's ok – FSE renames the block, so it doesn't matter what's the original name. And this PR does not change [the `name` of the block](https://github.com/Automattic/newspack-blocks/pull/499/files#diff-fbe7f8c598cf05d4603ba49fec909dedR27) – there's just some [refactoring](https://github.com/Automattic/newspack-blocks/pull/499/files#diff-ae86db7f5c511e7d4264d7614bfb4791L8-R8) to avoid duplication.

> store is initiated from a different place

Good catch, I've updated the code in d1ada72ccfbb37638bdf3ee9ba9ad4928843cc10 to be backwards-compatible.

> a few other differences […] changes in this branch are likely to break that implementation

Besides the [`settings` and `registerQueryStore` importing](https://github.com/Automattic/wp-calypso/blob/25d5220d68eaa859891c6276acfbc2ee649ceaae/apps/full-site-editing/full-site-editing-plugin/newspack-blocks/blog-posts-block-editor.js#L13-L14) I can't see any other "points of contact". FSE integration uses this repo's `edit.js`, but it's own version of `editor.js` – which register the block & store. I think the changes in this PR are now backwards-compatible.

> since the blocks may take a noticeable interval to update and since the changes applied mid-update are harmless, removing the state change might remove friction from the editing process.

I agree that not disabling the inputs makes for a smoother editing experience, but "dimming" of the block content does not prevent the user from editing. I think it's good feedback that the reflow is still in progress – perhaps there could be less dimming, so that it's more subtle?

adekbadek:
As @jeffersonrabb pointed out, some changes have indeed broken backwards compatibility – namely internal handling the block name. Fixed in 4ee4b33

src/components/query-controls.js
```diff
@@ -159,6 +159,7 @@ class QueryControls extends Component {
 			tagExclusions,
 			onTagExclusionsChange,
 			enableSpecific,
+			disabled,
```
jeffersonrabb, line 162: Can be removed?

src/components/query-controls.js
```diff
@@ -169,6 +170,7 @@ class QueryControls extends Component {
 					checked={ specificMode }
 					onChange={ onSpecificModeChange }
 					label={ __( 'Choose Specific Posts', 'newspack-blocks' ) }
+					disabled={ disabled }
```
jeffersonrabb, line 173: Can be removed?

src/components/query-controls.js
```diff
@@ -183,6 +185,7 @@ class QueryControls extends Component {
 						'Begin typing post title, click autocomplete result to select.',
 						'newspack-blocks'
 					) }
+					disabled={ disabled }
```
jeffersonrabb, line 188: Can be removed?

src/components/query-controls.js
```diff
@@ -194,6 +197,7 @@ class QueryControls extends Component {
 					fetchSuggestions={ this.fetchAuthorSuggestions }
 					fetchSavedInfo={ this.fetchSavedAuthors }
 					label={ __( 'Authors', 'newspack-blocks' ) }
+					disabled={ disabled }
```
jeffersonrabb, line 200: Can be removed?

src/components/query-controls.js
```diff
@@ -204,6 +208,7 @@ class QueryControls extends Component {
 					fetchSuggestions={ this.fetchCategorySuggestions }
 					fetchSavedInfo={ this.fetchSavedCategories }
 					label={ __( 'Categories', 'newspack-blocks' ) }
+					disabled={ disabled }
```
jeffersonrabb, line 211: Can be removed?

src/components/query-controls.js
```diff
@@ -214,6 +219,7 @@ class QueryControls extends Component {
 					fetchSuggestions={ this.fetchTagSuggestions }
 					fetchSavedInfo={ this.fetchSavedTags }
 					label={ __( 'Tags', 'newspack-blocks' ) }
+					disabled={ disabled }
```
jeffersonrabb, line 222: Can be removed?

src/components/query-controls.js
```diff
@@ -236,6 +242,7 @@ class QueryControls extends Component {
 					fetchSuggestions={ this.fetchTagSuggestions }
 					fetchSavedInfo={ this.fetchSavedTags }
 					label={ __( 'Excluded Tags', 'newspack-blocks' ) }
+					disabled={ disabled }
```
jeffersonrabb, line 245: Can be removed?

src/components/autocomplete-tokenfield.js
```diff
@@ -161,7 +161,7 @@ class AutocompleteTokenField extends Component {
 	 * Render.
 	 */
 	render() {
-		const { help, label = '' } = this.props;
+		const { help, label = '', disabled } = this.props;
```
jeffersonrabb, line 164: Remove?

src/components/autocomplete-tokenfield.js
```diff
@@ -172,6 +172,7 @@ class AutocompleteTokenField extends Component {
 					onChange={ tokens => this.handleOnChange( tokens ) }
 					onInputChange={ input => this.debouncedUpdateSuggestions( input ) }
 					label={ label }
+					disabled={ disabled }
```
jeffersonrabb, line 175: Remove?

src/components/query-controls.js
adekbadek: Done in 3580fa7284e4cfdf8c81a8f4293c59bb894c9f14

src/components/query-controls.js
adekbadek: Done in 3580fa7284e4cfdf8c81a8f4293c59bb894c9f14

src/components/query-controls.js
adekbadek: Done in 3580fa7284e4cfdf8c81a8f4293c59bb894c9f14

src/components/query-controls.js
adekbadek: Done in 3580fa7284e4cfdf8c81a8f4293c59bb894c9f14

src/components/query-controls.js
adekbadek: Done in 3580fa7284e4cfdf8c81a8f4293c59bb894c9f14

src/components/query-controls.js
adekbadek: Done in 3580fa7284e4cfdf8c81a8f4293c59bb894c9f14

src/components/query-controls.js
adekbadek: Done in 3580fa7284e4cfdf8c81a8f4293c59bb894c9f14

src/components/autocomplete-tokenfield.js
adekbadek: Done in 3580fa7284e4cfdf8c81a8f4293c59bb894c9f14

src/components/autocomplete-tokenfield.js
adekbadek: Done in 3580fa7284e4cfdf8c81a8f4293c59bb894c9f14

PR #503, opened 2020-06-02
Homepage Posts: Remove multiple image sizes from the homepage posts block

laurelfulford:
### All Submissions:

This is an in-progress PR to remove as many `add_image_size()`s from the Homepage Posts block as possible.

Originally the block generated multiple sizes of the same image shape, in hopes that the featured images would actually be cropped (if the original is smaller than the crop size, it's not cropped). This creates a ton of extra images, though -- 13 per post. 

This PR switches the block to just create one image (cropped to 1200px wide by infinite height -- it uses the original 'uncropped' option) and creates a visual crop for the images using CSS and `object-fit: cover`. 

Current known issues: 

* **It still creates a new image for every post.** While better than 13, it's still... not great. The Article Carousel block uses the 'Large' image size already included in WordPress, which is, by default 1024 x 1024. This could be a possibility, if we can switch the default for Newspack sites to 1200px wide (to match the homepage content width), but I'm not sure if doing that on WP.com sites would have unexpected consequences -- for one, it's a size users can technically edit; though it's buried in WP Admin we can't safely assume it will always be 1024x1024. For two, the smaller images could appear lower quality, or appear to be cropped differently than the original. So it'll need more testing to determine whether removing that image size will cause unexpected issues. 
* **IE11 doesn't support `object-fit`** There is some alternative IE11-specific CSS that gets this close, but there are cases where it doesn't create a 1:1 outcome. One example is if you have a landscape-oriented image and are using the landscape shape, if the original image is a more extreme aspect ratio (say, 16:9 vs the block's 4:3), it will retain that aspect ratio, and appear shorter -- the first post in this screenshot illustrates that:

image

Something similar happens with a narrower portrait oriented image. 

Next steps:

* This has been tested pretty thoroughly with the Newspack theme and different settings, but I need to run it through some WP.com themes to make sure it doesn't have unexpected effects. The styles added to the block are specific enough that they won't affect the theme, but some themes may have styles that would affect the block. 

1.
2.
3.

### Other information:


laurelfulford:
After some brief testing with the FSE themes on WP.com, I can see this approach isn't going to be as straight forward as I had hoped. The themes override the block's 100% width on images, so they're not filling the available space. The styles in this PR expect the images to be trying to fill the available space, so it's quite a striking visual difference. 

Here's an example of the Leven theme before apply this PR (smaller images), and after:

https://cloudup.com/iqpSopxRjGd

The editor has as similar issues, but the images are not sized up as large (they are also incorrectly aligned left in the 'before' but that seems like a separate thing to worry about):

https://cloudup.com/imLubUcMOKL

This issue is most noticeable when the potential image area is big and the original image is small; you can see as you scroll down the first examples, the images get closer and closer to being the same the larger the originals were.  

I'm not sure the best way to move forward from here. We could move the sizing styles to the Newspack theme (including all of the `object-fit` stuff), but that still leaves the question of how to set the correct sizes for these images on WP.com.

laurelfulford:
In 39c859a, I updated the code to grab the featured image's original width, so we can use it to set a max-width so the containing element does not get much larger than the actual image.

The images are still not 1:1 (with the update, small images will always respect the set aspect ratio, but may be smaller), but it's closer:

https://cloudup.com/cyTpyLuQUIR

laurelfulford:
Posting another update on this still in-progress PR:
* The original approach generated multiple images to make sure even smaller original images would be cropped to the right size. This has been replaced with a function that will pull a Photon-generated image of the same size, and, when Photon is not enabled, pull the 'Large' thumbnail image size. 
* CSS has been added to help set the correct image orientation (so even if the block is falling back to the Large image, it will use the correct aspect ratio).
* I've also reworked the 'behind' image placement a bit, since it never makes sense for it to use the landscape, portrait or square options; so it always uses the uncropped (but scaled down to 1200px wide via Photon, or the large thumbnail) version of the image. 
* I've started testing and making screencaps on a Newspack site with and without Photon, and on WP.com site using Varia and Twenty Twenty to start. I'll follow up with screenshots of specifics, but so far the main visual change seems to be with too small images. Originally, they would not be cropped and would be sized up to fit the available space; now they will be cropped to the correct shape via CSS, and, if quite small, will be a very small version of that shape. Switching the block to use 'uncropped' will get it closer to the original behaviour, so I think this will be an acceptable change, but would appreciate feedback.

Next steps before marking PR ready for review:
* Post existing screenshots, highlighting differences of note.
* Test updates with some of the existing FSE patterns, to see if there are issues.

laurelfulford:
There's a lot of information that goes with this PR, so I'm going to close this version and open a fresh one, so the relevant, current comments are the focus.

PR #512, opened 2020-06-05
fix: don't use main wp_query when rendering homepage posts

jeffersonrabb:
### All Submissions:

Full description and testing instructions: https://github.com/Automattic/newspack-popups/pull/136

To distinguish the `WP_Query` in Homepage Posts from the main loop, this branch holds the `wp_query` global in a temporary variable, then restores it after the block is rendered.

h/t @claudiulodro for the fix!

See https://github.com/Automattic/newspack-popups/pull/136

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/templates/articles-loop.php
+++ b/src/blocks/homepage-articles/templates/articles-loop.php
@@ -30,8 +30,8 @@ call_user_func(
 			$post_counter++;
 			echo Newspack_Blocks::template_inc( __DIR__ . '/article.php', array( 'attributes' => $attributes ) ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
 		}
-
 		$wp_query = $main_query; // phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
+		wp_reset_postdata();
 	},
 	$data // phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UndefinedVariable
 );
--- a/src/blocks/homepage-articles/templates/articles-loop.php
+++ b/src/blocks/homepage-articles/templates/articles-loop.php
@@ -10,8 +10,15 @@
 
 call_user_func(
 	function( $data ) {
+		global $wp_query;
+		$main_query = $wp_query;
+		wp_reset_postdata();
+
 		$attributes    = $data['attributes'];
 		$article_query = $data['article_query'];
+
+		$wp_query = $article_query; // phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
+
 		global $newspack_blocks_post_id;
 		$post_counter = 0;
 		while ( $article_query->have_posts() ) {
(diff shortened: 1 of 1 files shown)


src/blocks/homepage-articles/templates/articles-loop.php
```diff
@@ -23,7 +30,8 @@ function( $data ) {
 			$post_counter++;
 			echo Newspack_Blocks::template_inc( __DIR__ . '/article.php', array( 'attributes' => $attributes ) ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
 		}
-		wp_reset_postdata();
+
+		$wp_query = $main_query; // phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
```
claudiulodro, line 34:
```suggestion
		$wp_query = $main_query; // phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
		wp_reset_query();
```

I believe the query needs another reset after we put it back.

src/blocks/homepage-articles/templates/articles-loop.php
```diff
@@ -10,8 +10,15 @@
 
 call_user_func(
 	function( $data ) {
+		global $wp_query;
+		$main_query = $wp_query;
+		wp_reset_postdata();
```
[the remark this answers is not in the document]
jeffersonrabb: > I believe you want `wp_reset_query` here instead.

I originally had `wp_reset_query` but I get this phpcs warning: `The use of function wp_reset_query() is discouraged; use wp_reset_postdata() instead`

Do we need `wp_reset_query` in spite of the warning?

Also note: https://github.com/Yoast/wordpress-seo/issues/558

src/blocks/homepage-articles/templates/articles-loop.php
jeffersonrabb: > I believe the query needs another reset after we put it back.

Works fine when I do it 2x, but I'm not sure I completely understand why it's needed? By replacing the global with the stashed query aren't we destroying everything that just happened anyway?

claudiulodro: Good point. At minimum I think the `wp_reset_postdata` is still necessary, for resetting the global `$post`.

src/blocks/homepage-articles/templates/articles-loop.php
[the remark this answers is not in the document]
claudiulodro: Interesting. Thanks for sharing that link. Let's go with `wp_reset_postdata`. 👌 We're not going to be using `query_posts` anyways.

src/blocks/homepage-articles/templates/articles-loop.php
claudiulodro: Otherwise WP will think that we're still using the last post of the query, so we'll get weird side-effects in the footer and other areas under the homepage block.

jeffersonrabb:
> This needs the post reset at the end of the block again, but otherwise it looks good. Just a heads-up this is still marked as a draft PR.

Done in https://github.com/Automattic/newspack-blocks/pull/512/commits/d9d01a664b7a1132bf67b1c5f49fed965b763fee and taken out of Draft mode.

jeffersonrabb:
Curiously I noticed this PR resolves the issue even without the change in the Newspack Campaigns plugin, since the `is_singular()` check now accurately weeds out the Homepage Posts loop. Once this PR lands, I'll probably put the other one on hold so we don't introduce checks that aren't needed.

PR #515, opened 2020-06-08
fix(homepage-posts): exclude posts already on the page from query

### LINKED ISSUE #510: Blog Posts Block: Load More Posts button only shows on first of multiple block instances
Multiple **Blog Posts** blocks on a single page seem to never show more than one **Load More Posts** button on Preview & published pages. Reproduced at https://widgettheworldwatcherfansite.blog/blog-posts-block-test/ as follows:

- Three **Blog Posts** blocks, all set to 3 items/3 columns/**Show More** enabled
- Each **Blog Posts** block set to show posts from different exclusive categories, 6/3/6 posts respectively

**Findings**

- Each block with **Show More** enabled shows a **Load More Posts** button while editing, but this is not necessarily reflected in Preview or published post
- Regardless of whether first block on page shows a **Load More Posts** button or not, no block further down will show a **Load More Posts** button
- Regardless of whether first block on page has **Show More** enabled or not, no block further down will show a **Load More Posts** button

Allowing multiple instances of **Blog Posts** block on page seems to be a relatively recent development (March 2020, p5cPuC-29b-p2#comment-2468).

_21606477-hc • 3029980-zd • p1591268313195200-slack-wpcom-support_

_Originally posted by @tullyhansen in https://github.com/Automattic/newspack-blocks/issues/507#issuecomment-638779333 – on reflection seemed to warrant a separate issue_

p-jackson:
### All Submissions:

The "Load more posts" button isn't always appearing on the front end of the site, even when the site is public and there are definitely more posts to load.

**Without this PR**
Screenshot 2020-06-08 at 5 40 58 PM

I think this is to do with the code that prevents a post from appearing twice on the same page. From what I can tell the block [keeps a track of which posts have already been rendered on the page](https://github.com/Automattic/newspack-blocks/blob/31181a5536aef03779cf789d2dcd4d9de05b4371/src/blocks/homepage-articles/templates/articles-loop.php#L22) so that it doesn't render them again. I think the idea is to [over-fetch posts](https://github.com/Automattic/newspack-blocks/blob/31181a5536aef03779cf789d2dcd4d9de05b4371/class-newspack-blocks.php#L364)  when rendering the following blocks to make up for the fact that some of the posts [will be filtered out](https://github.com/Automattic/newspack-blocks/blob/31181a5536aef03779cf789d2dcd4d9de05b4371/src/blocks/homepage-articles/templates/articles-loop.php#L20).

The problem is that because size of the response page is larger than we actually render, it messes with the logic for deciding whether there are more pages of response to load. Which means the more posts button gets hidden.

This PR changes the approach so that rather than over-fetching posts, it excludes posts that have already rendered from the query.

A downside of this approach could be that the exclusion list grows in an unbounded way because the author can add many blocks which show many posts each. So the query size is unbounded. On the other hand by over-fetching we are growing the size of the query response body in an unbounded way and then throwing away the results. So I can't imagine query performance would be worse than before.

**With this PR**
Jun-08-2020 17-38-44

Closes #510

Addition: The change in query approach is contingent on a `wp-config.php` constant, to allow easy switching between old/new approaches to assess performance. To test this branch, add this to `wp-config.php`:

`define( 'NEWSPACK_BLOCKS_EXPERIMENTAL_MODE', 1 );`

1. Create 2 categories, e.g. 'odd' and 'even'
2. Create 6 posts, assign 3 to each of the new categories
3. Create a new page
4. Add a homepage posts block, configure to show 2 blocks, filter by the 'odd' category, and show the more posts button
5. Add another home posts block, configure to show 2 blocks, filter by the 'even' category, and show the more posts button
6. Preview the page
7. The more posts button should appear on both blocks, and when clicked should display the extra

### DIFF
--- a/class-newspack-blocks.php
+++ b/class-newspack-blocks.php
@@ -361,9 +361,15 @@ class Newspack_Blocks {
 			$args['post__in'] = $specific_posts;
 			$args['orderby']  = 'post__in';
 		} else {
-			$args['posts_per_page'] = $posts_to_show + count( $newspack_blocks_post_id );
+			$args['posts_per_page'] = $posts_to_show;
 			if ( count( $newspack_blocks_all_specific_posts_ids ) ) {
 				$args['post__not_in'] = $newspack_blocks_all_specific_posts_ids;
+			}
+			if ( count( $newspack_blocks_post_id ) ) {
+				$args['post__not_in'] = array_merge(
+					$args['post__not_in'] ?? [],
+					array_keys( $newspack_blocks_post_id )
+				);
 			}
 			if ( $authors && count( $authors ) ) {
 				$args['author__in'] = $authors;
--- a/class-newspack-blocks.php
+++ b/class-newspack-blocks.php
@@ -362,10 +362,13 @@ class Newspack_Blocks {
 			$args['orderby']  = 'post__in';
 		} else {
 			$args['posts_per_page'] = $posts_to_show;
+			if ( ! self::is_experimental_mode() ) {
+				$args['posts_per_page'] += count( $newspack_blocks_post_id );
+			}
 			if ( count( $newspack_blocks_all_specific_posts_ids ) ) {
 				$args['post__not_in'] = $newspack_blocks_all_specific_posts_ids;
 			}
-			if ( count( $newspack_blocks_post_id ) ) {
+			if ( self::is_experimental_mode() && count( $newspack_blocks_post_id ) ) {
 				$args['post__not_in'] = array_merge(
 					$args['post__not_in'] ?? [],
 					array_keys( $newspack_blocks_post_id )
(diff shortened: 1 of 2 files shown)


jeffersonrabb:
Hi @p-jackson, when the block rendering was first written we were concerned about the performance risks of `post__not_in` queries and opted for the overshoot approach instead. We now have a number of sites in production with large numbers of Homepage Posts blocks on the front page, so I'm going to start by A/B-ing this branch against master and noting performance differences. Hopefully we'll find the exclusion query is the clear winner.

p-jackson:
How long do we generally run A/B tests for? Just wondering if there's any action I need to take on this.

jeffersonrabb:
> How long do we generally run A/B tests for? Just wondering if there's any action I need to take on this.

Here's how I propose we do this: 

- In https://github.com/Automattic/newspack-blocks/pull/515/commits/bfb4619dec66d0c23e04532875a296a56e2a4f3b I put the new approach behind a config flag (`NEWSPACK_BLOCKS_EXPERIMENTAL_MODE`), so we can easily switch between the two approaches on a production site. (There's also a little bit in `articles-loop.php` which should be removed when we switch to the exclusion query approach)
- We'll review/merge/release/deploy on Newspack sites.
- Later today there we're setting up a stress test instance with tens of thousands of fake posts and a complex homepage. 
- First thing will be to test the exclusion query approach on this site to verify no disasters.
- Then, we'll do some tests on real sites where we switch between the two approaches and compare Lighthouse scores before and after. 

This should give us a full picture, and if the exclusion performance is equivalent or better we can do a new PR removing the flag and switching the query approach.

PR #530, opened 2020-06-17
refactor: remove experimental flag for frontend exclusion queries

jeffersonrabb:
### All Submissions:

This PR fully adopts the approach of https://github.com/Automattic/newspack-blocks/pull/515, using `post__not_in` queries rather than the original approach of "overshooting" and removing duplicates in PHP. This PR removes the experimental flag and applies the exclusion approach everywhere. 

Using the experimental flag some testing was done on production sites comparing the performance of the old and new approaches. Results were similar but the new approach looks a bit faster, and since it resolves the Load More issue it should be adopted.

cc: @p-jackson 

1. Create a complex homepage. Verify the results seen in the editor match the frontend, with no duplicate posts shown. 
2. Go through testing instructions in https://github.com/Automattic/newspack-blocks/pull/515 and verify the Load More buttons are functioning as expected.

### Other information:

### DIFF
--- a/class-newspack-blocks.php
+++ b/class-newspack-blocks.php
@@ -361,19 +361,14 @@ class Newspack_Blocks {
 			$args['post__in'] = $specific_posts;
 			$args['orderby']  = 'post__in';
 		} else {
-			$args['posts_per_page'] = $posts_to_show;
-			if ( ! self::is_experimental_mode() ) {
-				$args['posts_per_page'] += count( $newspack_blocks_post_id );
-			}
+			$args['posts_per_page'] = $posts_to_show + count( $newspack_blocks_post_id );
 			if ( count( $newspack_blocks_all_specific_posts_ids ) ) {
 				$args['post__not_in'] = $newspack_blocks_all_specific_posts_ids;
 			}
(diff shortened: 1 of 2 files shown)


class-newspack-blocks.php
```diff
@@ -361,19 +361,14 @@ function ( $acc, $block ) use ( $block_name ) {
 			$args['post__in'] = $specific_posts;
 			$args['orderby']  = 'post__in';
 		} else {
-			$args['posts_per_page'] = $posts_to_show;
-			if ( ! self::is_experimental_mode() ) {
-				$args['posts_per_page'] += count( $newspack_blocks_post_id );
-			}
+			$args['posts_per_page'] = $posts_to_show + count( $newspack_blocks_post_id );
```
p-jackson, line 364:
```suggestion
			$args['posts_per_page'] = $posts_to_show;
```

I think the logic should be the other way round: we were adding the count if it was **not** experimental

jeffersonrabb: Oops, quite correct! Fixed in https://github.com/Automattic/newspack-blocks/pull/530/commits/19c7a6c5946eb9cbd60e838f6b9b9db06dda0855

Issue #516, opened 2020-06-08
Homepage: Add filters to manually override specific posts 


benmay:
Most publishing businesses we work with like to reserve the premium realestate at the top of their site for a more manually curated list of articles that at "Home page editor" or otherwise, will manage throughout the course of a day/week.

The current homepage blocks make this a very difficult task to complete. The only workaround right now, is an editor has to go into the 3 cols on homepage and manually remove and re-add a "Specific Post" - but this is highly risky that they'll accidentally adjust an block setting or what not.

It also is difficult to transition a story from slot 1 to slot 2. In the past, we've built custom logic either directly, or even tools like ACF to enable authors to easily drag and drop articles, very easily and without risk of breaking the layout of the home page.

 Example:

![home_page_curation](https://user-images.githubusercontent.com/653266/84091371-8769c300-aa37-11ea-8d24-370beab31176.jpg)

We've currently tried this approach with Newspack - and the Block itself lacks a lot of filters.

What could be a very easy way to approach this would be if we could get some additional filters added to `public static function build_articles_query( $attributes ) {` so we could override those specific post IDs on render, and we could build a simple UI elsewhere to set the values of what post IDs should be shown.

Would be happy to submit a PR for the filters that could make this happen.

gamebits:
> It also is difficult to transition a story from slot 1 to slot 2.

This may be addressed by https://github.com/Automattic/newspack-blocks/pull/1502.

PR #532, opened 2020-06-22
New filter to get and apply a custom set of posts for newspack blocks

jdau-n:
### All Submissions:

Add a filter to allow posts in newspack blocks to be set dynamically (from an ACF repeater, for instance).

1. Create a filter that passes through post IDs for display.
2. The posts will display in newspack blocks were appropriate.

### Other information:

### DIFF
--- a/class-newspack-blocks.php
+++ b/class-newspack-blocks.php
@@ -357,6 +357,21 @@ class Newspack_Blocks {
 			'suppress_filters'    => false,
 			'ignore_sticky_posts' => true,
 		);
+
+		$custom_post_set = apply_filters( 'newspack_set_posts', array() );
+		// Remove post IDs that are part of the excluion array.
+		$custom_post_set = array_diff( $custom_post_set, $newspack_blocks_all_specific_posts_ids );
+		if ( ! empty( $custom_post_set ) ) {
+			// There are still posts in the custom list that have not been shown, so force specific mode.
+			$specific_mode = true;
+
+			// Pop off a number of posts equal to $posts_to_show, set them as the specific_posts array.
+			$specific_posts = array_slice( $custom_post_set, 0, $posts_to_show );
+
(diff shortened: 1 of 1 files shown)


benmay:
This filter is how we were able to address building manual curation plugins for the home page. Certainly happy to look at other ways of achieving it.

https://github.com/Automattic/newspack-blocks/issues/516

claudiulodro:
Hi. Thanks for the contribution!

I think [changing this line ](https://github.com/Automattic/newspack-blocks/blob/6261e3ff50d325471e981bd6dd0bc4b133facdfa/class-newspack-blocks.php#L385)(`return $args;`) into the following would let you modify the query however you want without introducing extra logic into Newspack Plugin:

```
return apply_filters( 'newspack_articles_query_args', $args, $attributes );
```

It seems much simpler but just as effective IMO.

benmay:
Thank you @claudiulodro - will check in on this, and maybe you've solved it in one!

claudiulodro:
No worries. If it'll work for you, feel free to modify this PR or I can create a new PR.

jdau-n:
Thank you for looking at this @claudiulodro. My only concern about your proposed solution is that it will require handling of the $newspack_blocks_all_specific_posts_ids array from outside the plugin to prevent duplicate IDs from appearing in any other section.

It is not impossible as the variable is a global, but my concern here is future refactors changing the variable name or not making it available outside the plugin scope, so perhaps it is worth adding two filters? One for the wp_query $args and another to filter the $newspack_blocks_all_specific_posts_ids array, which should fix the problem in a sustainable way into the future.

Let me know what you think!

claudiulodro:
Hi folks, is this filter still something you need?

benmay:
@claudiulodro it would be handy to have this kind of control when needed

adekbadek:
Closing for now.

PR #556, opened 2020-07-28
feat: experimental title-only specific posts lookup endpoint

jeffersonrabb:
### All Submissions:

We've received feedback from Bangor Daily News that specific posts lookups in the Homepage Posts block can take 45-60 seconds, making homepage management a painfully slow operation. This seems to be because we're using the `/wp/v2/search` endpoint which does a full-text search. The site in question is very large (over 400k posts), so this operation runs slow. This branch introduces a new endpoint that queries `post_title` only, which should be orders of magnitude faster. Although I suspect most use of the specific posts field does use the title, this change is a feature removal, so for the moment it's behind a new `wp-config` flag (`NEWSPACK_BLOCKS_EXPERIMENTAL`). If this fix works well and we determine the full text search isn't essential, we can remove the flag and apply it universally.

1. Add `define( 'NEWSPACK_BLOCKS_EXPERIMENTAL', true );` to `wp-config.php`.
2. Checkout and build this branch.
3. Add a Homepage Posts block to a page.
4. Toggle "Specific Posts" on, try searching for various posts, verify the results are as expected.
5. Optionally, set `NEWSPACK_BLOCKS_EXPERIMENTAL` to `false` and compare request times. On a small instance the difference may be unnoticeable, but on a larger instance (e.g. the Newspack McMansion) it may be clear.

### Other information:

### DIFF
--- a/class-newspack-blocks-api.php
+++ b/class-newspack-blocks-api.php
@@ -276,6 +276,31 @@ class Newspack_Blocks_API {
 	}
 
 	/**
+	 * Register specific posts endpoint.
+	 */
+	public static function register_post_lookup_endpoint() {
+		if ( ! Newspack_Blocks::use_experimental() ) {
+			return;
+		}
+		register_rest_route(
+			'newspack-blocks/v1',
+			'/specific-posts',
+			[
+				'methods'  => \WP_REST_Server::READABLE,
+				'callback' => [ 'Newspack_Blocks_API', 'specific_posts_endpoint' ],
+				'args'     => [
+					'search'   => [
+						'sanitize_callback' => 'sanitize_text_field',
+					],
+					'per_page' => [
+						'sanitize_callback' => 'absint',
+					],
+				],
+			]
+		);
+	}
+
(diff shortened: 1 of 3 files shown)


class-newspack-blocks-api.php
```diff
@@ -286,6 +311,54 @@ public static function video_playlist_endpoint( $request ) {
...
+
+	/**
+	 * Add title wildcard search to post lookup query.
+	 *
+	 * @param String   $where Where clause.
+	 * @param WP_Query $query The query.
+	 */
+	public static function add_post_title_wildcard_search( $where, $query ) {
+		$search = ! empty( $query->query['title_wildcard_search'] ) ? $query->query['title_wildcard_search'] : null;
+		$where .= ' AND post_title LIKE "%' . $search . '%" ';
```
claudiulodro, line 358: Probably want to run `$search` through [$wpdb->esc_like](https://codex.wordpress.org/Class_Reference/wpdb/esc_like).

PR #563, opened 2020-08-05
Homepage Posts, Post Carousel Blocks: Add sponsored labels

laurelfulford:
### All Submissions:

This PR adds some visuals elements when a post is marked as native sponsored content using the Newspack Sponsors plugin (the underwritten posts only show visual elements on single posts).

It needs to be tested in partnership with https://github.com/Automattic/newspack-theme/pull/1008 to apply the correct styles.

Closes #563

1. Install and activate the Newspack Sponsors plugin (ping myself or Derrick if unsure where to find it).
2. Switch your site to use AMP transitional so you can test both the AMP and non-AMP versions, just to make sure there's no weirdness with the scaled down logos.
3. Add at least four sponsors; make sure to include the following combinations; make sure to use at least one horizontal and one vertically-oriented logo:

* Set scope to native content, attached to a category
* Set scope to native content, attached to a specific story
* Set scope to underwritten content, attached to a category
* Set scope to underwritten content, attached to a specific story.
* Make sure at least one story will end up with more than one sponsor.

4. Set up at least one Homepage Posts block and at least one Post Carousel block on the homepage. Make sure both will show at least one post that's native content, one that's underwritten content, and one that has no sponsor assigned at all. Turn on the categories and all author settings.
5. View on the front-end and in the editor; confirm that your sponsored content is displaying the 'Sponsored' flag, logos and byline, using the custom values set on your testing site, and the underwritten/non-sponsored content doesn't show anything:

image

image

6. Navigate to the editor, and turn off the categories and authors for all the blocks; confirm that the author and category are removed for posts with underwritten content or no sponsors, but not the posts with native content:

image

7. If you haven't already, change the sponsored content strings (under Sponsors > Settings), and the Sponsored Content flag colour (under Customizer > Sponsor Settings); confirm the blocks are picking up both on the front-end and in the editor:

image

8. Cycle through the different themes and confirm that things look OK:

**Newspack Joseph:**

image

**Newspack Katharine:**

image

**Newspack Nelson:**

image

**Newspack Sacha:**

image

**Newspack Scott:**

image

### Other information:

### DIFF
--- a/class-newspack-blocks-api.php
+++ b/class-newspack-blocks-api.php
@@ -86,6 +86,21 @@ class Newspack_Blocks_API {
 				],
 			]
 		);
+
+		/* Sponsors */
+		register_rest_field(
+			'post',
+			'newspack_post_sponsors',
+			[
+				'get_callback' => [ 'Newspack_Blocks_API', 'newspack_blocks_sponsor_info' ],
+				'schema'       => [
+					'context' => [
+						'edit',
+					],
+					'type'    => 'array',
+				],
+			]
+		);
 	}
 
 	/**
@@ -259,6 +274,34 @@ class Newspack_Blocks_API {
 	 */
 	public static function newspack_blocks_get_cat_tag_classes( $object ) {
 		return Newspack_Blocks::get_term_classes( $object['id'] );
+	}
+
+	/**
+	 * Get all sponsor information for the rest field.
+	 *
+	 * @param array $object The object info.
+	 * @return array sponsor information.
+	 */
+	public static function newspack_blocks_sponsor_info( $object ) {
+		if ( Newspack_Blocks::get_post_sponsors( $object['id'] ) ) {
+			$sponsors = Newspack_Blocks::get_post_sponsors( $object['id'] );
+			foreach ( $sponsors as $sponsor ) {
+				$sponsor_logo   = Newspack_Blocks::get_sponsor_logo_sized( $sponsor['sponsor_id'] );
+				$sponsor_info[] = array(
+					'flag'          => Newspack_Blocks::get_sponsor_label( $sponsor['sponsor_id'] ),
+					'display_name'  => $sponsor['sponsor_name'],
+					'sponsor_url'   => $sponsor['sponsor_url'],
+					'byline_prefix' => $sponsor['sponsor_byline'],
+					'id'            => $sponsor['sponsor_id'],
+					'src'           => $sponsor_logo['src'],
+					'img_width'     => $sponsor_logo['img_width'],
+					'img_height'    => $sponsor_logo['img_height'],
+				);
+			}
+			return $sponsor_info;
+		} else {
+			return false;
+		}
 	}
 
 	/**
--- a/class-newspack-blocks-api.php
+++ b/class-newspack-blocks-api.php
@@ -289,10 +289,11 @@ class Newspack_Blocks_API {
 				$sponsor_logo   = Newspack_Blocks::get_sponsor_logo_sized( $sponsor['sponsor_id'] );
 				$sponsor_info[] = array(
 					'flag'          => Newspack_Blocks::get_sponsor_label( $sponsor['sponsor_id'] ),
-					'display_name'  => $sponsor['sponsor_name'],
+					'sponsor_name'  => $sponsor['sponsor_name'],
 					'sponsor_url'   => $sponsor['sponsor_url'],
 					'byline_prefix' => $sponsor['sponsor_byline'],
 					'id'            => $sponsor['sponsor_id'],
+					'scope'         => $sponsor['sponsor_scope'],
 					'src'           => $sponsor_logo['src'],
 					'img_width'     => $sponsor_logo['img_width'],
 					'img_height'    => $sponsor_logo['img_height'],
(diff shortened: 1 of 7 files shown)


class-newspack-blocks-api.php
```diff
@@ -86,6 +86,21 @@ public static function register_rest_fields() {
 				],
 			]
 		);
+
+		/* Sponsors */
+		register_rest_field(
```
dkoo, line 91: Would this be better off in the plugin? Hoping to simplify logic in the theme + blocks if we can keep all data handling in the plugin.

class-newspack-blocks.php
```diff
@@ -502,6 +502,145 @@ public static function get_patterns_for_post_type( $post_type = null ) {
...
+	 * @return array Array of sponsors.
+	 */
+	public static function get_post_sponsors( $post_id ) {
+		if ( function_exists( '\Newspack_Sponsors\get_sponsors_for_post' ) ) {
+			// Get all assigned sponsors.
+			$sponsors_all = \Newspack_Sponsors\get_sponsors_for_post( $post_id );
+
+			// Loop through sponsors and remove duplicates.
+			$sponsors   = array();
+			$duplicates = array();
```
dkoo, line 518: Hoping you can remove this logic once we take care of https://github.com/Automattic/newspack-sponsors/issues/15 (no action needed on this PR)

class-newspack-blocks-api.php
```diff
@@ -261,6 +276,35 @@ public static function newspack_blocks_get_cat_tag_classes( $object ) {
 		return Newspack_Blocks::get_term_classes( $object['id'] );
 	}
 
+	/**
+	 * Get all sponsor information for the rest field.
+	 *
+	 * @param array $object The object info.
+	 * @return array sponsor information.
+	 */
+	public static function newspack_blocks_sponsor_info( $object ) {
```
dkoo, line 285: Wonder if we can just modify the plugin helper output to give the theme + blocks the data you need without having to transform it? Reference: https://github.com/Automattic/newspack-sponsors/issues/14 (no action needed for this PR)

class-newspack-blocks.php
```diff
@@ -502,6 +502,145 @@ public static function get_patterns_for_post_type( $post_type = null ) {
...
+
+	/**
+	 * Returns scaled down logo sizes based on the provided width and height; this is necessary for AMP.
+	 *
+	 * @param string $sponsor_id Sponsor ID.
+	 * @param string $maxwidth Maximum logo width.
+	 * @param string $maxheight Maximum logo height.
+	 * @return array Array with image's src, width and height.
+	 */
+	public static function get_sponsor_logo_sized( $sponsor_id, $maxwidth = 80, $maxheight = 40 ) {
```
dkoo, line 622: Again, hoping you can remove this logic if we handle in the plugin. Reference: https://github.com/Automattic/newspack-sponsors/issues/14 (no action needed for this PR)

laurelfulford:
After spending _way_ too much time trying to figure out why I couldn't re-create the date and flag issues, I realized I had not pushed up my latest commits before creating this PR 🤦‍♀️ Both should now be resolved!

class-newspack-blocks-api.php
laurelfulford: The API bit is only used in the blocks plugin and not the theme; it may be better from a code perspective to move anyway, but if it's just to avoid duplication between the two, it can probably stay put :)

class-newspack-blocks.php
laurelfulford: I think it makes sense to move this to the plugin, too!

class-newspack-blocks-api.php
laurelfulford: I think my dealing with _most_ of the sponsor information one way, and then the image size another way in the blocks plugin definitely makes this clunkier than it needs to be!

class-newspack-blocks-api.php
dkoo: Sounds good!

class-newspack-blocks-api.php
dkoo: Ok, I've created an issue to come back to this once we refactor the plugin: https://github.com/Automattic/newspack-blocks/issues/569

class-newspack-blocks-api.php
```diff
@@ -261,6 +276,35 @@ public static function newspack_blocks_get_cat_tag_classes( $object ) {
...
+		if ( Newspack_Blocks::get_post_sponsors( $object['id'] ) ) {
+			$sponsors = Newspack_Blocks::get_post_sponsors( $object['id'] );
+			foreach ( $sponsors as $sponsor ) {
+				$sponsor_logo   = Newspack_Blocks::get_sponsor_logo_sized( $sponsor['sponsor_id'] );
+				$sponsor_info[] = array(
+					'flag'          => Newspack_Blocks::get_sponsor_label( $sponsor['sponsor_id'] ),
```
dkoo, line 291:
```suggestion
					'flag'          => $sponsor['sponsor_flag'],
```

laurelfulford: Thanks Derrick! Updated in 2ae5c35  - honestly have no idea why the original was working for me, but this does, too :)

laurelfulford:
Thanks!

Issue #495, opened 2020-05-26
Blog Posts Block: Add option for manual Continue Reading link


tijosh:
**Is your feature request related to a problem? Please describe.**

Users cannot currently:

1. Display a Continue Reading link and
2. Control what comes before the link

If you set a manual excerpt or put a More block into the posts, no `Continue Reading` link is shown. 

**Describe the solution you'd like**

Add option to manually include a Continue Reading button

**Context**

#21404141-hc

laurelfulford:
Just noting for anyone looking into this:

The 'Continue Reading' link will be coming from the theme, rather than the block. There are a number of themes that replace the `[...]` at the end of automatically generated excerpts with something else by filtering `excerpt_more` ([example](https://github.com/Automattic/themes/blob/b6ae2c8a45c2317cef960b8c0ef94da55427ffa5/textbook/inc/template-tags.php#L181-L201)). The reason the link doesn't happen with Custom Excerpts is that the `[...]` is not there, since it's assumed that those would end with a full stop and not need an ellipsis. 

If we added this enhancement, we might want this to fully replace the `[...]` for automatically generated excerpts (so theme filters won't interfere with it), and have the option to always include a customizable link at the end of the excerpt to the full post.

tijosh:
> and have the option to always include a customizable link at the end of the excerpt to the full post.

I think this would be helpful. It allows consistency no matter how the posts are constructed.

dcoleonline:
Also requested in 3264365-zen, so if this is ever added, we should notify them.

mikeicode:
Requested 26351581-hc

Issue #420, opened 2020-03-22
Blog Post Block Missing Sharing Icons


arcangelini:
**Describe the bug**
When using the Blog Posts block with the sharing icons enabled in Jetpack the logos do not load on the posts. This appears to specifically happen when you *disable* sharing icons on **pages**, but enable them on **posts**.

**To Reproduce**
1. Enable Jetpack sharing buttons and set to "Icons Only" 
2. Adjust the setting to Allow on "Posts" only
3. Add a Blog Post block on any page

**Screenshots**
[![Screenshot](https://d.pr/i/Hl06Sn+)](https://d.pr/i/Hl06Sn)

**Expected behavior**
The sharing buttons should display logos inside the circle

[![Screenshot](https://d.pr/i/uxijGP+)](https://d.pr/i/uxijGP)

#2809894-zen

ccwalburn:
Another user report here: 3170967-zen

liviopv:
Another case #22911851-hc
Slack: https://a8c.slack.com/archives/C03TY6J1A/p1595763616100300

donalirl:
Another case in #3253775-zen

donalirl:
It seems that you specifically need to enable sharing buttons on Pages under Marketing → Sharing Buttons. Without that, the Blog Posts block will not show the sharing button icon on the *page*, but it will still show it on the individual *posts*.

nagpai:
Another issue #24191494-hc

@katiebethbrown  helped me with this. The sharing buttons appeared correctly after Marketing >> Sharing buttom settings were enabled correctly. If it is disabled then the buttons should not show at all instead of erroneous buttons, or the block settings should override the marketing settings.

eduardozulian:
Another similar report in 24960692-hc  However, in this case, we have an Elementor posts widget. I believe that the issue is actually with Jetpack: once the sharing is disabled for pages, it treats the sharing buttons inside post listings the same way:

<img width="1215" alt="Captura de Tela 2020-10-30 às 17 31 23" src="https://user-images.githubusercontent.com/1136160/97754161-c5927c80-1ad5-11eb-9469-79dc14b7406f.png">


cc @jeherve

jeherve:
> once the sharing is disabled for pages, it treats the sharing buttons inside post listings the same way

That is correct. It was built that way on purpose. 

In the past, we've recommended to folks that they remove the sharing markup from those areas if they don't intend to show sharing buttons, like it's done here for example:
[...]

PR #605, opened 2020-10-21
Homepage Posts: Add option to change excerpt length

### LINKED ISSUE #22: Homepage blocks: Add ability to adjust excerpt length
Right now, the blocks just use whatever the default is for the current site (55 words). It would be great to add a control, so the automatic excerpt could be adjusted by block, depending where in the layout it sits. 

This could be done via a slider with a numeric value field, similar to the number of posts.

laurelfulford:
### All Submissions:

This PR adds an option to the Homepage Posts block to control the excerpt length. It uses the WordPress default (55 words) as the default, and allows you to increase the length up to 100 words, or down to 10 words. 

It appears to work well on Newspack sites but could break some things that currently work on WP.com:

The 'excerpt' is no longer an excerpt, so it will no longer pick up filters for `the_excerpt`. This won't be an issue on Newspack sites, but on WP.com, some themes -- like Varia -- filter `the_excerpt` to replace the `[...]` with a "Continue Reading" link. 

Similarily, things like the share and like buttons can be made to appear with certain settings using themes other than the Newspack theme (since it removes it from the bottom of the content/excerpts).

If this approach otherwise makes sense, adding an option for a Continue Reading link in the block (https://github.com/Automattic/newspack-blocks/issues/495) could be a way to make it behave more consistently and as an alternative for sites that may like this behaviour. 

I'm not sure what we can do about the share buttons; they also don't work properly with some settings (like if you have them set for posts and not pages, the buttons will appear, but the icons will not -- https://github.com/Automattic/newspack-blocks/issues/420).

It's possible there's another approach that will maintain the excerpts (filtering the excerpt length on a per-block basis seemed like a bad idea, but maybe it's better than this?), or ways to work around the above (only making the option available when a theme opts in, similar to the Article Subtitle?). I'd be very interested to hear other's thoughts about that!

Closes #22.

1. Make sure you have a few recent posts with the following settings: a regular post; a very short (< 10 words) post, a post with a longer custom excerpt, and a post that has the post format set to aside.
2. Apply the PR and run `npm run build`.
3. Create/edit a page and add the Homepage Posts block. 
4. With the block selected, confirm that there is now a range slider in the right column, under the Post Control Settings, with a default value of 55:

image

5. Toggle off the Show Excerpt option and confirm that the range slider disappears, too.
6. Toggle Show Excerpt back on again.
7. Adjust the excerpt length up and down. Confirm that only the posts with the dynamically generated excerpt are affected, and that the excerpt length changes in the preview. **Note:** By default in WordPress, Custom Excerpts are displayed in full, and we've designed the Aside post format to display the full-content in the blocks, too. Because of this, I opted to only have this slider affect the dynamically generated excerpts, but am open to feedback on that!
8. Publish the post.
9. Confirm that your custom excerpt length is also reflected on the front end.
10. Try steps 7-9 with both longer, and shorter excerpts, to confirm it works as expected.

image

### Other information:

### DIFF
--- a/class-newspack-blocks-api.php
+++ b/class-newspack-blocks-api.php
@@ -360,7 +360,7 @@ class Newspack_Blocks_API {
 	 * Pass whether there is a custom excerpt to the editor.
 	 *
 	 * @param array $object The object info.
-	 * @return string post format.
+	 * @return boolean custom excerpt status.
 	 */
 	public static function newspack_blocks_has_custom_excerpt( $object ) {
 		$post_has_custom_excerpt = has_excerpt( $object['id'] );
--- a/class-newspack-blocks-api.php
+++ b/class-newspack-blocks-api.php
@@ -116,6 +116,21 @@ class Newspack_Blocks_API {
 				],
 			]
 		);
+
+		/* Has custom excerpt */
+		register_rest_field(
+			'post',
+			'newspack_has_custom_excerpt',
+			[
+				'get_callback' => [ 'Newspack_Blocks_API', 'newspack_blocks_has_custom_excerpt' ],
+				'schema'       => [
+					'context' => [
+						'edit',
+					],
+					'type'    => 'boolean',
+				],
+			]
+		);
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


dkoo (approved):
@laurelfulford This treatment works for me. I'll approve for now but will re-review if you want to try the `excerpt_length` filter instead.

laurelfulford:
Just adding a note that this may be replaced with #607, so it shouldn't be merged unless that one doesn't seem like a suitable approach.

laurelfulford:
#607 will replace this PR, so closing it for now!

PR #607, opened 2020-10-22
Homepage Posts: Add option to change excerpt length

laurelfulford:
### All Submissions:

This is a rework of #605 with the same testing steps.

It replaces the front-end approach to truncating the excerpts with a filter, so they will still be normal excerpts that pick up other filters. 

The one remaining issue is that the_excerpt is no longer filtered in the editor, so theme's 'Continue Reading' and social links will no longer show up there. I think this is an acceptable loss, between the 'Continue Reading' styles not usually being included in the editor styles, and the social links not displaying correctly anyway.

If it comes up as an issue, we can always rework it there, too!

1. See #605.

### Other information:

### DIFF
--- a/class-newspack-blocks-api.php
+++ b/class-newspack-blocks-api.php
@@ -360,7 +360,7 @@ class Newspack_Blocks_API {
 	 * Pass whether there is a custom excerpt to the editor.
 	 *
 	 * @param array $object The object info.
-	 * @return string post format.
+	 * @return boolean custom excerpt status.
 	 */
 	public static function newspack_blocks_has_custom_excerpt( $object ) {
 		$post_has_custom_excerpt = has_excerpt( $object['id'] );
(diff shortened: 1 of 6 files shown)


dkoo:
Adding another reviewer since I worked on the code here, too.

laurelfulford:
Ah, good call -- thanks @dkoo!

claudiulodro (approved):
This approach works well, and the code looks good. 👍 

> Note: By default in WordPress, Custom Excerpts are displayed in full, and we've designed the Aside post format to display the full-content in the blocks, too. Because of this, I opted to only have this slider affect the dynamically generated excerpts, but am open to feedback on that!

I noticed this in my testing, and I think it makes sense. If a bunch of people complain, we can always change the behavior.

laurelfulford:
Thanks Claudiu!

PR #628, opened 2020-11-13
Proposed fix for #404

### LINKED ISSUE #404: Blog Posts Block: Load More Posts button doesn't appear even though there are more posts to load
**Describe the bug**
User is using a blog post block to display posts to 5 posts at a time. The Load More button does appear as expected for the next 5 posts, but it doesn't seem to appear after the 2nd time even though there are more posts.

When I tested this on my own site, at a certain point when trying to click on the Load More button for more posts, the Load More button will stop appearing, even though there are still more posts that can load.

**Screenshare/Screenshot**

https://d.pr/i/AjoWdA 
https://d.pr/i/ADm2an

**To Reproduce**
Steps to reproduce the behavior: I tested this on my own site: https://susanjane.blog/testing-blog-posts-block/
1. Click on `Load More Posts` to load next 5 posts, then again, and then again.
2. The last post at the bottom should be `BIRTHDAY BOOK CLUB` which is not the last post published (there are more before this). You won't see the `Load More Posts` button, however, there are 12 published posts before this post, so the `Load More Posts` button should appear to display more posts.

**Expected behavior**
The Load More Posts button should appear at the bottom to load the next 5 posts until there are no more published posts to display. For Reference, on my site, the last published post should be titled `The Coolest Hair`

soatok:
### All Submissions:

Closes #404 .

1. Create a blog with more than 30 test posts, with 10 displayed in the list block.
2. Verify that you can press the button more than twice without the block disappearing.

### Other information:

### DIFF
--- a/amp/homepage-articles/view.js
+++ b/amp/homepage-articles/view.js
@@ -17,6 +17,7 @@ function buildLoadMoreHandler( blockWrapperEl ) {
 		AMP.getState( 'newspackHomepagePosts.exclude_ids' ).then( function( exclude_ids ) {
 			const requestURL = new URL( btnEl.getAttribute( 'data-next' ) );
 			requestURL.searchParams.set( 'exclude_ids', JSON.parse( exclude_ids ).join( ',' ) );
+			requestURL.searchParams.delete( 'page' ); // If we're using exlcude_ids, the page parameter isn't helpful. See issue #404
 			apiFetchWithRetry( { url: requestURL.toString(), onSuccess, onError }, 3 );
 		} );
 		function onSuccess( data ) {


dkoo:
@soatok Thanks for the proposed fix—we'll review and test this shortly.

dkoo (changes requested):
@soatok Thanks for the proposed fix! I believe you've correctly identified the cause of the issue here, but there are a few things we'd want to implement in a fix:

1. We'd need to implement the fix in both AMP JS (the file you've edited here) and non-AMP JS (`/src/blocks/homepage-articles/view.js`).
2. We'd like to refactor the AMP JS to avoid the use of `new URL`, which is not supported by IE11. This would be an improvement over our current code, which uses the `URL` class in AMP JS while relying on plain strings in non-AMP JS.
3. We'd need to ensure that the custom REST route still works correctly if not passed the `exclude_ids` param, or if the param is an empty array. So this might mean selectively retaining the `page` param and/or fixing its implementation on the PHP side.

Would you like to try to implement the above? If not, we're happy take over this fix (with props to you for identifying the solution). Thanks again!

soatok:
> Would you like to try to implement the above? If not, we're happy take over this fix (with props to you for identifying the solution). Thanks again!

Unfortunately I'm a bit triple-booked at work right now. Please feel free to take over this fix. Thanks!

dkoo:
Rather than messing with the fork, I'm closing this PR in favor of #672.

PR #649, opened 2020-12-11
fix(api): fallback to standard if post_format is false

david-szabo97:
### All Submissions:

Make sure `newspack_post_format` REST API field fallbacks to `standard` value if the post format is `false`. This change is for consistency. Right now we are returning `false` which doesn't comply with the schema we use. We specified `string` as the field type. So returning a boolean doesn't make sense.

It also fixes an issue with updating posts via REST API. If you are reusing the post object returned by the GET WP REST API for a PUT request, then you are going to get an error: `newspack_post_format is not of type string`. Since we received `false` as a value for `newspack_post_format` which isn't a string so it doesn't comply with our schema.

1. `GET /wp/v2/posts/`
2. Make sure body contains `newspack_post_format` and it isn't `false`

### Other information:

### DIFF
--- a/class-newspack-blocks-api.php
+++ b/class-newspack-blocks-api.php
@@ -353,7 +353,7 @@ class Newspack_Blocks_API {
 	 */
 	public static function newspack_blocks_post_format( $object ) {
 		$post_format = get_post_format( $object['id'] );
-		return $post_format;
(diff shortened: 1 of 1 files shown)


dkoo (approved):
@david-szabo97 Thanks for the fix—tested and confirmed.

PR #656, opened 2020-12-17
Homepage Posts block: add optional 'continue reading' link underneath excerpts

### LINKED ISSUE #495: Blog Posts Block: Add option for manual Continue Reading link

laurelfulford:
### All Submissions:

This PR adds an option to add a customize-able 'Read More' link for each story on a per-post basis.

Unfortunately, current behaviour in WP.com themes could complicate this. This is a comment from me from the [original enhancement request](https://github.com/Automattic/newspack-blocks/issues/495):

> There are a number of themes that replace the [...] at the end of automatically generated excerpts with something else by filtering excerpt_more ([example](https://github.com/Automattic/themes/blob/b6ae2c8a45c2317cef960b8c0ef94da55427ffa5/textbook/inc/template-tags.php#L181-L201)). The reason the link doesn't happen with Custom Excerpts is that the [...] is not there, since it's assumed that those would end with a full stop and not need an ellipsis.

To work around this, I added a filter that should override any `entry_more` customization in the theme, switching it to `...` when the 'Read More' link is turned on, then removing the filter so it falls back to whatever a theme might be doing for blocks that have it left off. 

Closes #495

Note: I've added the 'Don't Merge' tag so some theme styles can be added for this as well (there is a bit adding an arrow already, but it could use refining); in the meantime the block uses the comment class name `more-link` on the link, so it should pick up styles from most WordPress themes that it's already getting links from.

1. Apply the PR and run `npm run build`
2. Add a homepage posts block to the editor; in the Post Control settings section, confirm there is an option to enable a 'Read More' link that is off by default. Toggle it on and confirm a field to change the read more text appears below it, with the default text "Keep reading":

image

3. Add 4 Homepage posts block to the editor with different configurations including:
* A block with the excerpt on and the read more link on; change the read more text.
* A block with the excerpt on, a custom excerpt length, and the read more link on.
* A block with no excerpt, and the read more link on; change the read more link text.
* A block with the excerpt on, and the read more link off. 
4. Confirm on the front end and in the editor, your settings are respected, including the different read more link text, and, when enabled, that the default `[...]` is replaced with `...`.
5. Note: on the front-end, you can see that the `.more-link` already has styles in the theme, adding an arrow; that will be updated:
image

6. Download Varia from WordPress.com and enable it on your test site. 
7. View on the front end and confirm that Varia's custom `excerpt_more` filter ("Continue Reading") is being picked up on blocks that don't have the 'Read More' link set, but the ones with a 'Read More' link set should still be applying it. All of the links should be applying the arrow from Varia, since it also uses the `.more-link` CSS class. (Note: Varia's custom filter will not be applied in the editor, since the theme only applies it to the front-end; in that case, the blocks without the 'Read More' link enabled should still fall back to the default `[...]`):

image

8. Enable the 'Load More Posts' option on the blocks, and confirm that the Read More options are being applied correctly to posts that are newly loaded, both with and without AMP enabled.
9. On the front-end, click on the Read More links to confirm they go to the correct posts. 

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/class-wp-rest-newspack-articles-controller.php
+++ b/src/blocks/homepage-articles/class-wp-rest-newspack-articles-controller.php
@@ -80,6 +80,7 @@ class WP_REST_Newspack_Articles_Controller extends WP_REST_Controller {
 		$next_url = '';
 
 		Newspack_Blocks::filter_excerpt_length( $attributes );
+		Newspack_Blocks::filter_excerpt_ellipses( $attributes );
 
 		// The Loop.
 		while ( $article_query->have_posts() ) {
@@ -99,6 +100,7 @@ class WP_REST_Newspack_Articles_Controller extends WP_REST_Controller {
 		}
 
 		Newspack_Blocks::remove_excerpt_length_filter();
+		Newspack_Blocks::remove_excerpt_ellipsis_filter();
 
 		// Provide next URL if there are more pages.
 		if ( $next_page <= $article_query->max_num_pages ) {
--- a/src/blocks/homepage-articles/class-wp-rest-newspack-articles-controller.php
+++ b/src/blocks/homepage-articles/class-wp-rest-newspack-articles-controller.php
@@ -80,7 +80,7 @@ class WP_REST_Newspack_Articles_Controller extends WP_REST_Controller {
 		$next_url = '';
 
 		Newspack_Blocks::filter_excerpt_length( $attributes );
-		Newspack_Blocks::filter_excerpt_ellipses( $attributes );
+		Newspack_Blocks::filter_excerpt_more( $attributes );
 
 		// The Loop.
 		while ( $article_query->have_posts() ) {
@@ -100,7 +100,7 @@ class WP_REST_Newspack_Articles_Controller extends WP_REST_Controller {
 		}
 
 		Newspack_Blocks::remove_excerpt_length_filter();
-		Newspack_Blocks::remove_excerpt_ellipsis_filter();
+		Newspack_Blocks::remove_excerpt_more_filter();
 
 		// Provide next URL if there are more pages.
 		if ( $next_page <= $article_query->max_num_pages ) {
--- a/class-newspack-blocks.php
+++ b/class-newspack-blocks.php
@@ -676,5 +676,33 @@ class Newspack_Blocks {
 			);
 		}
 	}
+
+	/**
+	 * Return a excerpt more replacement when using the 'Read More' link.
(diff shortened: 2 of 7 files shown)


src/blocks/homepage-articles/class-wp-rest-newspack-articles-controller.php
```diff
@@ -80,6 +80,7 @@ public function get_items( $request ) {
 		$next_url = '';
 
 		Newspack_Blocks::filter_excerpt_length( $attributes );
+		Newspack_Blocks::filter_excerpt_ellipses( $attributes );
```
dkoo, line 83: Looks like this doesn't match the method name in `class-newspack-blocks.php`? This works for me:

```suggestion
		Newspack_Blocks::filter_excerpt_more( $attributes );
```

src/blocks/homepage-articles/class-wp-rest-newspack-articles-controller.php
```diff
@@ -99,6 +100,7 @@ public function get_items( $request ) {
 		}
 
 		Newspack_Blocks::remove_excerpt_length_filter();
+		Newspack_Blocks::remove_excerpt_ellipsis_filter();
```
dkoo, line 103: Same issue, this doesn't match the method name in `class-newspack-blocks.php`.

```suggestion
		Newspack_Blocks::remove_excerpt_more_filter();
```

dkoo (changes requested):
Works great, except when using the "Load More" button I'm seeing 500 fatal errors like this:

```
PHP Fatal error:  Uncaught Error: Call to undefined method Newspack_Blocks::remove_excerpt_ellipsis_filter() in /Users/dkoo/Local Sites/newspack/app/public/wp-content/plugins/newspack-blocks/src/blocks/homepage-articles/class-wp-rest-newspack-articles-controller.php:103
```

I think we just need to update some method names in the API controller as suggested below. Otherwise, looks good!

laurelfulford:
Thanks @dkoo! 

I ran into this exact issue while running though the texting steps for the PR yesterday, committed a fix in 430ffcc... and then forgot to push it up 🤦‍♀️ That update is now actually on GitHub, and not just hanging around on my local copy of the branch!

dkoo (approved):
Looks good now! :shipit:

PR #670, opened 2021-01-15
Fix duplicate category names

### LINKED ISSUE #669: Subcategories cannot be accurately selected
**Describe the bug**
There are two elements to this: a UI bug and a data bug. This may also be related to #207.

First of all, when attempting to select a subcategory as a filter, if there is more than one subcategory with the same name it's not possible to distinguish between them, as the UI simply lists the sub-category name without the parent as context, like so:

Screenshot from 2021-01-15 12-00-10

Secondly, which ever of those subcategories are selected, only one of them is ever actually used in the filter, which means some of the subcategories can never be selected for the filter.

**To Reproduce**
1. Create a sub-category structure. Something like this (example IDs in brackets):

Category | Subcategory
--- | ---
Entertainment (4) | Feature (18)
Health (13) | Feature (19)
Politics (10) | Feature (20)

2. In a page, add three sets of homepage posts blocks and add filters as follows:

Block | Category | Subcategory
--- | --- | ---
1 | Entertainment (4) | Feature (18)
2 | Health (13) | Feature (19)
3 | Politics (10) | Feature (20)

Given the UI issue, you'll have to guess the right sub-category by considering the ordering of how they display in the Posts > Categories screen.

3. Review the code and observe that in each block the second category ID is always "18", when they should be "18", "19" and "20". See example:

```<!-- wp:columns -->
<div class="wp-block-columns"><!-- wp:column -->
<div class="wp-block-column"><!-- wp:newspack-blocks/homepage-articles {"postsToShow":1,"categories":["4","18"]} /--></div>
<!-- /wp:column -->

<!-- wp:column -->
<div class="wp-block-column"><!-- wp:newspack-blocks/homepage-articles {"postsToShow":1,"categories":["13","18"]} /--></div>
<!-- /wp:column -->

<!-- wp:column -->
<div class="wp-block-column"><!-- wp:newspack-blocks/homepage-articles {"postsToShow":1,"categories":["10","18"]} /--></div>
<!-- /wp:column --></div>
<!-- /wp:columns -->
```

**Expected behavior**
1. When categories with the same name are present it should be possible to distinguish between them in the UI based on context (such as parent/child relationship)
2. When selecting categories that have the same name, the correct category ID should be added to the block params JSON.

adekbadek:
### All Submissions:

Closes #669.

1. Observe #669 is not reproducible
2. Create four categories named "Foo", assign different parrent to two of them
3. Insert a Homepage Posts block and use the category filter in the sidebar, insert "Foo"
4. Observe that the four are present in the suggestions: the ones with the parents have `Foo – <parent>` as label, while the duplicated parent-less one has `Foo (<id>)` as label

### Other information:

### DIFF
--- a/src/components/autocomplete-tokenfield.js
+++ b/src/components/autocomplete-tokenfield.js
@@ -119,6 +119,10 @@ class AutocompleteTokenField extends Component {
 					const currentSuggestions = [];
 
 					suggestions.forEach( suggestion => {
+						const duplicatedSuggestionIndex = currentSuggestions.indexOf( suggestion.label );
+						if ( duplicatedSuggestionIndex >= 0 ) {
+							suggestion.label = `${ suggestion.label } (${ suggestion.value })`;
+						}
 						currentSuggestions.push( suggestion.label );
 						validValues[ suggestion.value ] = suggestion.label;
 					} );
--- a/src/components/query-controls.js
+++ b/src/components/query-controls.js
@@ -17,6 +17,9 @@ import { decodeEntities } from '@wordpress/html-entities';
  */
 import AutocompleteTokenField from './autocomplete-tokenfield';
 
+const getCategoryTitle = category =>
+	decodeEntities( category.name ) || __( '(no title)', 'newspack-blocks' );
+
 class QueryControls extends Component {
 	state = {
 		showAdvancedFilters: false,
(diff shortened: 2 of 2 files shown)


philipjohn (approved):
Yep, all good - this works nicely, thanks Adam!

PR #677, opened 2021-01-22
refactor: use rest_url to get the /posts route, for Toolkit

dkoo:
### All Submissions:

WP.com Toolkit (formerly Full Site Editing) performs some kind of server-side transformation of REST API endpoints. This refactor will ensure that the path used for the new `/posts` endpoint comes from the server with the required transformations, rather than being hard-coded in the JS.

This is a pure refactor and shouldn't cause any changes in behavior when the plugin is run on a non-WP.com site.

Related to #404 

1. Create or edit a page and add a Homepage Posts block. 
2. Verify that the block loads posts in the editor as expected, with no change in behavior.

### Other information:

### DIFF
--- a/class-newspack-blocks-api.php
+++ b/class-newspack-blocks-api.php
@@ -390,7 +390,7 @@ class Newspack_Blocks_API {
 	public static function register_posts_endpoint() {
 		register_rest_route(
 			'newspack-blocks/v1',
-			'/homepage-article-posts',
+			'/newspack-blocks-posts',
 			[
 				'methods'             => \WP_REST_Server::READABLE,
 				'callback'            => [ 'Newspack_Blocks_API', 'posts_endpoint' ],
--- a/class-newspack-blocks-api.php
+++ b/class-newspack-blocks-api.php
@@ -385,80 +385,6 @@ class Newspack_Blocks_API {
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


class-newspack-blocks-api.php
```diff
@@ -390,7 +390,7 @@ public static function register_video_playlist_endpoint() {
 	public static function register_posts_endpoint() {
 		register_rest_route(
 			'newspack-blocks/v1',
-			'/posts',
+			'/homepage-article-posts',
```
adekbadek, line 393: With the plan to use this endpoint for the Posts Carousel block, this name change might become misleading. Why not leave it as it was?

dkoo: Good call. Renamed to `/newspack-blocks-posts` in 3982fd2

dkoo:
Adding a Do Not Merge! label because renaming the route needs to be tested to see if it solves a bug downstream.

adekbadek (approved):

dkoo:
Tested the downstream application and 95fcb09 should fix the problem there. Again, no changes in behavior on a standalone plugin installation.

dkoo:
@adekbadek I had to push a new commit here: 95fcb09

Can I get a re-review?

adekbadek (approved):

PR #661, opened 2021-01-05
feat: custom endpoint for homepage posts in editor

jeffersonrabb:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-theme/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Replaces core WordPress `/wp/v2/posts` with a custom endpoint. This is done in preparation for CPT support in Homepage Posts block. Core v2 endpoints are limited to a single post type, so to enable a mixed group of CPTs it is necessary to switch to a custom endpoint. This PR is a pure refactor and should work exactly like `master`. Once this is merged, a follow-up PR will add support for CPT selection. This PR also replaces the specific posts look up with a custom API which was previously considered experimental. The custom API should be much faster and more efficient. 

### How to test the changes in this Pull Request:

1. Edit a page or post with many Homepage Posts blocks. 
2. Open Dev Tools->Network->XHR and verify all API requests go to `/wp-json/newspack-blocks/v1/posts`. On `master` the same requests should go to `wp-json/wp/v2/posts`.
3. In one Homepage Posts Block switch on "Choose specific posts," then type something in the search field. On `master` the API call should go to `/wp-json/wp/v2/search`. On this branch, verify the API call goes to `/wp-json/newspack-blocks/v1/specific-posts`
4. There should be no functional changes in this branch. Verify all behaviors of the block work exactly as they do on `master`. 
5. These changes affect the editor environment only. There should be no need to test anything on the front end.
6. As a sanity check, verify the Posts Carousel block continues to work as it does on `master`.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #681, opened 2021-01-27
fix: in specific posts mode, post order should match input order

dkoo:
### All Submissions:

Fixes a bug reported by a Newspack publisher:

> We had this come in from the editorial team and am curious if anything changed in how the post order is displayed for editors. For context, here is the editor's note " The order of stories displayed in the three-story row when editing the home page no longer reflects the list of post order in the editing window on the rail, or on the actual page. (fortunately, the order in the editing window matches the actual page, but until now changing the order on editing window also changed the preview display on the backend.)"

To summarize: since #661, when editing a block in Specific Posts mode, the posts in the editor preview are always shown in descending order of publish date (same as in query mode), which doesn't match the front-end display. Posts in the front-end are shown in the order they appear in the Specific Posts autocomplete field.

Note: I'm removing the `orderby` param from the `/posts` endpoint callback, since we're no longer using it anywhere at the moment, but we can always add it back if we find a need to pass that arg in from the JS.

1. On `master`, create a Homepage Posts block and enable Specific Posts mode. Add several posts, ensuring that you add them in an order that won't match their publish date. Observe that the posts in the editor preview appear in descending order of publish date.
2. Publish the page and view the front-end, and observe that the order of the posts in the front-end matches the order you entered in them into the autocomplete field.
3. Check out this branch and run `npm run build`.
4. Refresh the page editor and confirm that the editor preview now shows the posts in the same order as both the autocomplete field and the front-end.

### Other information:

### DIFF
--- a/class-newspack-blocks-api.php
+++ b/class-newspack-blocks-api.php
@@ -423,9 +423,6 @@ class Newspack_Blocks_API {
 						),
 						'default' => array(),
 					],
-					'orderby'      => [
-						'sanitize_callback' => 'sanitize_text_field',
-					],
 					'per_page'     => [
 						'sanitize_callback' => 'absint',
 					],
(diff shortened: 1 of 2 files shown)


jeffersonrabb (approved):
Nice fix!

PR #689, opened 2021-02-04
refactor: get specific posts REST endpoint from localized value

### LINKED ISSUE #685: Blog Posts Block: Hide "Load More" option in editor when using Specific Posts mode
Related to https://github.com/Automattic/wp-calypso/issues/47086

The above issue is no longer replicable after #640, because when the block is in Specific Posts mode, we hide the "max number of items" option and simply get all of the specified posts up to 100 max, rather paginating the results. This means the "Load More" button will never be visible on the front-end.

Therefore, for Specific Posts blocks, we should also hide the option to show the load more button at all, and not show it in the editor preview render, to match the front-end display.

dkoo:
### All Submissions:

The same fix as #677, but for `specific-posts`.

Refactors the `newspack-blocks/v1/specific-posts` custom endpoint so that it is a.) retrieved from a localized value, so that it gets run through the `rest_url` filter, and b.) moved to the Homepage Posts REST controller, so that the downstream application knows to pick it up.

Also fixes #685.

#### Endpoint test

This is a pure refactor and shouldn't affect behavior in a non-WP.com environment. To test, create a block in Specific Posts mode and ensure that the posts load as expected in both the editor and on the front-end.

#### "Load More" test

1. Create a new Homepage Posts block and enable "specific posts" mode.
2. Confirm that the `Show "Load more posts" Button` is hidden when specific posts mode is on, and shown when it's off.

### Other information:

### DIFF
--- a/class-newspack-blocks-api.php
+++ b/class-newspack-blocks-api.php
@@ -385,38 +385,6 @@ class Newspack_Blocks_API {
 	}
 
 	/**
-	 * Register specific posts endpoint.
-	 */
-	public static function register_post_lookup_endpoint() {
-		register_rest_route(
-			'newspack-blocks/v1',
-			'/specific-posts',
-			[
-				'methods'             => \WP_REST_Server::READABLE,
-				'callback'            => [ 'Newspack_Blocks_API', 'specific_posts_endpoint' ],
-				'args'                => [
-					'search'   => [
-						'sanitize_callback' => 'sanitize_text_field',
-					],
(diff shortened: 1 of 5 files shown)


jeffersonrabb (approved):

dkoo:
@jeffersonrabb could use another quick review. Also tacking on another fix for a bug that was reported by the .com folks.

jeffersonrabb (approved):
🚢

Issue #700, opened 2021-02-16
Homepage Posts: excerpt in editor displaying extraneous content


creativecoder:
**Describe the bug**

The excerpt displayed in the editor for the Homepage Posts block is showing extraneous information that is normally filtered out using the standard WordPress functions (like `get_the_excerpt`).

It looks like this was introduced by the changes in #607...  See this note in the description of that PR:

> The one remaining issue is that the_excerpt is no longer filtered in the editor

Any PHP rendered block that uses "fallback content" in the save function will show that content as part of the excerpt in the editor. (Fallback content is needed for contexts where the post content is shown without executing PHP, like sending new posts via email.)

**To Reproduce**

1. Create a post and add a block that uses "fallback" content at the beginning of the post (2 example blocks in Jetpack are the Podcast Player or Eventbrite Checkout block)
2. Use the Homepage Posts block on a page, that displays in the post you just created, including the excerpt
3. In the editor, see that the fallback content displays in the preview of the excerpt

**Screenshots**

| Editor | Front-end |
| - | - |
| <img width="600" alt="Screen Shot 2021-02-15 at 17 53 24" src="https://user-images.githubusercontent.com/1699996/108003073-2b3c4300-6fb7-11eb-947f-fa8651872e2c.png"> | <img width="600" alt="Screen Shot 2021-02-15 at 17 54 13" src="https://user-images.githubusercontent.com/1699996/108003074-2bd4d980-6fb7-11eb-9d88-f1112605efb0.png"> |

It looks quite odd when you're previewing several posts, for example

![image](https://user-images.githubusercontent.com/1699996/108003253-cc2afe00-6fb7-11eb-8761-d9e09478f5e1.png)

**Expected behavior**

The excerpt displayed in the editor matches what's on the front-end.

cc @laurelfulford

PR #701, opened 2021-02-16
refactor: render custom excerpt length server-side

jeffersonrabb:
### All Submissions:

The custom excerpt length control introduced in https://github.com/Automattic/newspack-blocks/pull/607 had one drawback, which is that `the_excerpt is no longer filtered in the editor.` This has come up in https://github.com/Automattic/newspack-blocks/issues/700 as a problem for blocks with fallback content. The custom posts endpoint introduced in https://github.com/Automattic/newspack-blocks/pull/661 opens up a cleaner way to deal with this problem. In this PR, the block excerpt length and visibility attributes are passed to the endpoint and the endpoint uses excerpt filtering identical to the block's front-end rendering. This should also mean the truncated excerpt is _slightly_ more WYSIWYG since the editor and front-end are now using the exact same approach.

Closes https://github.com/Automattic/newspack-blocks/issues/700

1. Add a few Homepage Posts blocks to a post, toggle on `Show Excerpt` on some of them, and try a variety of `Max number of words in excerpt` settings.
2. Verify the results seen in the editor match the results seen in front end.
3. On `master`, replicate the issue described in https://github.com/Automattic/newspack-blocks/issues/700 using a block with fallback content (Podcast Player or Eventbrite Checkout). 
4. Switch to this branch, verify the issue is resolved.

### Other information:

### DIFF
--- a/class-newspack-blocks-api.php
+++ b/class-newspack-blocks-api.php
@@ -438,6 +438,12 @@ class Newspack_Blocks_API {
 			$args['post_type'] = $params['post_type'];
 		}
 
+		$block_attributes = [
+			'showExcerpt'   => $params['show_excerpt'],
+			'excerptLength' => $params['excerpt_length'],
+		];
+		Newspack_Blocks::filter_excerpt_length( $block_attributes );
+
 		$query        = new WP_Query();
 		$query_result = $query->query( $args );
 		$posts        = [];
@@ -486,6 +492,9 @@ class Newspack_Blocks_API {
 
 			$posts[] = array_merge( $data, $add_ons );
 		}
+
+		Newspack_Blocks::remove_excerpt_length_filter();
+
 		return new \WP_REST_Response( $posts );
 	}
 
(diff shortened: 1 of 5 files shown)


creativecoder (commented):
@jeffersonrabb Thank you for the quick fix.

✅ This change fixes #700 and I seems like a much cleaner way to display the excerpt in the editor!

One more improvement I think would be good before merging this is debouncing the excerpt length change. As is, a lot of requests can be triggered while adjusting the range slider:

newspack-post-requests-when-changing-excerpt-length

adekbadek:
> One more improvement I think would be good before merging this is debouncing the excerpt length change. As is, a lot of requests can be triggered while adjusting the range slider

Good catch! It appears that these are actually the preview blocks in the "Styles" tab making these requests; the requests made by the block "in the editor" are using the `newspack-blocks/v1/newspack-blocks-posts` endpoint (and have debouncing in place).

adekbadek:
@creativecoder debounce added in b5824c9 👍

adekbadek (approved):

dkoo (approved):
Works as described 👍

PR #731, opened 2021-04-02
Carousel: handle posts without featured images

### LINKED ISSUE #443: Articles Carousel: missing placeholder for empty state
**Describe the bug**
When there are no posts with featured images, the Articles Carousel renders without slides. To the user this will appear broken and leaves them at a loss as to what is wrong.

**Expected behavior**
We should provide the user with clear instructions on how to resolve the issue. Ideally, when there are no featured image posts, there should be a placeholder instead of the carousel that tells the user that they need to add featured images to their posts in order for those to appear in the Articles carousel. 

**To Reproduce**
1. Create posts without a featured image
2. Create a new post and added the Articles Carousel

**Screenshots**

Screenshot 2020-04-28 at 13 38 11

adekbadek:
### All Submissions:

Closes #443

1. Insert three Carousel blocks on a page: one that renders posts with featured images, one that renders posts without any, and one which would render no posts
2. In the editor, observe the first is behaving as expected, the second is displaying grey surface in place of the featured image, and the last is displaying a center-aligned error message (and no carousel arrows)
3. On frontend, first two should behave as in the editor, while the last should not render any visible content

### Other information:


src/blocks/carousel/view.php
```diff
@@ -66,28 +63,30 @@ function newspack_blocks_render_block_carousel( $attributes ) {
 		while ( $article_query->have_posts() ) :
```
dkoo, line 63: If the carousel has no posts, I still see empty carousel UI on the front-end. At this point, if `! $article_query->have_posts()`, we should just return early before the UI below line 220 gets rendered.

Screen Shot 2021-04-05 at 8 50 12 PM

src/blocks/carousel/edit.js
```diff
@@ -146,70 +149,71 @@ class Edit extends Component {
...
-							/>
-							<button
-								className="amp-carousel-button amp-carousel-button-next swiper-button-next"
-								ref={ this.btnNextRef }
-							/>
-							<div
-								className="swiper-pagination-bullets amp-pagination"
-								ref={ this.paginationRef }
-							/>
+							{ ! hasNoPosts && (
```
dkoo, line 201: We should also not render the next/previous UI (buttons and bullets) if there's only one post, both here and on the front-end.

dkoo (changes requested):
Looking good overall, just a couple of minor UI requests below.

src/blocks/carousel/edit.js
adekbadek: Done in c439d60f50f5f72fd091be2cd60dd1abd822a088

src/blocks/carousel/view.php
adekbadek: Done in 9660922eda7f97fc2005bd71eb15e1ed67385b35

dkoo (approved):
This looks good now, thanks for the updates!

PR #750, opened 2021-05-03
feat: add carousel block attribute for slidesPerView

dkoo:
### All Submissions:

Adds a new Post Carousel block attribute to set the number of visible slides at a time. This will effectively make each slide visually smaller.

Also fixes a dependency bug for non-AMP pages where `@wordpress/x` modules weren't being imported correctly due to the `view.js` being enqueued too early.

Also renders carousels with a `slidesPerView` config setting that never exceeds the actual number of slides in the carousel, regardless of the block attribute setting.

Also fixes a bug where if a post in the carousel has neither a valid permalink nor an external URL, it was linking to a "not found" page.

#### Dependency bugfix

1. On `master`, add a Post Carousel to a post or page and turn off AMP.
2. On the front-end, observe a JS console error similar to `'@wordpress/i18n' Cannot read property 'sprintf' of undefined`
3. Also observe that the swiper bullets don't update properly when changing slides.
4. Check out this branch, rebuild and refresh. Confirm that both issues are fixed.

#### New attribute

1. On this branch, create or edit a Post Carousel.
2. Under the Slideshow Settings panel, observe a new range slider for "Number of slides to show at once"
3. Confirm that the min and default is `1`
4. Confirm that the max is either:
  - The "Number of items" setting, if in query mode
  - The total number of posts returned by the query, in either query or specific posts mode
  - 6, if any of the above is more than that
5. Confirm that changing the setting updates the slider view in both the preview and on the front-end, in both AMP and non-AMP mode.

#### `slidesPerView` never exceeds the number of slides

1. Create a Post Carousel and make sure it's rendering a bunch of posts (more than 5).
3. Set "Number of slides to show at once" to the max (6). Save the post.
4. Change the block to "specific posts" mode and select a small number of posts (2 works well).
5. Confirm that in the editor, the carousel changes to show only 2 posts at a time (there shouldn't be any extra whitespace at the end of the carousel).
6. Confirm that the "Number of slides to show at once" now shows `2`, or the number of posts in the carousel (if lower than the initial setting).
7. Confirm the same on the front-end, in both AMP and non-AMP mode.

#### No valid link fix

1. Create a sponsor and a supporter post (using the [Sponsors](https://github.com/Automattic/newspack-sponsors) and [Supporters](https://github.com/Automattic/newspack-supporters) plugins—Supporters must have [this branch](https://github.com/Automattic/newspack-supporters/pull/3) active). Add featured images and other info to both, but leave the "External URL" (Sponsors) and "Supporter URL" (Supporters) fields empty.
2. Create a Post Carousel and Homepage Posts block. Add the sponsor and supporter lacking links. For the Homepage Posts, block turn on the "Add a 'Read More' link" option.
3. Confirm that both in the editor preview on the front-end, the items are rendered in both blocks without linking anywhere, because they have neither a valid permalink nor external URL to link to.

### Other information:


dkoo:
Note: discovered #751 while testing for this, but will work on that in a separate PR.

src/blocks/carousel/view.php
```diff
@@ -70,6 +72,7 @@ function newspack_blocks_render_block_carousel( $attributes ) {
 									array(
 										'object-fit' => $attributes['imageFit'],
 										'layout'     => 'fill',
+										'class'      => 'contain' === $attributes['imageFit'] ? 'image-fit-contain' : 'image-fit-cover',
```
adekbadek, line 75: I'm getting a PHP notice here: `Undefined index: imageFit` - probably because the attribute is not registered in the [block registration call](https://github.com/Automattic/newspack-blocks/blob/006e7f8eb7acfc477f167e56bf63ffa6630c4fc6/src/blocks/carousel/view.php#L343).

adekbadek (changes requested):
The switch to `amp-base-carousel` might enable us to start using [Bento AMP, as it is a ported component](https://amp.dev/documentation/components/amp-base-carousel/?format=websites) 😎

1. The block is cropped in non-AMP & editor contexts when set to more than one slides to show at once:

image

2. In AMP context, something's off with the layout:

image

image

src/blocks/carousel/view.php
dkoo: Fixed in c331f21

dkoo:
> The switch to `amp-base-carousel` might enable us to start using [Bento AMP, as it is a ported component](https://amp.dev/documentation/components/amp-base-carousel/?format=websites) 😎

Nice! I haven't seen any major issues with the switch from `amp-carousel` to `amp-base-carousel`, but the latter is what lets us display multiple slides at once.

> 1. The block is cropped in non-AMP & editor contexts when set to more than one slides to show at once

The use of `position: absolute` with no fixed height on the `.entry-wrapper` element makes this hard to avoid when there are multiple slides being shown at once. It's extra unavoidable in the AMP context because of the `layout="responsive"` on the `amp-base-carousel`, which forces the slides to a specific aspect ratio (4:3).

Since the main use case of the multiple images being shown at once is sponsor logos with no `.entry-meta` content, I think we can leave this for now, and if necessary advise users to not try to cram too much content into each slide.

> 2. In AMP context, something's off with the layout:

I think c331f21 addresses the AMP styles a bit better—can you re-test?

src/blocks/carousel/view.php
dkoo: This was also fixed in #754, so resolved the conflict in 995f524.

src/blocks/carousel/view.php
```diff
@@ -56,27 +58,36 @@ function newspack_blocks_render_block_carousel( $attributes ) {
 			$counter++;
 			$has_featured_image = has_post_thumbnail();
 			$post_type          = get_post_type();
-			$external_url       = get_post_meta( $post_id, 'newspack_sponsor_url', true );
+			$sponsor_url        = get_post_meta( $post_id, 'newspack_sponsor_url', true );
+			$supporter_url      = get_post_meta( $post_id, 'newspack_supporter_url', true );
+			$external_url       = ! empty( $sponsor_url ) ? $sponsor_url : $supporter_url;
 			$link               = ! empty( $external_url ) ? $external_url : get_permalink();
+			$post_type_info     = get_post_type_object( $post_type );
+			$has_link           = ! empty( $post_type_info->public ) || ! empty( $external_url ); // False if a sponsor or supporter without an external URL.
```
adekbadek, line 66: Duplicated logic (from `includes/class-newspack-blocks-api.php`) could be abstracted to a `Newspack_Blocks` method.

src/blocks/carousel/view.php
```diff
@@ -224,18 +235,24 @@ function newspack_blocks_render_block_carousel( $attributes ) {
...
 		$selector    = sprintf(
 			'<amp-selector id="wp-block-newspack-carousel__amp-pagination__%1$d" class="swiper-pagination-bullets amp-pagination" on="select:wp-block-newspack-carousel__amp-carousel__%1$d.goToSlide(index=event.targetOption)" layout="container">%2$s</amp-selector>',
 			absint( $newspack_blocks_carousel_id ),
 			implode( '', $buttons )
 		);
 		$carousel    = sprintf(
-			'<amp-carousel width="4" height="3" layout="responsive" type="slides" data-next-button-aria-label="%1$s" data-prev-button-aria-label="%2$s" controls loop %3$s id="wp-block-newspack-carousel__amp-carousel__%4$s" on="slideChange:wp-block-newspack-carousel__amp-pagination__%4$s.toggle(index=event.index, value=true)">%5$s</amp-carousel>',
+			'<amp-base-carousel class="wp-block-newspack-carousel__amp-carousel" width="%1$s" height="3" layout="responsive" snap="true" type="slides" data-next-button-aria-label="%2$s" data-prev-button-aria-label="%3$s" controls loop %4$s id="wp-block-newspack-carousel__amp-carousel__%5$s" on="slideChange:wp-block-newspack-carousel__amp-pagination__%5$s.toggle(index=event.index, value=true)" advance-count="1" visible-count="%6$s">%7$s</amp-base-carousel>',
```
adekbadek, line 249: In order to mitigate the cropping issue, how about setting the `width` to `1` and allowing the user to choose aspect ratio via an attribute that would translate to `height` (in 0-1 range)?

adekbadek (changes requested):

adekbadek:
> Since the main use case of the multiple images being shown at once is sponsor logos with no .entry-meta content, I think we can leave this for now

With the block being available in Gutenberg, I don't think we have the luxury of making such assumptions ;)

dkoo:
> With the block being available in Gutenberg, I don't think we have the luxury of making such assumptions ;)

I hear you—maybe your suggestion to make the aspect ratio a configurable attribute will help with this.

src/blocks/carousel/view.php
dkoo: This is a great idea—will explore this option.

src/blocks/carousel/view.php
dkoo: Good idea! I added a helper method for this in 54d2be2.

includes/class-newspack-blocks.php
```diff
@@ -195,6 +195,9 @@ public static function block_classes( $type, $attributes = array(), $extra = arr
 		if ( ! empty( $attributes['align'] ) ) {
 			$classes[] = 'align' . $attributes['align'];
 		}
+		if ( self::is_amp() ) {
+			$classes[] = 'is-amp';
```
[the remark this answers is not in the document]
dkoo: Removed in 54d2be2.

dkoo:
@adekbadek, in 91e87e4 I added some new block attributes to control aspect ratio (also another simpler attribute that lets you hide the slideshow controls). Test instructions:

#### Hide controls

1. Under "Slideshow Settings", observe a new option to "Hide Controls", which is disabled by default.
2. Enable the option and confirm that the slideshow controls are visually hidden a.) in the editor preview, b.) on the front-end with AMP on, and c.) on the front-end with AMP off.

#### Aspect ratio

Note that it's still possible to create a slideshow with text that ends up overflowing the container, especially if you use a short aspect ratio such as `16:9`, but it's much easier to avoid using the options available now.

1. Under "Slideshow Settings", observe a new set of buttons to set the aspect ratio. The default value is `4:3`, which is what the AMP carousel was previously set to under all conditions.
2. Try out the various options available. Confirm that changing the option updates the aspect ratio of all the slides in the slideshow in the editor preview.
3. Test the various options on the front-end with both AMP and non-AMP. Confirm that the slides render as expected.

dkoo:
Added another feature in 3da829f: below the mobile breakpoint (600px as defined in the theme), the carousel will display a max of 2 slides (only if the attribute is set to greater than 1) in both AMP and non-AMP. This addresses publisher feedback that the slides get too visually small on mobile devices.

src/blocks/carousel/view.php
```diff
@@ -258,6 +274,9 @@ function newspack_blocks_render_block_carousel( $attributes ) {
 	}
 	$data_attributes = [
 		'data-current-post-id=' . $post_id,
+		'data-slides-per-view=' . $attributes['slidesPerView'],
+		'data-slide-count=' . $counter,
+		'data-aspect-ratio=' . $attributes['aspectRatio'],
```
adekbadek, line 279: Getting a PHP notice here: `Undefined index: aspectRatio`

includes/class-newspack-blocks-api.php
```diff
@@ -367,6 +367,7 @@ public static function posts_endpoint( $request ) {
 				'newspack_post_format'            => self::newspack_blocks_post_format( $data ),
 				'newspack_post_sponsors'          => self::newspack_blocks_sponsor_info( $data ),
 				'post_type'                       => $post->post_type,
+				'post_link'                       => Newspack_Blocks::get_post_link( $post_id ),
```
adekbadek, line 370: Getting a PHP notice here: `Undefined variable: post_id`

adekbadek (changes requested):
Just some small tweaks needed

src/blocks/carousel/view.php
dkoo: Ah, forgot to register that attribute in PHP. Fixed in e484fbd.

includes/class-newspack-blocks-api.php
dkoo: Good catch—fixed in e484fbd.

adekbadek (changes requested):
I didn't notice before, sorry - in non-AMP version, the controls are hidden when the screen is below 600px wide.

adekbadek (approved):
The missing arrows are ok - this is for mobile users, who are more likely to swipe anyway.

PR #763, opened 2021-05-28
feat: author profile block

### LINKED ISSUE #723: New block: Author Profile
_Description and mockups courtesy @avcascade and @thomasguillot_

### What the block would do

This block would allow a publisher to insert a guest author profile (from Co-Authors Plus) anywhere they wanted. It would likely be used most frequently on publishers’ About pages to build staff directories, but it could also be used on other pages, or in posts.

Upon inserting the block, a user would see a search box where they could search for the name of one or more guest authors (including guest authors mapped to users). Upon selecting an author, that person’s headshot/avatar would appear, if one is set, along with their name and biography, by default. The author’s name would be linked to their author profile page, with a listing of what they have published.

Options could include:

- Hide the biography for the authors shown
- Turn off links to the authors’ profiles
- Hide the avatars/images for the authors

### Example use in a page

Page outline:

**Page Title** (e.g. Publication Staff)

- Paragraph block — introduction to the staff
- Subheading with or without paragraph — Newsroom
- Author profile block (Reporters and editors)
- Subheading with or without paragraph — Editorial & Opinion
- Author profile block (Editorial writers)
- Subheading with or without paragraph — Business
- Author profile block (Advertising and circulation representatives)
- Mockup of the hypothetical block
- Upon inserting the block, a user might see something like this:

### Mockups

Placeholder prior to selecting an author: a search box but also the 10 most recent authors (organised by latest post published)
author-profile-placeholder

Front-end: The block itself would be pretty straightforward. In the toolbar maybe we could have an option for the vertical alignment
author-profile-block

Sidebar attributes:
author-profile-sidebar

Some options in the sidebar may include:

- Display Bio Info
- Display Social Links
- Display Email Address
- Maybe even “Display Website”?
- Link to Author Archive
- Display Avatar:  if “Display Avatar” is enabled, but the Author doesn’t have one, maybe we could have an Image block with the upload placeholder? Uploading an image here would update the author’s Profile Picture.
- (if enabled) Border Radius

dkoo:
### All Submissions:

Registers a new "Author Profile" block which allows editors to choose either vanilla WP users or Co-authors Plus guest authors and display them in a standardized format in post or page content.

Also updates some NPM dependencies because I was having trouble with the latest version of `newspack-components` without updating, and many packages are very outdated. Please let me know if you see any weirdness in other blocks that may be a result of the updates.

Closes #723.

1. In a post, confirm there's a new "Author Profile" block under the Newspack category in the block inserter.
2. Insert an author profile block. Observe the search UI and confirm that the list of suggestions shows both regular WP users and Co-authors guest authors*:

Screen Shot 2021-05-28 at 4 34 45 PM

3. Save the post without selecting an author. View on the front-end; confirm that the block renders nothing if no author is selected.
4. Back in the editor, select an author from the suggestions list and by typing a name into the autocomplete input. (Use the "clear" button to choose another author after selecting one.) Add a second Author Profile block and make sure one contains a regular WP user and the other a guest author. Confirm that the preview shows the author info in both cases as expected**:

Screen Shot 2021-05-28 at 4 57 43 PM

5. Save the post, view on the front-end. Confirm that both blocks render as expected( also test at smaller viewports to confirm that the styles collapse with the avatar on top of the bio info):

Screen Shot 2021-05-28 at 4 57 34 PM

6. Back in the editor, play with the various attribute options. Confirm that both the editor preview and the front-end render as expected based on the options selected.

Screen Shot 2021-05-28 at 4 40 10 PM

7. Disable the Co-authors Plus plugin. On the front-end, confirm that blocks containing guest authors now render nothing. In the editor, confirm that these blocks show a "not found" warning:

Screen Shot 2021-05-28 at 4 29 49 PM

### Notes

\* Because guest authors are a custom post type, we can't query both regular users and guest authors in a single query. Instead, we must query them separately and combine the results. For this reason, the list of suggestions will always alternate between the two types of authors and may not display strictly in reverse chronological order.

** The only _display_ difference between regular WP users and guest authors is in the social links: WP users have more metafield options for various social media services, whereas guest authors basically have just a website and email field (also Yahoo, AIM, and Jabber, but these aren't used elsewhere in the theme so I didn't include handling for them). WP users and guest authors also differ in the format of their edit links and archive page links.

### Other information:


dkoo:
@thomasguillot I notice that for square images like the avatars, there's no visual difference between `border-radius: 50%` and any value higher than 50%. Should we limit the range to a max of 50?

src/blocks/author-profile/edit.js
```diff
@@ -0,0 +1,278 @@
...
+						__( 'No authors or guest authors found for ID %s.', 'newspack-blocks' ),
+						authorId
+					)
+			);
+		}
+		setIsLoading( false );
+	};
+
+	// Combine social links and email, which are shown together.
+	const socialLinks = showSocial && author && author.social ? { ...author.social } : {};
```
adekbadek, line 83: Nit: `{...obj}` is same as `obj`:
```suggestion
	const socialLinks = ( showSocial && author && author.social ) || {};
```

src/blocks/author-profile/index.js
```diff
@@ -0,0 +1,46 @@
...
+export const settings = {
+	title,
+	icon: {
+		src: <Icon icon={ people } />,
+		foreground: '#36f',
+	},
+	keywords: [
+		__( 'donate', 'newspack-blocks' ),
+		__( 'memberships', 'newspack-blocks' ),
+		__( 'subscriptions', 'newspack-blocks' ),
```
adekbadek, line 34: These keywords are a bit odd in the context of authors - should be updated ;)

adekbadek (changes requested):
Words fine, I like how versatile the `AutocompleteWithSuggestions` component is! Just a small block-keywords nitpick to fix.

thomasguillot:
> And, I couldn't figure out how to create a number input like you had in the mockup

The Columns block is using this input

Screenshot 2021-06-02 at 17 20 18
Screenshot 2021-06-02 at 17 20 26

If it's too complex, let's limit the range to a max of 50%.

thomasguillot:
Screenshot 2021-06-02 at 17 52 48

Where is Dev McFly's avatar? 🤔  You can see in my dev tools that there is an image... is it because of lazy loading and it's a Gutenberg issue?

thomasguillot (changes requested):
Overall this looks good. A few things:

* Move the "Clear" option to the toolbar with an Edit icon
* Remove the alignment option (is this really needed?)
* Disable links
* When in a group:
    * Fix the avatar resizing issue
    * Make sure the links and headings have the correct colour
Screenshot 2021-06-02 at 18 01 24

dkoo:
@thomasguillot Thanks for the design feedback! Just a few questions to clarify:

> Disable links

What do you mean by this? Remove any links in the author's bio text?

> Make sure the links and headings have the correct colour

Do we have functions in the theme or blocks plugin to determine the correct color based on background color? Could you point me to an example of a feature that does this so I can follow the same pattern?

thomasguillot:
>  I think I get it—you mean disabling outgoing links in the editor preview, right?

Yes

thomasguillot:
> Do we have functions (or CSS class names) in the theme or blocks plugin to determine the correct color based on background color? Could you point me to an example of a feature that does this so I can follow the same pattern?

Should be fixed in https://github.com/Automattic/newspack-blocks/pull/763/commits/a9b8e3e60e30e8e1aca7cba981e71f84f74b95df

thomasguillot:
Another idea, instead of a strict 240px avatar. What if we had a S/M/L/XL option?

* 72
* 128 (default)
* 192
* 256

dkoo:
> Another idea, instead of a strict 240px avatar. What if we had a S/M/L/XL option?

That's a good idea. That way, the relative widths of the avatar vs. author meta would be fluid based on the avatar size.

src/blocks/author-profile/edit.js
dkoo: I was maybe a bit overzealous when it comes to not mutating the `author` state object, but your suggestion doesn't seem to cause any issues as far as I can tell. Updated in d7e4a91.

src/blocks/author-profile/index.js
dkoo: Whoops, fixed in  in d7e4a91!

dkoo:
> The Columns block is using this input

Thanks for the example! Updated to the `UnitControl` in bc637f5.

dkoo:
> A few things:
> 
> * Move the "Clear" option to the toolbar with an Edit icon
> * Remove the alignment option (is this really needed?)
> * Disable links

All updated in 7aa9e38!

dkoo:
> Another idea, instead of a strict 240px avatar. What if we had a S/M/L/XL option?

This is implemented in b8f79f0.

Screen Shot 2021-06-03 at 4 42 45 PM

thomasguillot (approved):
Looks good! Thanks @dkoo

adekbadek (approved):

PR #766, opened 2021-06-02
Added margin-bottom definition for grid layout

pbking:
When the homepage articles block was configured to be in grid form, use
borders style (or regular style and the view width was smaller than tablet),
had featured image content and content text the last item
was taller than the rest (it had no bottom margin).

This change copies the behavior of NON-borders style by having a higher
specificity margin-bottom for the article and updates both to break on
MOBILE instead of TABLET.

Before:
image

After:
image

Closes: [Themes Bug 1797](https://github.com/Automattic/themes/issues/1797

### DIFF
--- a/src/blocks/homepage-articles/view.scss
+++ b/src/blocks/homepage-articles/view.scss
@@ -36,7 +36,7 @@
 		article {
 			flex-basis: 100%;
 
-			@include media( tablet ) {
+			@include media( mobile ) {
 				margin-bottom: 1em;
 			}
 		}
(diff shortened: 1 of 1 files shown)


laurelfulford (commented):
Thanks @pbking! 

I can recreate the issue with the featured image set to behind, the posts set to grid, and when the screensize is between mobile and tablet -- your PR also fixes that for me.

However I couldn't recreate the issue with the borders -- I tried with the featured image behind, and above, and it didn't happen for me regardless of screen size. Is there a specific theme you've noticed this issue with? I'm not noticing any issues with the changes you made, but I'm also not sure if this addresses everything it's meant to.

Thanks!

adekbadek:
Closing due to lack of response.

PR #768, opened 2021-06-08
Blog Posts Block: Force image to be the right size when used behind the block

scruffian:
### All Submissions:

When using the "show media behind" option in the block, these rules need to be stronger than others that might be applied to images. Using `!important` in CSS is generally considered bad practice, but I believe in this case it is the right tool for the job, since we always want this CSS to apply regardless of the context the block is loaded in. 

This will solve issues like https://github.com/Automattic/themes/issues/3341.

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/view.scss
+++ b/src/blocks/homepage-articles/view.scss
@@ -249,10 +249,11 @@
 				top: 0;
 
 				img {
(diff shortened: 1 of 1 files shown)


thomasguillot (approved):
LGTM 🚢

PR #785, opened 2021-07-01
Homepage Posts: add paragraph markup to excerpt, content placeholders

### LINKED ISSUE #783: Homepage Posts: Post excerpt not picking up font scale on Widgets screen
In WordPress 5.8 RC1, the Homepage Posts block is picking up most of the font scale styles from the front-end, but the post excerpt is not applying it -- you can see it in cases like where you set the font size very small:

image

It looks like the issue is that the excerpt preview doesn't include a `<p>` tag; this looks like it should be fixable by adding it to the example content for both the excerpt and content [here](https://github.com/Automattic/newspack-blocks/blob/427c4e087d57dee1fba9f4d81ffc8486f735d2c0/src/blocks/homepage-articles/utils.js#L124-L128).

laurelfulford:
### All Submissions:

In WordPress 5.8, the Homepage Posts block placeholders are used pretty prominently on the Widget screen, and actually apply the different styles used by the block settings.

Currently, the excerpt and content placeholders don't pick up the font scale setting, because they're not wrapped in a paragraph tag like the actual excerpt and content markup is. 

This PR adds a paragraph tag to each.  

Closes #783 

1. Start with a test site running WordPress 5.8; navigate to WP Admin > Appearance > Widgets and add the Homepage Posts block.
2. Try increasing and decreasing the font scale on the right; note the excerpt font size doesn't change in the preview, but everything else (headline, post meta) does.
3. Apply the PR and run `npm run build`
4. Confirm that the excerpt now picks up the font scale with the other elements. 

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/utils.js
+++ b/src/blocks/homepage-articles/utils.js
@@ -121,11 +121,11 @@ const PREVIEW_IMAGE_BASE = window.newspack_blocks_data.assets_path;
 const generatePreviewPost = () => ( {
 	author: 1,
 	content: {
-		rendered: __( 'The post content.', 'newspack' ),
+		rendered: '<p>' + __( 'The post content.', 'newspack' ) + '</p>',
 	},
 	date_gmt: new Date().toISOString(),
 	excerpt: {
-		rendered: __( 'The post excerpt.', 'newspack' ),
+		rendered: '<p>' + __( 'The post excerpt.', 'newspack' ) + '</p>',
 	},
 	featured_media: uniqueId(),
 	id: uniqueId(),


dkoo (approved):
This works as described, but just wanted to confirm: the `p` elements don't get any bigger after a type scale setting of `5`—is that intentional?

https://d.pr/i/2lYIlu

laurelfulford:
Thanks @dkoo! 

> the p elements don't get any bigger after a type scale setting of 5—is that intentional?

Yes, that's correct! It basically works like:

Font scale 1-3: everything changes (title, post meta, text, avatars)
Font scale 4-7: title, subtitle and avatars change
Font scale 8-10: only the title changes

... to make the post overall look bigger, without ending up with huge paragraphs or post meta - the paragraphs never get larger than the default text size.

PR #798, opened 2021-07-12
[WIP] fix: align grid items to left even if not enough items to fill last row

### LINKED ISSUE #396: Blog Posts Block: Weird distribution of grid items
#### Steps to reproduce
1. Test on a site with 5 blog posts.
2. Add the Blog Posts block to a page. Set the posts to display as a grid, 3 per row, display 6 posts at a time.
3. Notice how in the second row the last post is aligned to the right, leaving big gap in the middle.
4. Tested with Maywood, TwentyTwenty, Independent Publisher 2.

#### What I expected
I expected that if there weren't enough posts to fill the grid, the big gap would be at the bottom right.

#### Screenshot / Video
image

#dogfooding

dkoo:
### All Submissions:

Adds a small CSS hack to ensure that all grid child items are aligned as far to the left as possible, even if there aren't enough elements to fill the row.

Closes #396.

1. Create a Homepage Posts block and set the number of posts to something less than the number of columns, e.g. 2 posts /3 columns or 5 posts/6 columns. Observe a gap in the middle of the last row due to the use of `justify-content: space-between` on the grid:

Screen Shot 2021-07-12 at 12 02 50 PM

2. Check out this branch, rebuild. Confirm that the gap no longer exists in the editor or front-end, and that no extra space is created if you fill the last space (by making the number of posts and columns equal):

Screen Shot 2021-07-12 at 12 02 24 PM

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/view.scss
+++ b/src/blocks/homepage-articles/view.scss
@@ -30,6 +30,11 @@
 		justify-content: space-between;
 		padding: 0;
 		list-style: none;
+
+		&::after {
+			content: '';
+			display: block;
+		}
 	}
 
 	&.is-grid {
@@ -43,15 +48,21 @@
 	}
 
 	@include media( mobile ) {
-		&.columns-3 article,
-		&.columns-6 article {
-			flex-basis: calc( 33.333% - 16px );
+		&.columns-3,
+		&.columns-6 {
+			article,
+			&.is-grid > div::after {
+				flex-basis: calc( 33.333% - 16px );
+			}
 		}
 
-		&.columns-2 article,
-		&.columns-4 article,
-		&.columns-5 article {
-			flex-basis: calc( 50% - 16px );
+		&.columns-2,
+		&.columns-4,
+		&.columns-5 {
+			article,
+			&.is-grid > div::after {
+				flex-basis: calc( 50% - 16px );
+			}
 		}
 
 		&.columns-5 article:last-of-type {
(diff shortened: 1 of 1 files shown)


dkoo:
Ah, this doesn't fix instances where there's more than one "empty space" at the end of a row. For example, if there are only 2 posts to show, but 4 columns. Might need to revisit this fix for that reason, unless you have any ideas, @laurelfulford?

laurelfulford:
> Might need to revisit this fix for that reason, unless you have any ideas, @laurelfulford?

One other approach I can think of would be to make the container use `justify-content: flex-start`, and add the gutters using side margins on each article (like 8px left and right), and a negative left and right margin on the container so the block/articles appear to be the same width as the other content on the page.

However, this might be a bit too many changes for something that's also on WP.com, I'm not sure what it might mess up there 😬 On the Newspack side, would necessitate a small change in the theme (there's some max-width styles for the blocks that we'd need to override -- `max-width: calc( 100% + 32px)` rather than just `max-width: 100%`, and we'd also need to test with the block's border styles to make sure they still look right. 

(CSS grid could be a better approach overall at this point, but that's even change-ier than switching the `justify-content` style to something else!)

dkoo:
> However, this might be a bit too many changes for something that's also on WP.com

We could tackle this after we do the sync for 5.8 fixes, which would give us more time to test it on WP.com for a future sync, if you want?

dkoo:
This is a pretty stale PR. Since we're now _two_ WP versions beyond when this was tested, let's close this PR and revisit in the future.

PR #802, opened 2021-07-14
Carousel: Ensure that the swiper arrows don't overlap in RTL

ramonjd:
### All Submissions:

Adds `right: auto` to `.swiper-button-prev` to ensure the previous arrow button does not overlap the next arrow.

In RTL locales such as Arabic or Hebrew, the previous arrow styles are being dictated by SwiperJS CSS (`right: 10px`). 

**What I expected:**

Screen Shot 2021-07-14 at 7 57 11 pm

**What I see:**

Screen Shot 2021-07-14 at 7 46 24 pm

1. Insert a Post Carousel block
2. Switch your site and user language to Arabic/Hebrew
3. Ensure that the carousel arrows do not overlap in the editor or the frontend
4. Switch back to a LTR language and make sure these changes don't affect the arrows.

### Other information:

### DIFF
--- a/src/blocks/carousel/view.scss
+++ b/src/blocks/carousel/view.scss
@@ -166,6 +166,7 @@
 	.swiper-button-prev {
 		left: 1.5em;
 		display: none;
+		right: auto;
 
 		@include media( mobile ) {
 			display: block;
--- a/src/blocks/carousel/editor.scss
+++ b/src/blocks/carousel/editor.scss
@@ -30,6 +30,7 @@
 			left: 1.5em;
 			margin-top: -24px; // buttons height / 2
 			transform: translateY( -50% );
+			right: auto;
 		}
 
 		&.amp-carousel-button-next {


src/blocks/carousel/view.scss
```diff
@@ -166,6 +166,7 @@
 	.swiper-button-prev {
```
adekbadek, line 166: This will work for non-AMP pages only - SwiperJS is used only when page is non-AMP. The PR should be updated to handle RTL on AMP pages too, but since it's working (even if partially) I'll accept.

adekbadek (approved):

ramonjd:
Thanks for the feedback @adekbadek !

For future reference, where's the best place to add the relevant CSS fix for AMP pages? In the same files? I thought inserting the definition under

```css
.amp-carousel-button-next,
.amp-carousel-button-prev ...
```

would have done the trick, but no?

adekbadek:
We've switched from `amp-carousel` component to `amp-base-carousel` recently, and it looks like some parts were not updated. This move opens the way to use the same code on both AMP and non-AMP pages, since `amp-base-carousel` is one of non-AMP-ready components (called Bento AMP). I think we should first switch to using `amp-base-carousel` instead of SwiperJS, then revisit the RTL issue.

cc @Automattic/newspack-product

PR #805, opened 2021-07-16
feat/nrh donations handling

adekbadek:
### All Submissions:

Removes forcing the Donate block to manual mode (done by the Newspack Plugin when the donation platform is NRH).

1. Set Newspack Plugin to branch `fix/donations-nrh-tweaks` (https://github.com/Automattic/newspack-plugin/pull/1046)
2. Set the platform in Reader Revenue wizard to News Revenue Hub and fill in NRH Organization ID and NRH Salesforce Campaign ID (use `montanafreepress` as the former for a valid ID)
3. Insert a donate block on a page, observe that it can be - optionally - configured manually
4. Test different configuration of the block (tiered/untiered, different suggested amounts)
5. Observe that with each configuration, after clicking submitting the donation form on the frontend, you're redirected to NRH checkout with the correct donation amount

### Other information:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,6 +27,7 @@
         "react": "^16.12.0",
         "redux": "^4.0.5",
         "redux-saga": "^1.1.3",
+        "regenerator-runtime": "^0.13.7",
         "swiper": "4.5.1"
       },
       "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
     "react": "^16.12.0",
     "redux": "^4.0.5",
     "redux-saga": "^1.1.3",
+    "regenerator-runtime": "^0.13.7",
     "swiper": "4.5.1"
   },
   "scripts": {
--- a/src/blocks/donate/edit.js
+++ b/src/blocks/donate/edit.js
@@ -118,7 +118,6 @@ class Edit extends Component {
 	};
 
 	getSettings() {
-		const { attributes, setAttributes } = this.props;
 		const path = '/newspack/v1/wizard/newspack-donations-wizard/donation';
 
 		this.setState( { isLoading: true }, () => {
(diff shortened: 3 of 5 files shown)


dkoo (changes requested):
This works as described and it's nice to be able to use the Donation settings with NRH checkout! Just a couple of things I saw which may or may not be actual issues:

1. I happened to have AMP Plus enabled and Stripe settings configured from the last time I tested the streamlined donation flow, so "streamlined" is still an option for my Donate blocks when NRH is the active platform. But this time when I turned it on, the block on the front-end threw a JS console error:

```
Uncaught ReferenceError: regeneratorRuntime is not defined
    <anonymous> webpack:///./src/blocks/donate/streamlined.js?:59
    <anonymous> webpack:///./src/blocks/donate/streamlined.js?:222
    [...]
```

This is after running `composer install && npm ci && npm run build` in both the plugin and Blocks repos. Am I missing an extra setup step somewhere?

2. This is a tiny thing that I don't consider a blocker, but mentioning just in case. When selecting a price in the block that's not a whole dollar amount, such as $7.50, this is rounded down in the NRH checkout for to the nearest dollar ($7) even though the param is passed correctly in the URL. This is probably a shortcoming on the NRH side, but should we consider allowing only integers in the Donation settings for this reason?

adekbadek:
> the block on the front-end threw a JS console error

Yeah, that's a WP 5.8 thing - `regenerator-runtime` is no longer enqueued by default. Fixed in 542d5743ccabd822bbe339e3b553747f5b0c5f1d

> When selecting a price in the block that's not a whole dollar amount, such as $7.50, this is rounded down in the NRH checkout for to the nearest dollar ($7) even though the param is passed correctly in the URL. 

Hm, that's unfortunate. I think this is rather an edge case - I've [updated the default on the plugin side](https://github.com/Automattic/newspack-plugin/pull/1046/commits/608fe321b34e05bca08d822fbac97ad87a09e51b), while it does not really handle this issue, it's close enough IMO.

dkoo (approved):
Not seeing any JS errors after the `regenerator-runtime` fix. Agree about making the default donation levels integers—should be enough for now. Thanks!

PR #859, opened 2021-09-07
feat: add iframe block

kariae:
add an iframe block that will let publishers to embed an iframe either by its URL or by uploading
the iframe archive

### All Submissions:

Add an iframe block to let publishers embed iframes by adding the iframes URL or by uploading the iframes content.

1. Pull the branch related to this PR.
2. Add a new post.
3. In the new post add the iframe block.
4. Add the iframe source URL and click on the embed button.
5. Observe the iframe embed in the post.
6. If set to full-screen mode in the sidebar, preview the post and observe that the iframe is filling all the page.
7. On another post add again an iframe block, and this time upload a zip archive that contains the iframe files (HTML, CSS, images, ...).
8. Observe the embed iframe in the post.

### Other information:


src/blocks/iframe/edit.js
```diff
@@ -0,0 +1,178 @@
...
+				<div className="iframe-container">
+					<FocusableIframe
+						title="iframe title"
+						src={ src }
+						style={ {
+							width: isFullScreen ? '100vw' : width,
+							height: isFullScreen ? '100vh' : height,
+							'max-width': '100%',
+							'max-height': '100%',
+						} }
```
adekbadek, line 120: Full-screen does not really work for the editor - I think a better solution would be to add a large notice saying that this block will take over the page rendering.

src/blocks/iframe/view.php
```diff
@@ -0,0 +1,97 @@
...
+ */
+function newspack_blocks_get_iframe_html( $src, $height, $width, $is_full_screen ) {
+	$height = $is_full_screen ? '100' : $height;
+	$width  = $is_full_screen ? '100' : $width;
+	$style  = "height: $height; width: $width;";
+	$layout = 'responsive';
+
+	if ( $is_full_screen ) {
+		$layout = 'fill';
+		$style  = 'height: 100vh; width: 100vw; max-width: 100vw; max-height: 100vh; position: fixed; top: 0; left: 0; right: 0; bottom: 0; z-index: 999999; margin-top: 0!important;';
```
adekbadek, line 77: The actual post is still rendered underneath, which causes scrolling issues - not mentioning all other possible side effects of having a full content rendered but hidden. The rendering of a full-screen iframe should take over the rendering of the post content (`the_content` hook), or even the whole page.

src/blocks/iframe/class-wp-rest-newspack-iframe-controller.php
```diff
@@ -0,0 +1,184 @@
...
+ * @package WordPress
+ */
+
+// phpcs:disable WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedClassFound
+
+/**
+ * Class WP_REST_Newspack_Iframe_Controller.
+ */
+class WP_REST_Newspack_Iframe_Controller extends WP_REST_Controller {
+	const IFRAME_UPLOAD_DIR = '/iframes/';
```
adekbadek, line 14: The folder name should indicate that it's related to Newspack Blocks' Iframe block.

src/blocks/iframe/class-wp-rest-newspack-iframe-controller.php
```diff
@@ -0,0 +1,184 @@
...
+	 */
+	public function process_iframe_archive( $request ) {
+		$response = [];
+		$files    = $request->get_file_params();
+
+		if ( count( $files ) > 0 && array_key_exists( 'archive_file', $files ) ) {
+			$archive_file      = $files['archive_file'];
+			$wp_upload_dir     = wp_upload_dir();
+			$iframe_upload_dir = $wp_upload_dir['path'] . self::IFRAME_UPLOAD_DIR;
+			$iframe_folder     = pathinfo( $archive_file['name'] )['filename'] . wp_generate_password( 8, false );
```
adekbadek, line 76: Nit: adding a dash between the filename and random string would improve readability.

src/blocks/iframe/edit.js
```diff
@@ -0,0 +1,178 @@
...
+		},
+	];
+
+	return (
+		<Fragment>
+			{ src && ( ! isSelected || showPreview ) ? (
+				<div className="iframe-container">
+					<FocusableIframe
+						title="iframe title"
+						src={ src }
```
adekbadek, line 114: After uploading a ZIP archive, the preview did not work for me - despite the source URL returning the correct content when visited directly in the browser.

src/blocks/iframe/edit.js
```diff
@@ -0,0 +1,178 @@
...
+			isActive: src && showPreview,
+		},
+	];
+
+	return (
+		<Fragment>
+			{ src && ( ! isSelected || showPreview ) ? (
+				<div className="iframe-container">
+					<FocusableIframe
+						title="iframe title"
```
adekbadek, line 113: That's not very descriptive - it should rather say something about the block, and be translatable.

src/blocks/iframe/view.php
```diff
@@ -0,0 +1,97 @@
...
+ * @return string
+ */
+function newspack_blocks_render_block_iframe( $attributes ) {
+	Newspack_Blocks::enqueue_view_assets( 'iframe' );
+
+	return newspack_blocks_get_iframe( $attributes );
+}
+
+/**
+ * Registers the `newspack-blocks/donate` block on server.
```
adekbadek, line 22:
```suggestion
 * Registers the `newspack-blocks/iframe` block on server.
```

src/blocks/iframe/view.php
```diff
@@ -0,0 +1,97 @@
...
+		)
+	);
+}
+add_action( 'init', 'newspack_blocks_register_iframe' );
+
+/**
+ * Get video playlist data for general args.
+ *
+ * @param array $args Arguments. See $defaults.
+ * @return array of video info.
```
adekbadek, line 46: Here & couple of lines up - should be changed from 'video'

src/blocks/iframe/view.php
```diff
@@ -0,0 +1,97 @@
...
+		'width'        => '100%',
+		'isFullScreen' => false,
+	);
+	$args     = wp_parse_args( $args, $defaults );
+
+	return newspack_blocks_get_iframe_html( $args['src'], $args['height'], $args['width'], $args['isFullScreen'] );
+}
+
+/**
+ * Get embed html for an array of YouTube video URLs.
```
adekbadek, line 61: Description to be updated after copying

adekbadek (changes requested):
Apart from the individual comments, I'm wondering whether the block name is not too technical, but I don't have any counter-propositions.

src/blocks/iframe/edit.js
kariae: I will still set the iframe size so it can be rendered with a different size than the one set by the user in case they toggle the fullscreen mode.

I added the notice too, good point!

src/blocks/iframe/view.php
kariae: `the_content` hook seems like a good one for this case, I changed the rendered content and also hide the rest of the page element.

src/blocks/iframe/edit.js
kariae: Do you have any errors in the console?

thomasguillot:
@kariae I've pushed some simple visual changes. But I'd like to see a few more.

I'm not a fan of the Embed Block placeholder style and I think it looks messy having a URL input and a Button on 2 different row. I much prefer the Image Block placeholder style.

For the Height/Width settings, can we use something similar to the Cover Block regarding the unit? At the moment it's hard-coded in the input.

I'm not sure we need to have the "Preview" state when the block is unfocused. The editor "jumps" when I click on the block. Plus, there's already a button in the toolbar to show the preview.

kariae:
@thomasguillot thank you for the feedback.

> I'm not a fan of the Embed Block placeholder style and I think it looks messy having a URL input and a Button on 2 different row. I much prefer the Image Block placeholder style.

If we do go with the Image Block placeholder style, I'll add an `insert from URL` button, when we do click on it do I need to show an input+button on the bottom or I'll open a popup with the URL input and a confirmation button?

>For the Height/Width settings, can we use something similar to the Cover Block regarding the unit? At the moment it's hard-coded in the input.

I like this one, I'll add it.

> I'm not sure we need to have the "Preview" state when the block is unfocused. The editor "jumps" when I click on the block. Plus, there's already a button in the toolbar to show the preview.

After using the block on some posts I can confirm that, I'll remove it :)

thomasguillot:
> If we do go with the Image Block placeholder style, I'll add an `insert from URL` button, when we do click on it do I need to show an input+button on the bottom or I'll open a popup with the URL input and a confirmation button?
> 

Popup, just like the Image Block placeholder.

Screenshot 2021-09-20 at 13 45 46

adekbadek:
What's the status here? I was waiting for an approval from @thomasguillot as a sign that design-wise the work is done.

src/blocks/iframe/class-wp-rest-newspack-iframe-controller.php
```diff
@@ -0,0 +1,250 @@
...
+			$iframe_path       = $iframe_upload_dir . $iframe_folder;
+
+			// create iframe directory if not existing.
+		if ( ! file_exists( $iframe_upload_dir ) ) {
+			wp_mkdir_p( $iframe_upload_dir );
+		}
+
+			// unzip iframe archive.
+			$zip = new ZipArchive();
+			$res = $zip->open( $media_path );
```
adekbadek, line 160: Indentation should be fixed between here and this function block beginning - unfortunately `phpcbf` does not handle that.

src/blocks/iframe/class-wp-rest-newspack-iframe-controller.php
```diff
@@ -0,0 +1,250 @@
...
+		if ( count( $files ) > 0 && array_key_exists( 'archive_file', $files ) ) {
+			$archive_file  = $files['archive_file'];
+			$iframe_folder = pathinfo( $archive_file['name'] )['filename'] . '-' . wp_generate_password( 8, false );
+
+			$response = $this->process_iframe_archive( $request, $iframe_folder, $archive_file['tmp_name'] );
+		} else {
+			$response = new WP_Error(
+				'newspack_blocks',
+				/* translators: %s: iframe entry file (e.g. index,html) */
+				sprintf( __( 'Could not find the iframe archive.', 'newspack-blocks' ), self::IFRAME_ENTRY_FILE ),
```
adekbadek, line 95: Placeholder is missing in the translatable string.

src/blocks/iframe/class-wp-rest-newspack-iframe-controller.php
```diff
@@ -0,0 +1,250 @@
...
+			$media_path = get_attached_file( $data['media_id'] );
+
+			if ( $media_path && 'application/zip' === $media->post_mime_type ) {
+				$iframe_folder = $media->post_title . '-' . wp_generate_password( 8, false );
+
+				$response = $this->process_iframe_archive( $request, $iframe_folder, $media_path );
+			} else {
+				$response = new WP_Error(
+					'newspack_blocks',
+					sprintf( __( 'Could not find the assets archive media file.', 'newspack-blocks' ) ),
```
adekbadek, line 124: Placeholder is missing in the translatable string (no need to `sprintf`).

src/blocks/iframe/class-wp-rest-newspack-iframe-controller.php
```diff
@@ -0,0 +1,250 @@
...
+					[ 'status' => '400' ]
+				);
+			}
+		} else {
+			$response = new WP_Error(
+				'newspack_blocks',
+				/* translators: %s: iframe entry file (e.g. index,html) */
+				sprintf( __( 'Could not find the iframe archive.', 'newspack-blocks' ), self::IFRAME_ENTRY_FILE ),
```
adekbadek, line 132: Placeholder is missing in the translatable string.

src/blocks/iframe/view.php
```diff
@@ -0,0 +1,97 @@
...
+
+	if ( $is_full_screen ) {
+		$layout = 'fill';
+		$style  = 'height: 100vh; width: 100vw; max-width: 100vw; max-height: 100vh; position: fixed; top: 0; left: 0; right: 0; bottom: 0; z-index: 999999; margin-top: 0!important;';
+	}
+
+	ob_start();
+	?>
+	<figure class='wp-block-newspack-blocks-iframe'>
+		<div class='wp-block-embed__wrapper' style="height:<?php echo esc_attr( $height ); ?>; width:<?php echo esc_attr( $width ); ?>;">
```
adekbadek, line 83: A blank block (without URL nor .zip) should not render anything - it renders an empty 600px high element.

src/blocks/iframe/edit.js
```diff
@@ -0,0 +1,234 @@
...
+							width: isFullScreen ? '100vw' : width,
+							height: isFullScreen ? '100vh' : height,
+							'max-width': '100%',
+							'max-height': '100%',
+						} }
+						onFocus={ () => setShowPreview( false ) }
+					></FocusableIframe>
+				</div>
+			) : (
+				<IframePlaceholder
```
adekbadek, line 179: lAfter uploading a .zip or providing the iframe URL, the block stops displaying the placeholder (correct). Then, after clicking on the block again, the placeholder is displayed, but it does not go back to non-placeholder state after clicking away. IMO the preview should by auto-toggled only once - after setting URL/uploading .zip.

src/blocks/iframe/edit.js
```diff
@@ -0,0 +1,234 @@
...
+				body: formData,
+			} );
+
+			setAttributes( { src: iframeArchiveSrc, archiveFolder: iframeArchiveFolder } );
+			setFormSrc( iframeArchiveSrc );
+			setShowPreview( true );
+		} catch ( e ) {
+			setError(
+				e.message ||
+					e ||
```
adekbadek, line 108: I assume `e` will always be defined. If so, the fallback message will never be displayed. I'd suggest removing this line to handle only errors with a `message` key (which should be a string).

src/blocks/iframe/class-wp-rest-newspack-iframe-controller.php
```diff
@@ -0,0 +1,250 @@
...
+						sprintf( __( '%s was not found in the iframe archive.', 'newspack-blocks' ), self::IFRAME_ENTRY_FILE ),
+						[ 'status' => '400' ]
+					);
+				}
+			}
+		} else {
+			$response = new WP_Error(
+				'newspack_blocks',
+				/* translators: %s: iframe entry file (e.g. index,html) */
+				sprintf( __( 'Could not unzip the iframe archive.', 'newspack-blocks' ), self::IFRAME_ENTRY_FILE ),
```
adekbadek, line 210: Placeholder is missing in the translatable string.

src/blocks/iframe/class-wp-rest-newspack-iframe-controller.php
```diff
@@ -0,0 +1,250 @@
...
+				],
+			]
+		);
+
+		register_rest_route(
+			$this->namespace,
+			'/newspack-blocks-remove-iframe-archive',
+			[
+				[
+					'methods'             => WP_REST_Server::EDITABLE,
```
adekbadek, line 66: To stick to JSON APIs conventions, this should be `DELETABLE` (`DELETE` HTTP method) type route.

includes/class-newspack-blocks.php
```diff
@@ -38,6 +39,38 @@ public static function mark_view_script_as_amp_plus_allowed( $tag, $handle ) {
...
+		if ( has_block( 'newspack-blocks/iframe' ) ) {
+			$blocks = parse_blocks( get_post()->post_content );
+
+			foreach ( $blocks as $block ) {
+				if ( 'newspack-blocks/iframe' === $block['blockName']
+					&& array_key_exists( 'isFullScreen', $block['attrs'] )
+					&& $block['attrs']['isFullScreen']
+					) {
+					// we don't need the post content since the iframe will be fullscreen.
+					$content = render_block( $block );
```
adekbadek, line 58: Also, `newspack-popups` should be disabled on this page if the content is taken over by the iframe ([`newspack_popups_assess_has_disabled_popups` filter](https://github.com/Automattic/newspack-popups/blob/4b4fc5a4bcee9f72fdae0448ed23448ec64e382e/includes/class-newspack-popups-inserter.php#L620))

src/blocks/iframe/edit.js
```diff
@@ -0,0 +1,234 @@
...
+			setAttributes( { archiveFolder: null } );
+			setRemoveCurrentArchiveFolder( false );
+		}
+	}, [
+		src,
+		archiveFolder,
+		setArchiveFile,
+		removeCurrentArchiveFolder,
+		setRemoveCurrentArchiveFolder,
+	] );
```
adekbadek, line 140: Listing functions as `useEffect` dependencies is unnecessary - they will never be redefined.

src/blocks/iframe/edit.js
```diff
@@ -0,0 +1,234 @@
...
+
+const IframeEdit = ( { attributes, setAttributes } ) => {
+	const label = __( 'Iframe', 'block title' );
+	const { src, archiveFolder, isFullScreen, height, width } = attributes;
+	const [ formSrc, setFormSrc ] = useState( src );
+	const [ showPreview, setShowPreview ] = useState( false );
+	const [ isUploadingArchive, setIsUploadingArchive ] = useState();
+	const [ archiveFile, setArchiveFile ] = useState();
+	const [ error, setError ] = useState();
+	const [ removeCurrentArchiveFolder, setRemoveCurrentArchiveFolder ] = useState();
```
adekbadek, line 34: There's some repetition between the local state and block attributes, as well as unidiomatic use of state and effect hooks. The only really needed local state here is for error handling and preview display.

src/blocks/iframe/edit.js
adekbadek: My bad - I had a CORS issue due to local instance configuration.

src/blocks/iframe/view.php
```diff
@@ -0,0 +1,97 @@
...
+	register_block_type(
+		'newspack-blocks/iframe',
+		array(
+			'attributes'      => array(
+				'src'           => array( 'type' => 'string' ),
+				'archiveFolder' => array( 'type' => 'string' ),
+				'height'        => array( 'type' => 'string' ),
+				'width'         => array( 'type' => 'string' ),
+				'isFullScreen'  => array( 'type' => 'boolean' ),
+			),
```
adekbadek, line 34: To avoid repeating the attributes definition in JS & PHP, they can be defined in a JSON file shared by the two environments. See `src/blocks/donate/block.json`.

src/blocks/iframe/view.php
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * Registers the `newspack-blocks/iframe` block on server.
+ */
+function newspack_blocks_register_iframe() {
+	register_block_type(
+		'newspack-blocks/iframe',
+		array(
+			'attributes'      => array(
+				'src'           => array( 'type' => 'string' ),
+				'archiveFolder' => array( 'type' => 'string' ),
```
adekbadek, line 30: Is `archiveFolder` really needed as an attribute? The block rendering function does not use it.

src/blocks/iframe/edit.js
claudiulodro: I also ran into this issue. I agree, clicking off the block and then clicking on the block again should maintain the embed preview, not display the placeholder. 👍

src/blocks/iframe/view.php
adekbadek: Ah, yeah. Makes sense!

adekbadek (approved):
🙌

PR #874, opened 2021-09-17
fix: limit excerpt lengths even for custom excerpts

obenland:
Trims excerpts even if it's a custom excerpt for a uniform look.

### All Submissions:

Deprecated excerpt_length and excerpt_more methods in favor of a comprehensive excerpt callback.

Core only allows to control the length of generated excerpts, assuming that authors want to keep their custom excerpts as they're written. This PR will apply the excerpt length limit to both types of excerpts for a more uniform look.

Closes https://github.com/Automattic/wp-calypso/issues/51415

1. Once you checked out the branch and built the block…
2. Add a custom excerpt to a post, preferably a rather long one.
3. In a different post, insert the Homepage Posts block and manipulate the excerpt length.
4. Preview the post to make sure the front-end works as expected.

### Other information:

### DIFF
--- a/includes/class-newspack-blocks.php
+++ b/includes/class-newspack-blocks.php
@@ -733,7 +733,15 @@ class Newspack_Blocks {
 	/**
 	 * Closure for excerpt filtering that can be added and removed.
 	 *
-	 * @var newspack_blocks_excerpt_length_closure
+	 * @var Closure
+	 */
+	public static $newspack_blocks_excerpt_closure = null;
+
+	/**
+	 * Closure for excerpt length filtering that can be added and removed.
+	 *
+	 * @var Closure
+	 * @deprecated
 	 */
 	public static $newspack_blocks_excerpt_length_closure = null;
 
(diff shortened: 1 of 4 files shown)


mreishus:
> Once you checked out the branch and built the block…

@obenland I'm not sure how to do this. After I run `npm start` in this repository, I don't see any ports open up on my machine. Do I need to copy some of the built files to a calypso installation, or my sandbox to work with them?

mreishus:
Yup, I use wp-env, and I was able to get it working by adding the directory to the plugins list in `.wp-env.json`.

includes/class-newspack-blocks.php
```diff
@@ -771,6 +820,8 @@ function() use ( $attributes ) {
 
 	/**
 	 * Remove excerpt length filter after Homepage Posts block loop.
+	 *
+	 * @deprectated
```
mreishus, line 824: Spelling issue - this (and others) should be `@deprecated`

mreishus (approved):
The functionality is working well:



From a code standpoint, as someone not familiar with the code, I might wonder why both `filter_excerpt` and `filter_excerpt_length` exist, since they both seem to do the same thing. However the `@deprecated` tag might provide enough of an explanation.

~~Noted a small spelling error.~~

includes/class-newspack-blocks.php
```diff
@@ -752,6 +760,47 @@ public static function remove_wc_memberships_excerpt_limit() {
...
+
+			$post = get_post( $post );
+			$text = get_the_content( '', false, $post );
+
+			$text = strip_shortcodes( $text );
+			$text = excerpt_remove_blocks( $text );
+
+			/** This filter is documented in wp-includes/post-template.php */
+			$text = apply_filters( 'the_content', $text ); // phpcs:ignore
+			$text = str_replace( ']]>', ']]&gt;', $text );
```
adekbadek, line 779: Under what conditions would `]]>` appear in the content?

adekbadek (commented):
Works as expected, left one question.

includes/class-newspack-blocks.php
obenland: `]]>` marks the end of a CDATA section and could be part of a javascript snippet. [The changeset that introduced this to Core doesn't provide a whole lot more context](https://core.trac.wordpress.org/changeset/2443), but I wonder if it has anything to do with `wptexturize()` [running after it](https://github.com/WordPress/wordpress-develop/blob/f49d48fdcb30f48307ccab183f9103d8b4527e5c/src/wp-includes/default-filters.php#L181) and making sure it doesn't break RSS feeds? I mostly duplicated [`wp_trim_excerpt()`](https://developer.wordpress.org/reference/functions/wp_trim_excerpt/) to make sure we get a similar output for custom excerpts and generated ones (since they'll now be both somewhat generated).

adekbadek (approved):

PR #876, opened 2021-09-21
Add text alignment option when image featured image is at the top

### LINKED ISSUE #847: Homepage Posts: Centered alignment
We've had a few requests now to center the Homepage Posts text when the image is aligned top. 

It would be helpful to have a block option for this, so we don't have to use custom CSS classes.

philipjohn:
### All Submissions:

Adds an alignment control to the block when the featured image is shown and the image position is "top", allowing users to centre-align the block text when using that particular image style.

Closes #847.

1. Apply the PR and add a block, selecting the combination mentioned above for feature image
2. Observe the alignment control (dis-)appearing as you change the options
3. Observe that the correct CSS class is applied and the text is aligned accordingly.

Screenshot from 2021-09-21 15-13-26
Screenshot from 2021-09-21 15-13-13
Screenshot from 2021-09-21 15-13-00

One note: there is no horizontal line between the alignment control and the three dots - I wasn't sure why or how to correct that.

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/block.json
+++ b/src/blocks/homepage-articles/block.json
@@ -148,6 +148,10 @@
       "type": "array",
       "default": [ "post" ],
       "items": { "type": "string" }
-    }
+    },
+    "textAlign": {
+      "type": "string",
+	  "default": "left"
+	}
   }
 }
--- a/src/blocks/homepage-articles/edit.js
+++ b/src/blocks/homepage-articles/edit.js
@@ -30,6 +30,7 @@ import {
 	PanelColorSettings,
 	RichText,
 	withColors,
+	AlignmentControl,
 } from '@wordpress/block-editor';
 import {
 	Button,
@@ -608,6 +609,7 @@ class Edit extends Component {
 			showCaption,
 			showCategory,
 			specificMode,
+			textAlign,
 		} = attributes;
 
 		const classes = classNames( className, {
@@ -622,6 +624,7 @@ class Edit extends Component {
 			'has-text-color': textColor.color !== '',
 			'show-caption': showCaption,
 			'show-category': showCategory,
+			[ `has-text-align-${ textAlign }` ]: textAlign,
 			wpnbha: true,
 		} );
 
@@ -751,6 +754,14 @@ class Edit extends Component {
 					<Toolbar controls={ blockControls } />
 					{ showImage && <Toolbar controls={ blockControlsImages } /> }
 					{ showImage && <Toolbar controls={ blockControlsImageShape } /> }
+					{ showImage && mediaPosition === 'top' && (
+						<AlignmentControl
+							value={ textAlign }
+							onChange={ nextAlign => {
+								setAttributes( { textAlign: nextAlign } );
+							} }
+						/>
+					) }
 				</BlockControls>
 				<InspectorControls>{ this.renderInspectorControls() }</InspectorControls>
 			</Fragment>
--- a/src/blocks/homepage-articles/edit.js
+++ b/src/blocks/homepage-articles/edit.js
@@ -751,17 +751,17 @@ class Edit extends Component {
 				) }
 
 				<BlockControls>
-					<Toolbar controls={ blockControls } />
-					{ showImage && <Toolbar controls={ blockControlsImages } /> }
-					{ showImage && <Toolbar controls={ blockControlsImageShape } /> }
-					{ showImage && mediaPosition === 'top' && (
+					<Toolbar>
 						<AlignmentControl
 							value={ textAlign }
 							onChange={ nextAlign => {
 								setAttributes( { textAlign: nextAlign } );
 							} }
 						/>
-					) }
+					</Toolbar>
+					<Toolbar controls={ blockControls } />
+					{ showImage && <Toolbar controls={ blockControlsImages } /> }
+					{ showImage && <Toolbar controls={ blockControlsImageShape } /> }
 				</BlockControls>
 				<InspectorControls>{ this.renderInspectorControls() }</InspectorControls>
 			</Fragment>
--- a/src/blocks/homepage-articles/view.php
+++ b/src/blocks/homepage-articles/view.php
@@ -58,6 +58,9 @@ function newspack_blocks_render_block_homepage_articles( $attributes ) {
 	if ( isset( $attributes['className'] ) ) {
 		$classes .= ' ' . $attributes['className'];
 	}
+	if ( $attributes['textAlign'] ) {
+		$classes .= ' has-text-align-' . $attributes['textAlign'];
+	}
 
 	if ( '' !== $attributes['textColor'] || '' !== $attributes['customTextColor'] ) {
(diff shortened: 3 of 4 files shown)


laurelfulford (dismissed):
@philipjohn!!! 👋 ❤️ Thanks for tackling this! 

Looking at other blocks, the text alignment option is displayed in the same section as the 'wide-full' alignment in the toolbar -- what do you think about moving it there for this block, too?

image

In my testing, some items didn't pick up the alignment; I'm happy to assist with the CSS for this if it would help! The problem elements were:
* Categories
* Post meta (byline and date)
* Featured image captions
* The Load More Posts link. 

image

Lastly, I feel like there's some weirdness with the alignment itself, but I'm having a hard time articulating why:

Originally when I added this issue I just saw it as a way to centre the text under the image in the block, since that's something we'd been asked to do a couple times with Custom CSS. I don't think there's a good case for being able to align the text right -- I suspect all that is baked into the `AlignmentControl` option though. Would it make sense to roll our own so we can simplify that?

Also, I noticed when you align the text centre or right, and then change the image to be aligned left or right; you lose the option to change the text alignment, but the centred/right alignment is still applied. It should probably fall back to the default whenever the toggle isn't visible.

(I also worry that showing and hiding the toggle will be confusing, but on the flip-side, don't think an option to able to centre text next to a left-aligned image is a great thing to introduce). 

Please let me know if I can clarify anything above!

(Also, side note: we'll need to make sure the 'section titles' look okay with each of these options -- I created a ticket for that here and can take care of that! https://github.com/Automattic/newspack-theme/issues/1525)

philipjohn:
Thanks for the feedback Laurel!

> Looking at other blocks, the text alignment option is displayed in the same section as the 'wide-full' alignment in the toolbar -- what do you think about moving it there for this block, too?

Yeah, I had the same thought while building it. The main thing that stopped me was that I thought the UX could be janky. When you click on one of the image position options the text alignment would appear, moving those image position options (and others) to the left/right, changing which one your cursor is over. What do you think? Am I overthinking it?

> In my testing, some items didn't pick up the alignment

Oh well spotted, let me take a look at that and I'll shout if I get stuck.

> I don't think there's a good case for being able to align the text right -- I suspect all that is baked into the AlignmentControl option though. Would it make sense to roll our own so we can simplify that?

Initially I wondered if it would make sense to have the right option as folks might want that alignment when using the right image position. Part of me cringes at that idea :laughing: but I'm also wondering if there might be sensible use cases? If not, I'm sure we could role our own - I just wanted to avoid reinventing the wheel if possible! Like, if we did leave the right-align there, would it matter if it was never used, if it saves us re-creating our own?

philipjohn:
> Oh well spotted, let me take a look at that and I'll shout if I get stuck.

I've sorted some of the alignment issues in [3787a00] but struggled with the entry meta part as I'm just not familiar enough with flexbox (which is where the issue seems to lie) to know what the solution is I'm afraid.

laurelfulford:
@philipjohn no worries! I've pushed up some changes to align those items centre and right -- the weird required attribute is `justify-content: center` and `justify-content: flex-end`, respectively. (If you ever want to dig into that more, I found [this CSS Tricks guide](https://css-tricks.com/snippets/css/a-guide-to-flexbox/) invaluable, just for a list of the different options available).

> Initially I wondered if it would make sense to have the right option as folks might want that alignment when using the right image position. Part of me cringes at that idea 😆 but I'm also wondering if there might be sensible use cases? If not, I'm sure we could role our own - I just wanted to avoid reinventing the wheel if possible! Like, if we did leave the right-align there, would it matter if it was never used, if it saves us re-creating our own?

One possible use case I can think of is if we have a publisher with posts both in a LTR language and a RTL language, but that might be a stretch -- I'm not even sure if you would normally mix alignments in that case, or just go with one or the other. 

That being said, I _know_ it's going to be used when it shouldn't, and I'm not sure the opportunity offsets the bad decisions it makes possible.

These are the pros and cons I think to each approach:

**Just a centre control**
* **Pros:** avoid allowing text to be aligned right
* **Cons:** we'd have to roll our own control; we'll only want to show it in certain cases, which could make for a herky jerky toolbar (I really do think we should group it with the other alignment one regardless). 

**Regular text alignment**
* **Pros:** it uses an existing control; there's probably no need to remove it in certain cases, as it just becomes another block option.
* **Cons:** someone is going to start aligning text right "just because"; by introducing the control, we kind of give our approval that this is something that they should do 😕 

@thomasguillot -- your thoughts on this would be appreciated, too, if you see one approach better than another (or a third approach that would make sense instead!).

thomasguillot:
> @thomasguillot -- your thoughts on this would be appreciated, too, if you see one approach better than another (or a third approach that would make sense instead!).

@laurelfulford I'd say we should go with the second approach, "Regular text alignment". It offers a bit more flexibility with the right-align option but more importantly, it's consistent with the other blocks (e.g. Headings).

laurelfulford:
> I'd say we should go with the second approach, "Regular text alignment". It offers a bit more flexibility with the right-align option but more importantly, it's consistent with the other blocks (e.g. Headings).

Thanks @thomasguillot! That's a good point re: consistency -- I'll put aside my right aligned text worries 😆 

@philipjohn with the above approach, I think it makes sense to move the text alignment to sit next to the block alignment option, and have it always visible, so that also removes the changing toolbar problem. Once that's done, I'm happy to take it for another test drive to make sure nothing unexpected happens with the different image placements!

philipjohn:
Yep, that's cool. I've moved the alignment control over and made it unconditional in [092c77a]. I tried to put it in the same section as the width control, but I couldn't figure that out, so I put it in it's own toolbar next to it to separate it from the List/Block control. Let me know if that makes sense!

Screenshot from 2021-09-28 10-31-42

laurelfulford:
Thanks @philipjohn! 

I took a look at trying to get the alignment option in the same section, too, but the closest I got (using [the same `block="group"` attribute](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/heading/edit.js#L42) that the Headings block uses), but that just removes _all_ of the lines 😐 I think where you got it works much better! 

image

I'll take the different alignments for a spin later today!

laurelfulford:
@philipjohn I went through and tested the beside and behind image placements with the right and centred text alignments, and made a couple tweaks I want to highlight:

* I removed the alignment from the Load More Posts button. While it seemed 'normal' for it to be aligned centre, it seemed weird that it would be aligned right, or that they could be aligned inconsistently throughout the page. 
* With the Image Behind placement, the captions were originally styled to align right. After some hemming and hawing, I've left them this way with the left alignment, but with the centre and right alignments, set them to pick up those styles. 

Let me know if you have any questions about either of those! I think this looks good, but we might want to get a second set of eyes on it since I've been contributing code as well (or at least have you re-review the alignments, too)!

philipjohn:
Good call :+1: I think that's good. Perhaps @thomasguillot and @claudiulodro could give it a final review?

dkoo (approved):
The new alignment options work as described!

PR #886, opened 2021-09-30
add a filter to enable duplicating posts

kariae:
### All Submissions:

since we have a deduplication process for posts, we can't show the same post in two different
homepage-posts blocks. Which should be good for all cases. Using the `newspack-popups` and setting a
popup with a homepage-posts block will show the non-duplicated posts on the homepage and the popup
in the first time we're loading the page, but the second time (we won't have the popup since it's
showed only once or once a day) we'll have missing (robbed) posts from the homepage-posts block of
the page as explained here https://github.com/Automattic/newspack-popups/issues/126. We need to add
an exception for the deduplication system for such case.

This is needed for this https://github.com/Automattic/newspack-popups/pull/630

Normally this shouldn't change how the homepage-posts block works.
1. Check out this.
2. Add two homepage-posts blocks with the same category filter.
3. Observe the deduplication system is still working and the posts are not duplicated on both blocks.

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/templates/articles-loop.php
+++ b/src/blocks/homepage-articles/templates/articles-loop.php
@@ -19,9 +19,12 @@ call_user_func(
 		Newspack_Blocks::filter_excerpt_length( $attributes );
 		Newspack_Blocks::filter_excerpt_more( $attributes );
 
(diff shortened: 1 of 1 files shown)


miguelpeixe (approved):
Working as described!

kariae:
Thank you, Miguel!

PR #896, opened 2021-10-20
feat(homepage-posts): enable showing previously rendered posts too

kariae:
### All Submissions:

This is needed when we do use the homepage-posts block in an overlay, we need to show all the filtered posts even if they were rendered in another block previously

Technically this PR just adds a filter so nothing should change with this. If the filter is enabled, in the homepage-posts block, we'll render all filtered posts including the ones previously rendered. We needed this since the overlay prompts from the `newspack-popups` act as another content layer.

1. Create a prompt with placement before header, with a `homepage-posts` block, and filter the posts by a category.
2. Create an overlay prompt with a `homepage-posts` block, and filter the posts by the same category from 1.
3. Load the homepage and observe that the posts loaded in the overlay don't include the ones rendered in the before header prompt.
4. Checkout this, and in the `newspack-popups` checkout https://github.com/Automattic/newspack-popups/pull/630 since the filter is added when rendering an overlay.
5. Clear your cache, or re-open an incognito window and load the homepage.
6. Observe that the overlay prompt does include also the posts rendered in the before header prompt.

### Other information:

### DIFF
--- a/includes/class-newspack-blocks.php
+++ b/includes/class-newspack-blocks.php
@@ -465,14 +465,21 @@ class Newspack_Blocks {
 			$args['orderby']  = 'post__in';
 		} else {
 			$args['posts_per_page'] = $posts_to_show;
-			if ( count( $newspack_blocks_all_specific_posts_ids ) ) {
-				$args['post__not_in'] = $newspack_blocks_all_specific_posts_ids;
+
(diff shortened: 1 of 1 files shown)


miguelpeixe (approved):
🎉

PR #906, opened 2021-11-10
fix(iframe-block): add a trailing slash to iframe URL

kariae:
### All Submissions:

Without the trailing slash, the iframe URL does a redirection to add the slash and set the request on HTTP which raises a mixed content issue, adding the trailing slash would stop the redirection and minimize the calls to the server. The issue was raised in the Atomic environment but I was able to reproduce it locally too.

1. If you don't have HTTPS use a temporary Atomic site.
2. On a new post, add the iframe block.
3. Upload a zip archive with at least an index.html file that will be the entry point to your iframe (I attached a zip archive for tests).
4. Notice that the iframe doesn't preview, and the console shows a mixed content error due to a redirection to HTTP.
5. Pull this PR.
6. Upload again the same archive and notice this time there is no redirection and the iframe loads as expected.

### Other information:


dkoo (approved):
Works as described 👍

PR #928, opened 2021-11-22
feat(donate): agree to pay fees feature

adekbadek:
### All Submissions:

Adds a "Agree to pay fees" option to Donate block. Available only in the streamlined (Stripe) version.

image

If user has a non-standard Stripe fee (e.g. [is a non-profit](thttps://support.stripe.com/questions/fee-discount-for-nonprofit-organizations)), they can update it via options. UI for this is coming soon, for now these should be updated manually (`wp option set …`).

1. Enable AMP Plus and configure platform as Stripe in Reader Revenue wizard
2. Visit a page with the Donate block
3. Observe a new input in the form – "Agree to pay fees"
4. The input should be checked on by default and inform about the standard Stripe fee, which is 2.9% + ¢30
5. Test the donation with and without the input checked
6. Update the fee variables by setting `newspack_blocks_donate_fee_multiplier` & `newspack_blocks_donate_fee_static` options 

### Other information:


thomasguillot:
@adekbadek I made some small tweaks in https://github.com/Automattic/newspack-blocks/pull/928/commits/eca209292e9be9ecee2db05649ccf694a4978230

miguelpeixe (approved):
It's working as described!

Two non-blocking:

image

 - Missing space after `Agree to pay fees?` 
 - I miss seeing a total of what is going to be charged. When there are additional values it's common to have the total sum next to the donate/pay button.

adekbadek:
> Missing space after Agree to pay fees?

Fixed in e719d57c68537f063535c7b002dd7bdaf9dad0c7

> I miss seeing a total of what is going to be charged. When there are additional values it's common to have the total sum next to the donate/pay button.

Good idea, definitely something to consider. There are some planned changes to the block UI with the introduction of Apple/Google Pay – let's revisit this after that. Added https://github.com/Automattic/newspack-blocks/issues/965.

yogeshbeniwal:
@adekbadek Is there any extra settings or variable required for this? Still not getting sign up for newsletter and agree to pay fees tick box with AMP plus enabled, Using Stripe and Mailchimp connected.

adekbadek:
@yogeshbeniwal – 

The "Agree to pay fees" feature should be on by default. To be clear – this is the UI you're seeing (with Stripe, without the checkboxes):

image

?

---

As for the "sign up for newsletter" checkbox, make sure you have Newspack Newsletters connected to an ESP and then a subscriber list chosen in Reader Revenue wizard:

image

yogeshbeniwal:
@adekbadek Aha, it shows after clicking Donate Now button. My bad, thanks for clarification.

PR #946, opened 2021-11-29
Post Carousel: update sizing for multiple slides, mobile

### LINKED ISSUE #844: Post Carousel: Font size doesn't change, regardless how many slides are shown
**Describe the bug**
You can now set the Post Carousel to display 2 or 3 posts at once, but the font size doesn't change at all, and at smaller sizes, it feels like the post title should scale down a bit. 

image

We can also address this in the theme; I feel like it may make more sense here, but I'm open to feedback on that!

laurelfulford:
### All Submissions:

This PR makes some visual updates to the Post Carousel block:
* To scale down the title font size when displaying 3-4 slides at once
* To alter the number of posts shown on tablet and mobile-sized screens, so they don't get too croweded.
* To fix an issue with the aspect ratio getting smooshed on blocks displaying 3-4 slides, on smaller screen sizes. 

Closes #844.

1. Add four Post Carousel block to a page: one displaying one slide at a time, one displaying two, one displaying three and one displaying four. Publish. 
2. View on the front end; note the blocks showing 3-4 slides at a time are crowded: 

image

3. Scale down the browser window to roughly tablet and mobile sizes; note the appearance of the blocks, specifically how the text can get cut off on the 3-4 slide blocks, and how their aspect ratios seem squished when AMP is enabled: 

image

image

5. Check your page on a desktop sized window and confirm the entry title scales down when the slide count is higher: 

image

6. Scale down the browser window to a tablet-sized screen; the font should scale down, and the blocks with 3-4 slides visible should drop to two visible slides, and maintain their aspect ratios when AMP is enabled: 

image

7. Scale down the browser window to a mobile-sized screen; all blocks should only have one slide visible, and no aspect ratio issues: 

image

8. Check the page with AMP disabled, and make sure there aren't any visual issues.
9. Edit the page and try some different aspect ratios (like 1:1 and 16:9); confirm that they are used by your blocks without issue. (Note: the block has a max-height of 75vh, which is 75% of the current screen height. This means there are some cases where the aspect ratio won't be respected entirely -- for example, if you have a block displaying one slide with a 1:1 aspect ratio on desktop, it will probably be wider than it is tall. This is the expected behaviour with the max height that's set). 

image

image

(I did notice some combos -- like 16:9 with four slides in that last screenshot -- are still crowded. I think that's more of a case where we'd want to steer folks  away from using particular combinations of settings, rather than something we need to get to work well in all cases -- like in that case, it's probably best to only show the title, not the byline and date too. But I'm definitely open to feedback on that!)

### Other information:

### DIFF
--- a/src/blocks/carousel/create-swiper.js
+++ b/src/blocks/carousel/create-swiper.js
@@ -115,9 +115,12 @@ export default function createSwiper( els, config = {} ) {
 		touchStartPreventDefault: false,
 		breakpoints: {
 			320: {
+				slidesPerView: 1,
+			},
+			782: {
 				slidesPerView: config.slidesPerView > 1 ? 2 : 1,
 			},
-			600: {
+			1168: {
 				slidesPerView: config.slidesPerView,
 			},
 		},
--- a/src/blocks/carousel/edit.js
+++ b/src/blocks/carousel/edit.js
@@ -174,6 +174,7 @@ class Edit extends Component {
 		const classes = classnames(
 			className,
 			'wp-block-newspack-blocks-carousel', // Default to make styles work for third-party consumers.
+			'slides-per-view-' + slidesPerView,
 			'swiper',
 			{
 				'wp-block-newspack-blocks-carousel__autoplay-playing': autoplay,
--- a/src/blocks/carousel/view.php
+++ b/src/blocks/carousel/view.php
@@ -29,6 +29,7 @@ function newspack_blocks_render_block_carousel( $attributes ) {
 	if ( $autoplay ) {
 		$other[] = 'wp-block-newspack-blocks-carousel__autoplay-playing';
 	}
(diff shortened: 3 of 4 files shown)


dkoo (approved):
Very nice—these styles look much more responsive than on `master`. 👍 

> I think that's more of a case where we'd want to steer folks away from using particular combinations of settings, rather than something we need to get to work well in all cases

That was my thinking as well. The multiple slides per view was added to accommodate the use case of showing several small sponsor logos at a time. We won't always know what images publishers want to show or in what contexts, so I think the block benefits from the flexibility even if it's possible to create some very bad-looking carousels using the available options.

laurelfulford:
Thanks @dkoo!

> The multiple slides per view was added to accommodate the use case of showing several small sponsor logos at a time.

Oh, TIL! That makes a lot of sense with it showing two slides on smaller screens. Hopefully these changes doesn't mess with how that works too much!

PR #947, opened 2021-11-30
feat: author list block

### LINKED ISSUE #901: New block: Author Profile List
The individual Author Profile block is very handy! But it's a manual placement, which means updating the staff roster page every time someone leaves or is hired. And, more important, it doesn't scale: One Newspack publisher from 40 to 350+ authors in just one year. Placing 350 author blocks and then keeping that list updated isn't feasible.

We should implement a new block that shows all authors in lazy-loaded fashion using the same template and display options as the individual Author Profile block. It would also have sort options (first/last name) and the option to show WordPress users, Co-Authors Plus guest authors, or both.

dkoo:
### All Submissions:

Registers a new Author List block that uses the same template as the existing Author Profile block to display all WP users, guest authors, or both. Also adds a few new display features that apply to both blocks.

Closes #901.

1. Ensure your site has a number of WP users and CAP guest authors, some assigned to posts, others with 0 posts. Ensure that you also have some WP users of various roles, e.g. editor, admin, author, contributor, and subscriber, as well as some custom roles (such as the WooCommerce user roles).
2. Test the new options for the Author Profile block. Add one to a post or page and test the following new options in both editor and front-end:
  - **Text size:** controls the relative size of the text compared to the surrounding post. Should offer S, M, L, and XL options with M being the default.
  - **Hide default avatars:** If the "display avatars" option is enabled, enabling this option will only display avatars for authors that have a custom avatar image.

3. Test the new Author List block. Test the various attributes and display options available, paying special attention to:
  - **Author Type:** CAP guest authors, WP users, or both.
  - **WP user roles:** If the Author Type contains WP users, this should default to core WP user roles that have the `edit_posts` capability (Admin, Editor, Author, and Contributor) but it should have additional options to display other custom user roles registered with that capability.
  - **Column view:** In the block toolbar (not the sidebar). Default view is a simple list, but enabling columns will display the list in columns and enable the Columns attribute.
  - **Columns:** In the sidebar. If column view is enabled, should let you specify 2–6 columns.
  - **Display alphabetical separators:** If enabled, letter headings will separate the authors shown based on their display name (first letter of last name, or first letter of name if the display name is only one word).
  - **Group authors by alphabet:** If column view and "display alphabetical separators" are enabled, this option should be available. If enabled, it will break the authors into sections by alphabetical separator, so that each section becomes its own set of columns.
  - **Author List Exclusions:** Options to exclude authors who have 0 posts as well as specific authors by name.
  - **Author Profile Settings:** Should have the same options and behavior as these settings in the Author Profile block. Settings will apply to all items in the block.
  - **Avatar Settings:** Should have the same options and behavior as these settings in the Author Profile block. Settings will apply to all items in the block.

4. Also test after disabling the Co-Authors Plus plugin—all guest-author related options should be disabled/hidden and the block should only show WP users.

### Other information:


src/blocks/author-list/view.php
```diff
@@ -0,0 +1,175 @@
+<?php
+/**
+ * Server-side render functions for the Author Profile block.
```
adekbadek, line 3: Nit:

```suggestion
 * Server-side render functions for the Author List block.
```

src/blocks/author-list/view.php
```diff
@@ -0,0 +1,175 @@
...
+	?>
+	<div class="wp-block-newspack-blocks-author-list <?php echo esc_attr( $attributes['className'] ); ?>">
+
+	<?php
+	// Columns WITH alphabetical separator sections.
+	if ( $is_columns && $show_separators && $separator_sections ) :
+		// Associative array to store authors chunked by alphabet, if separatorSections is true.
+		$chunked_authors = [];
+		foreach ( $authors as $author ) {
+			$first_letter = strtoupper( substr( $author['last_name'], 0, 1 ) );
```
adekbadek, line 96: A user with login `admin`, without neither first nor last name, ends up in "W" section when chunked alphabetically:

image

This is the last item on the whole list, so maybe the it's a matter of listing un-sortable (nameless) users after the sorted users. This happens both in editor and on the frontend.

src/blocks/author-profile/view.php
```diff
@@ -107,87 +116,15 @@ function newspack_blocks_render_block_author_profile( $attributes ) {
...
-	if ( isset( $attributes['className'] ) ) {
-		$classes .= ' ' . $attributes['className'];
-	}
+	$content = Newspack_Blocks::template_include(
+		'author-profile-card',
+		[
+			'attributes' => $attributes,
+			'author'     => $author,
+		]
+	);
```
adekbadek, line 125: When a guest user has an avatar set, the avatar is **not displayed** on Author Block on the frontend (editor is fine). This does not happen on `master`.

src/templates/author-profile-card.php
```diff
@@ -0,0 +1,99 @@
...
+			$attributes,
+			$extra_classes
+		);
+
+		?>
+		<div class="<?php echo esc_attr( $classes ); ?>">
+			<?php if ( $attributes['showAvatar'] && isset( $author['avatar'] ) ) : ?>
+				<div class="wp-block-newspack-blocks-author-profile__avatar">
+					<figure style="border-radius: <?php echo esc_attr( $attributes['avatarBorderRadius'] ); ?>; width: <?php echo esc_attr( $attributes['avatarSize'] ); ?>px;">
+						<?php echo wp_kses_post( $author['avatar'] ); ?>
```
adekbadek, line 44: Hard to pin down the conditions (aspect ratio, size?), but when using this image, it comes out stretched: 

image

Image:

image

adekbadek (changes requested):
I imagine this will be a very useful features, not only for news publications 💪 
Gotta point out though that it's intimidating to review such a large PR 😬

src/blocks/author-list/class-wp-rest-newspack-author-list-controller.php
```diff
@@ -0,0 +1,366 @@
...
+		$options         = wp_parse_args( $options, $default_options );
+		$fields          = $options['fields'];
+		$current_page    = 1;
+
+		// Array to store all authors.
+		$all_guest_authors = [];
+		$all_users         = [];
+
+		// Get Co-authors guest authors, only if CAP plugin is active and the author type specified includes guest authors.
+		if ( 'users' !== $options['author_type'] && class_exists( 'CoAuthors_Guest_Authors' ) ) {
```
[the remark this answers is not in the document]
dkoo: Most likely it won't, but this is a good suggestion for readability. Done on lines 159 and 186 in 2f972bb.

src/blocks/author-list/edit.js
```diff
@@ -0,0 +1,523 @@
...
+					) }
+				</PanelBody>
+				<PanelBody title={ __( 'Author List Exclusions', 'newspack-block' ) }>
+					<PanelRow>
+						<ToggleControl
+							label={ __( 'Exclude authors with 0 posts', 'newspack-blocks' ) }
+							help={ sprintf(
+								// Translators: Help message for "include empty authors" toggle.
+								__( '%s authors with no published posts.', 'newspack-blocks' ),
+								excludeEmpty ? __( 'Hide', 'newspack-blocks' ) : __( 'Show', 'newspack-blocks' )
```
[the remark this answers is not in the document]
dkoo: Updated this and several other help messages in d586a1b.

src/blocks/author-list/edit.js
```diff
@@ -0,0 +1,523 @@
...
+								checked={ separatorSections }
+								onChange={ () => setAttributes( { separatorSections: ! separatorSections } ) }
+							/>
+						</PanelRow>
+					) }
+				</PanelBody>
+				<PanelBody title={ __( 'Author List Exclusions', 'newspack-block' ) }>
+					<PanelRow>
+						<ToggleControl
+							label={ __( 'Exclude authors with 0 posts', 'newspack-blocks' ) }
```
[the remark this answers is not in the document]
dkoo: This was a tricky one to debug. Investigating it uncovered a bug in Co-Authors Plus! When a guest author has a linked WP user the [`get_author_post_count` method](https://github.com/Automattic/Co-Authors-Plus/blob/3f71c41f3628bd5b848c401195798cf95f994c3b/co-authors-plus.php#L1763) only returns the post count of the linked WP user, not the sum of both guest author and user posts. 

Because of the way our block code is searching for WP users using `has_published_posts` as a `WP_User_Query` arg, then combining those results with a guest author query using `CoAuthors_Plus::get_author_post_count`, this bug results in the guest author getting dropped when "exclude authors with 0 posts" is on if either the WP user OR the guest author has 0 posts. 

This should be fixed in b48c67d, which foregoes the use of `CoAuthors_Plus::get_author_post_count` and just manually adds the post count of the guest author and the WP user to get the total post count.

This does bring up another potentially confusing edge case, where if a WP user with 0 posts is linked to a guest author with many posts, and the Author List block is set to display only WP users. In this case, displaying only WP users will display the WP user, not the linked guest author, and since the WP user has 0 posts, it won't be displayed if "exclude authors with 0 posts" is on. This, to me, is expected behavior even if it sounds confusing.

src/blocks/author-list/view.php
dkoo: Ah, this is because I was assuming all users would have at least a display name, which isn't always the case. 48e5e7d adds another fallback to alphabetize by user login if there's no last name or display name.

src/templates/author-profile-card.php
dkoo: This is happening on `master`, too. Seems to be a bug with the CSS for the avatars. I'll see if I can fix on this branch.

src/blocks/author-profile/view.php
dkoo: Good catch! 4c58457 should fix this.

src/templates/author-profile-card.php
dkoo: d498c8f should fix this. The aspect ratio of the `figure` should be fixed to 1:1 with the image inside set to `object-fit: cover`, but I forgot to set both the height and the width (only the width was set, so height was defaulting to `auto`).

src/blocks/author-list/edit.js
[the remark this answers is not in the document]
adekbadek: This makes sense, thank you for the detailed response!

src/blocks/author-list/edit.js
```diff
@@ -0,0 +1,525 @@
...
+						<ToggleControl
+							label={ __( 'Display avatar', 'newspack-blocks' ) }
+							checked={ showAvatar }
+							onChange={ () => setAttributes( { showAvatar: ! showAvatar } ) }
+						/>
+					</PanelRow>
+					{ showAvatar && (
+						<PanelRow>
+							<ToggleControl
+								label={ __( 'Hide default avatar', 'newspack-blocks' ) }
```
adekbadek, line 335: This seems to handle Guest Authors only – after toggling this option, WP users are still listed with the default avatars.

src/blocks/author-list/view.php
adekbadek: The issue is still happening. In the examples below, `Robers Wpuser` & `admin` are both WP users:

1. If set to display all authors, `admin` is at the end, without a letter "A" preceding (there's a guest user already in an "A" section:

image

2. If set to display WP users only, `admin` is again at the end, but in an "A" section (there are no other WP users with names starting on "a"):

image

src/blocks/author-list/edit.js
dkoo: Hmm, the logic applies to both author types, but it seems like the `avatar-default` class isn't always added to default avatars for some reason. b98c749 adds a bit of extra logic which should hopefully catch those instances.

src/blocks/author-list/view.php
dkoo: f8532a6 fixes this. The issue was happening because we weren't standardizing the case of the letters, so `A` and `a` were sorted differently.

adekbadek (approved):
All good now 👌

PR #949, opened 2021-12-01
Add/embed pdf

### LINKED ISSUE #897: Handle embedding PDFs on the Iframe Block
We need an AMP-compatible way to embed PDFs. The Iframe Block enables us to embed iframes, it can fit this too.

Embed PDFs using the Iframe Block. AMP has a nice tag to be used for this https://amp.dev/documentation/components/amp-google-document-embed/

The “File” block type doesn’t work using URLs (you have to add the file to the Media Library instead)

kariae:
### All Submissions:

Closes #897. Allow embedding documents (PDF, Doc, Excel, and PPT) with the Iframe block.

1. Add an Iframe block, and upload a PDF file as the iframe source.
2. Notice that the PDF is previewed in the editor.
3. Load the page in the front and notice that the PDF file is embedded.

### Other information:

This uses the Google Doc Viewer to embed the documents which is the same way is used when using the [amp-google-document-embed](https://amp.dev/documentation/components/amp-google-document-embed/) tag. The issue with the Google Doc Viewer is that [1 or 2 times out of 10 it doesn't load in iframes](https://stackoverflow.com/questions/40414039/google-docs-viewer-returning-204-responses-no-longer-working-alternatives). A solution for this in `non-amp` pages is to [keep reloading the iframe until its content is loaded](https://github.com/Automattic/newspack-blocks/pull/949/files#diff-e31b7d811bd799e3d22782e7babd1f230b2c28feb2dc5d14976a91e5046ac056R107-R120) which seems to do the trick. But unfortunately, we can't have on AMP since[`onload` is dismissed from the amp-iframe tag](https://github.com/ampproject/amphtml/blob/main/extensions/amp-iframe/1.0/component.js#L25).

A first alternative would be to use Microsoft Office Viewer but it doesn't support PDFs.

Another alternative is to use something like [pdf.js](https://github.com/mozilla/pdf.js) but I'm not sure how it will behave with AMP.


adekbadek:
@kariae – please resolve the conflict when you have a moment

kariae:
Done, thanks @adekbadek.

adekbadek (approved):
😎

PR #954, opened 2021-12-03
fix: view scripts in AMP Plus mode

### LINKED ISSUE #953: Posts block: "Load more" button doesn't work when AMP Plus is enabled
**Describe the bug**

On a site using AMP Plus, the "Load more" button does not do anything when clicked.

**To Reproduce**
1. Set up a posts block with the Load More button on a Newspack site.
2. In your wp-config, add `define('NEWSPACK_AMP_PLUS_ENABLED', true);`
3. On the frontend, click the Load More button. Observe nothing happens. No error is thrown in the JS console.
4. In your wp-config, disable the `NEWSPACK_AMP_PLUS_ENABLED` flag from step 2.
5. On the frontend, click the Load More button. Observe it works as normal.

**Expected behavior**
I expect the Load More button to work regardless of AMP status. I'm not sure the best solution for this situation, though -- make the button use the non-AMP handling if AMP Plus is enabled?

adekbadek:
### All Submissions:

Closes #953

Not all view scripts (JS loaded on the frontend) were handled in AMP Plus mode. This PR fixes that, ~so that the blocks' interactions in AMP Plus mode are as they do with AMP disabled.~ 
_Initially, this PR forced all blocks into behaving as with AMP disabled in AMP Plus. There's no need for this, though. The only block that really needs to use AMP Plus features is the streamlined version of the Donate block, because it needs to load additional scripts._

1. Observe that #953 is not reproducible
2. Confirm that Carousel block's frontend JS-reliant interactions work as expected with AMP Plus mode on, and exactly as with AMP enabled

### Other information:

---
  - https://app.asana.com/0/0/1200994000430649

### DIFF
--- a/includes/class-newspack-blocks.php
+++ b/includes/class-newspack-blocks.php
@@ -290,9 +290,6 @@ class Newspack_Blocks {
 	 * @return bool True if AMP, false otherwise.
 	 */
 	public static function is_amp() {
-		if ( class_exists( 'Newspack\AMP_Enhancements' ) && method_exists( 'Newspack\AMP_Enhancements', 'should_use_amp_plus' ) && Newspack\AMP_Enhancements::should_use_amp_plus( 'gam' ) ) {
-			return false;
-		}
 		if ( function_exists( 'is_amp_endpoint' ) && is_amp_endpoint() ) {
 			return true;
 		}
--- a/includes/class-newspack-blocks.php
+++ b/includes/class-newspack-blocks.php
@@ -9,6 +9,12 @@
  * Newspack blocks functionality
  */
 class Newspack_Blocks {
+
+	/**
+	 * Script handle for the streamlined donate block script.
+	 */
+	const DONATE_STREAMLINED_SCRIPT_HANDLE = 'newspack-blocks-donate-streamlined';
+
 	/**
 	 * Add hooks and filters.
 	 */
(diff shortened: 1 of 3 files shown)


dkoo (changes requested):
A couple of issues:

1. When in AMP Plus mode, images in posts loaded after clicking "load more" are rendered with incorrect sizing. This doesn't happen when AMP is turned off for the same page. We may have to add code in the non-AMP JS to manually size the images to get around AMP's image sizing behavior.

Screen Shot 2021-12-03 at 11 39 26 AM

2. Also when in AMP Plus mode, after clicking "Load more" the "loading" visual state is not shown. There's no indication of any action occurring after clicking the button, until the new posts appear.

[Example page](https://thin-marlin.jurassic.ninja/) where this is happening.

dkoo:
Another issue with the Post Carousel block: in AMP Plus mode, the images are not visible at all because the post-page-load resizing doesn't seem to be working. Maybe due to similar reasons as the sizing issue in Homepage Posts described above.

adekbadek:
I've changed the approach here – see updated PR description.

dkoo (approved):
This works with the reduced scope! However, sooner or later we'll have to figure out how to run non-AMP JS scripts on AMP pages in an AMP Plus environment, to really take advantage of AMP Plus.

PR #979, opened 2021-12-14
fix: validation errors for invalid attributes on amp-base-carousel

dkoo:
### All Submissions:

The Post Carousel block renders an `amp-base-carousel` on AMP-enabled pages, but several attributes in our markup have missing values or just aren't valid attributes for that AMP element. This means that any pages with a Post Carousel will fail AMP validation.

1. On `master`, add a Post Carousel block to a post.
2. In the sidebar, expand AMP and click "Open AMP Validation" then "Re-validate" to trigger AMP validation for the post content.
3. Observe several errors (particularly the `controls`, `loop` and `type` attributes) relating to the Post Carousel like so:

Screen Shot 2021-12-14 at 4 06 46 PM

4. Check out this branch, re-validate for AMP again. Confirm that the errors related to Post Carousel are no longer reported.

### Other information:

### DIFF
--- a/src/blocks/carousel/view.php
+++ b/src/blocks/carousel/view.php
@@ -261,7 +261,7 @@ function newspack_blocks_render_block_carousel( $attributes ) {
 		);
 
 		$carousel    = sprintf(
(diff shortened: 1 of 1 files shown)


laurelfulford (approved):
This works as described, and gets rid of the errors with the current version of AMP! 

I also tested this PR with AMP 2.0.11, since we have a few sites running it -- that version now throws an error for `controls="auto"`, but it doesn't affect how the slideshow works. So I think overall this is safe to move forward with! 🚢

dkoo:
> I also tested this PR with AMP 2.0.11, since we have a few sites running it -- that version now throws an error for `controls="auto"`, but it doesn't affect how the slideshow works. So I think overall this is safe to move forward with! 🚢 

That's annoying! They must have changed the standards between versions, I'm guessing? So because we can't support both standards, we'll have to go with the most widely used one, which is the latest version, right?

laurelfulford:
> So because we can't support both standards, we'll have to go with the most widely used one, which is the latest version, right?

That was my thinking, too! :)

PR #983, opened 2021-12-15
Post Carousel: add simplified CSS class to carousel to reduce style size

laurelfulford:
### All Submissions:

This PR adds a simplified CSS class to the Post Carousel block (`wpnbpc`, to replace the longer `wp-block-newspack-blocks-carousel` class in the CSS. 

This helps reduce the amount of CSS loaded by AMP, especially since this class is included in every style associated with this block -- something similar was done with the Homepage Posts block early on. In my testing, it reduces the CSS by 1.7% when the controls are visible:

With controls before changes: 7,183b = 9.6% of 75kb
With controls after changes: 5,996b = 7.9% of 75kb

There's probably more that can be done here but it's a start!

1. Set up a Post Carousel block, and check your AMP CSS usage.
2. Apply the PR and run `npm run build.`
3. Clear your browser cache; make sure the post carousel looks right with and without AMP enabled, on the front-end and in the editor. 
4. Re-check your AMP CSS usage; confirm it's gone done ~1.5ish percent. 

### Other information:

### DIFF
--- a/src/blocks/carousel/edit.js
+++ b/src/blocks/carousel/edit.js
@@ -174,6 +174,7 @@ class Edit extends Component {
 		const classes = classnames(
 			className,
 			'wp-block-newspack-blocks-carousel', // Default to make styles work for third-party consumers.
+			'wpnbpc', // Shortened version of the default classname.
 			'slides-per-view-' + slidesPerView,
 			'swiper',
 			{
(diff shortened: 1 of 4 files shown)


adekbadek (approved):
Works fine, but I wonder why the new class is _added_, rather than replacing the `wp-block-newspack-blocks-carousel` class?

laurelfulford:
Thanks @adekbadek! 

> I wonder why the new class is added, rather than replacing the wp-block-newspack-blocks-carousel class?

Strictly out of caution! :) There's no knowing if there's a site out there already using the `wp-block-newspack-blocks-carousel` class to style the block, so while it's probably unlikely, leaving it in place hopefully means we don't accidentally break anything with this change!

PR #985, opened 2021-12-16
feat(donate-stripe): hide fee checkbox if fees are zero

adekbadek:
### All Submissions:

Allows hiding of cover-fees checkbox (#928) if the fees amount to 0.

_Follow steps in https://github.com/Automattic/newspack-plugin/pull/1376_

### Other information:

### DIFF
--- a/src/blocks/donate/streamlined.js
+++ b/src/blocks/donate/streamlined.js
@@ -135,7 +135,7 @@ export const processStreamlinedElements = ( parentElement = document ) =>
 			if ( feesAmountEl ) {
 				const formValues = Object.fromEntries( new FormData( formElement ) );
 				const feeAmount = getFeeAmount( formElement );
-				feesAmountEl.innerHTML = `(${ CURRENCY_SYMBOL }${ feeAmount } ${ FREQUENCIES[
+				feesAmountEl.innerHTML = `(${ CURRENCY_SYMBOL }${ feeAmount.toFixed( 2 ) } ${ FREQUENCIES[
 					formValues.donation_frequency
 				].toLowerCase() })`;
 			}
--- a/src/blocks/donate/streamlined.js
+++ b/src/blocks/donate/streamlined.js
@@ -60,6 +60,12 @@ const getAmount = formValues => {
 	return parseFloat( formValues.amount );
 };
 
+export const computeFeeAmount = ( amount, feeMultiplier, feeStatic ) => {
+	return parseFloat(
+		( ( ( amount + feeStatic ) / ( 100 - feeMultiplier ) ) * 100 - amount ).toFixed( 2 )
+	);
+};
+
 const getFeeAmount = formElement => {
 	const formValues = Object.fromEntries( new FormData( formElement ) );
 	const amount = getAmount( formValues );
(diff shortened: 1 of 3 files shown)


src/blocks/donate/streamlined.js
```diff
@@ -60,19 +60,20 @@ const getAmount = formValues => {
 	return parseFloat( formValues.amount );
 };
 
+export const computeFeeAmount = ( amount, feeMultiplier, feeStatic ) => {
+	return parseFloat(
```
dkoo, line 64: We're using `toFixed( 2 )` to control the number of digits that appear in the fee amount, but `parseFloat` will subsequently drop trailing zeroes anyway, so a number like `5.30` will get printed as `5.3` on the front-end:

Screen Shot 2022-01-03 at 1 07 13 PM

This could be solved by using `toString()` on the return value instead, which will ensure that the amount will always get printed with two decimals.

But even better, would it be possible to use the site's active currency plus [`toLocaleString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString) to render the amount in the proper currency format? This might be overkill but it would ensure that the formatting always matches the active currency.

src/blocks/donate/streamlined.js
```diff
@@ -60,19 +60,20 @@ const getAmount = formValues => {
...
+		( ( ( amount + feeStatic ) / ( 100 - feeMultiplier ) ) * 100 - amount ).toFixed( 2 )
```
dkoo, line 65: I think this math is slightly off, if I'm not mistaken. Are the default values (`2.9` and `0.3`) coming from Stripe's [integrated pricing](https://stripe.com/pricing)? If so, shouldn't this calculation be:

```suggestion
		( amount * ( feeMultiplier / 100 ) + feeStatic ).toFixed( 2 )
```

Given a donation amount of $100, the fee should be:

Multiplier: 2.9% of $100 = $2.90
Static: $0.30
Total fee: $3.20

Right?

src/blocks/donate/streamlined.test.js
```diff
@@ -42,6 +42,21 @@ const createDOM = settings => {
 	return document.body;
 };
 
+describe( 'fee computation', () => {
+	it( 'computes fee with default values', () => {
+		expect( computeFeeAmount( 1, 2.9, 0.3 ) ).toBe( 0.34 );
```
dkoo, line 47: These test values will need to be updated along with the `computeFeeAmount` function.

dkoo (changes requested):
As I was testing this I think I caught a math error in the function computing the fee amount. Also have a suggestion on rendering the fee on the front-end as a formatted currency string instead of as a float.

src/blocks/donate/streamlined.js
adekbadek: Fixed in 57aabfcba58e83de5bffafa0ad07e3b77eb7a019. I think using `toLocaleString` is an overkill – and also on the PHP side the currency symbol is used, so that's more coherent.

src/blocks/donate/streamlined.js
dkoo: Aha, that makes sense. Thanks for explaining.

dkoo (approved):
Changes look good 👍

Issue #1019, opened 2022-01-17
Ability to not stack columns block on mobile doesn't work with border styles 


laurelfulford:
In WordPress 5.9, there is now an option not to stack the columns block on mobile.

While I don't think this needs to work with our column reordering styles (perhaps they can only be applied when this is not turned on), it needs to work with our column border styles -- right now, the borders move to the bottom even when the columns are not stacked: 

![image](https://user-images.githubusercontent.com/177561/149816252-9a762942-47e8-4e9b-9efb-c9aa2b30861e.png)

PR #1020, opened 2022-01-17
Update Column Block styles for WordPress 5.9.

### LINKED ISSUE #1018: Column block spacing changed in 5.9
This may be something we need to address in the theme as well as the blocks:

In WP 5.9 the spacing on the column block changed.

This is noticable because the gutter between columns has increased. 

It also causes the columns to be slightly "offset" when you use the block styles to reorder the columns on desktop - the first column is too far to the left. In the screenshot below, the second column block has the 'Move first column to second' style applied:

image

Edited to add: the border style also looks a bit "off" with this update -- the border 

image

image

laurelfulford:
### All Submissions:

In the Newspack Blocks suite, we have some column block styles that:

1. Update the default column block styles, to not wrap until they hit the mobile breakpoint.
2. Allow you to add borders to the columns.
3. Allow you to reorder the columns on desktop.

Changes coming in WP 5.9 to add the option to not wrap columns on mobile have overridden some of our own styles. This PR hopefully sorts them out and gets them working again, without causing issues in 5.8.

Closes #1018 and #1019

This will be easiest to test if you can work with two test sites and compare them: one running WordPress 5.8.X, and the other running the 5.9 RC.

1. In the test site running 5.9 RC, copy [this test content](https://github.com/Automattic/newspack-blocks/files/7885188/columns-59.txt) to the editor, or add a bunch of column blocks with different settings. Change the page you test to use the one column wide template.
2. In the test site running 5.8.X, copy [this test content](https://github.com/Automattic/newspack-blocks/files/7885189/columns-58.txt) to the editor, or add a bunch of column blocks with different settings. Change the page you test to use the one column wide template.
3. Apply this PR to both sites and run `npm run build`.
4. Apply this theme PR to both sites and run `npm run build` - it updates a couple styles in the theme: https://github.com/Automattic/newspack-theme/pull/1670
5. View and compare both posts at different breakpoints. Things to look for include:

* Making sure the columns block doesn't wrap on mobile when its set not to on the 5.9 site.
* That the columns styles and spacing look consistent between the different sections.
* That the border styles and spacing look consistent between the different sections.

Here are some examples of how things should look:

**WordPress 5.9**

Regular columns:
image

"Don't stack on mobile" setting (the first column doesn't have this checked, and the second one does)

image

Reordered blocks: 

image

Columns blocks with borders: 

image

**WordPress 5.8**

Regular columns:
image

Reordered columns:

image

Columns with borders: 

image

6. Review things in the editor and make sure things look fairly consistent between the two versions.

### Other information:

### DIFF
--- a/src/block-styles/core/columns/editor.scss
+++ b/src/block-styles/core/columns/editor.scss
@@ -40,7 +40,7 @@
 					bottom: 15px;
 					content: '';
 					position: absolute;
-					right: -20px;
+					right: -18px;
 					top: 15px;
 				}
 
--- a/src/block-styles/core/columns/editor.scss
+++ b/src/block-styles/core/columns/editor.scss
@@ -37,11 +37,11 @@
 
 				&::after {
 					border-right: 1px solid $color__border;
-					bottom: 0;
+					bottom: 15px;
 					content: '';
 					position: absolute;
-					right: -30px;
-					top: 0;
+					right: -20px;
+					top: 15px;
 				}
 
 				&:nth-child( odd )::after {
--- a/src/block-styles/core/columns/editor.scss
+++ b/src/block-styles/core/columns/editor.scss
@@ -3,7 +3,7 @@
 
 .wp-block-columns {
 	@include media( mobile ) {
-		&[class*='is-style-first-col-to'] [data-type='core/column']:first-child {
+		&[class*='is-style-first-col-to'] [data-type='core/column'] {
 			margin-left: 2em;
 		}
 
--- a/src/block-styles/core/columns/editor.scss
+++ b/src/block-styles/core/columns/editor.scss
@@ -3,18 +3,15 @@
 
 .wp-block-columns {
 	@include media( mobile ) {
-		&[class*='is-style-first-col-to'] [data-type='core/column'] {
-			margin-left: 46px;
+		&[class*='is-style-first-col-to'] [data-type='core/column']:first-child {
+			margin-left: 2em;
 		}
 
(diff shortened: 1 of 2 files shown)


miguelpeixe (changes requested):
I can still see some significant differences between 5.9 and 5.8, here's what I've identified:

| 5.9 RC2 | 5.8 | Comment |
| - | - | - |
| image | image | Grid is not behaving the same on equal viewport |
| image | image | Difference in grid also causing a border out of place |
| image | image | Flex gap disrupting border alignment |
| image | | Column containers have bigger margin between them, which doesn't occur on 5.8 |

dkoo:
I'm not seeing the same thing as @miguelpeixe. I found it difficult to compare apples to apples with two different sets of content, so what I did was to paste both sets of content into a single post on both 5.8 and 5.9 sites.

* [5.9 test site](https://vivacious-shark.jurassic.ninja/2022/01/18/columns-test/)
* [5.8 test site](https://equivalent-lynx.jurassic.ninja/2022/01/18/columns-test/)

With both sites using the same content, it's easier to see that the styles match well across both WP versions. The main differences I see are:

- The "Don't stack on mobile" columns do stack in WP 5.8 (as expected, since this is a new feature in 5.9)
- Slight differences in spacing in the editor, but not on the front-end

@miguelpeixe can you review these two test sites and let us know what you see?

miguelpeixe (approved):
I'm sorry about that! I was not on 5.9 RC2's Gutenberg. I now can confirm that the columns are behaving as expected!

laurelfulford:
Thank you @miguelpeixe and @dkoo! 🙌

PR #1026, opened 2022-01-24
fix: handle unavailable Newspack plugin

adekbadek:
### All Submissions:

1. Handles the case of Newspack plugin unavailable
2. Adds `.cache` directory (npm-related) to `.distignore` file, so it won't be included in the released ZIP 

1. On `master`, deactivate the Newspack Plugin 
2. Try to add a new page in Newspack admin, observe a PHP fatal `Fatal error:  Uncaught Error: Class 'Newspack\Donations' not found` thrown
3. Switch to this branch, observe no fatal thrown

### Other information:

### DIFF
--- a/.distignore
+++ b/.distignore
@@ -34,3 +34,4 @@ node_modules
 /release
 /tests
 /bin
+.cache
--- a/includes/class-newspack-blocks.php
+++ b/includes/class-newspack-blocks.php
@@ -1089,9 +1089,8 @@ class Newspack_Blocks {
 	 * Disable Jetpack's donate block when using Newspack donations.
 	 */
 	public static function disable_jetpack_donate() {
-
-		// Do nothing if Jetpack's blocks aren't being used.
-		if ( ! class_exists( 'Jetpack_Gutenberg' ) ) {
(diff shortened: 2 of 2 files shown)


dkoo (changes requested):
Nice, I had noticed this when testing but assumed it would be a more involved fix.

Suggestion: we should avoid registering the Donate block entirely if the `Newspack` class doesn't exist.

adekbadek:
> Suggestion: we should avoid registering the Donate block entirely if the Newspack class doesn't exist.

That's already in place :) 

https://github.com/Automattic/newspack-blocks/blob/0a374cdc328c3b39ed161e0842c038ecfca79e85/src/blocks/donate/view.php#L133-L135

adekbadek:
> Suggestion: we should avoid registering the Donate block entirely if the Newspack class doesn't exist.

The block will [not render when Newspack plugin is not active](https://github.com/Automattic/newspack-blocks/blob/0a374cdc328c3b39ed161e0842c038ecfca79e85/src/blocks/donate/view.php#L133-L135). Handling this in the editor is a good idea, in principle, but not worth the effort currently IMO.

PR #1035, opened 2022-02-02
Add/stripe payment request button

adekbadek:
### All Submissions:

Adds support for [Stripe's payment request button](https://stripe.com/docs/stripe-js/elements/payment-request-button) in the streamlined Donate block.

~One major design decision here is that the submit button in the Donate block (not only streamlined) will have the primary color as the background color.~ _Not anymore, see comments_

---

Here's how it looks like:

Google Pay on an Android phone:

image

Apple Pay on an iPhone:

image

On a desktop computer, with a saved payment method (Chrome browser):

image

On a desktop computer, without any payment methods available:

image

In WP editor:

image

1. Use `add/stripe-location-code` branch of `newspack-plugin` (https://github.com/Automattic/newspack-plugin/pull/1483)
2. Configure Stripe as the Reader Revenue platform*
3. The code here is not payment-proxy-specific, so testing just one of the available ones is enough – the simplest is probably to add a card to Chrome browser. Make a donation and verify it did go through in Stripe (if Stripe test instance is used, the card won't be really charged).
4. For a good measure, also make sure the card flow is working (use `4242 4242 4242 4242` as the card)   

\* Setting `NEWSPACK_AMP_PLUS_ENABLED` env. variable to `true` is a prerequisite

### Other information:


thomasguillot:
> It just made sense to me to have the CTA in the more prominent color, but maybe using the secondary color here was intentional? cc @thomasguillot

I don't know 🙂  but @laurelfulford should know.

laurelfulford:
> It just made sense to me to have the CTA in the more prominent color, but maybe using the secondary color here was intentional?

@adekbadek To be honest, I think this choice was a bit arbitrary :) the theme does use the secondary button by default for the buttons, and that's picked up by the donate block, but it's not being explicitly set on that block for that block. 

The secondary wouldn't always be less prominent -- though in the theme defaults it _definitely_ is, so I get where you're coming from here! @yogeshbeniwal's suggestion of having some kind of control to override the button's colour might be the best middle ground: it lets you make sure the colour is very prominent, and lets you get outside of your existing palette if you'd like -- it also makes sure we don't accidentally change how something looks on a live site if it's already set up the way someone likes it.

adekbadek:
I've reverted the button color change in 793a7b2da2b85ca407f063134ed748c8c2500883 (it's out of scope of this PR anyway). The button color should be customisable, but that can now be done via setting theme colors or custom CSS, so there's no need for this change. The button color will be as in other button on the site, and as @laurelfulford points out, it only really makes sense for the default theme, and the theme colors are probably always changed from the defaults.

src/blocks/donate/edit.js
```diff
@@ -319,14 +319,20 @@ class Edit extends Component {
...
 		return (
-			<button type="submit" onClick={ evt => evt.preventDefault() }>
-				<RichText
-					onChange={ value => setAttributes( { buttonText: value } ) }
-					placeholder={ __( 'Button text…', 'newspack-blocks' ) }
-					value={ buttonText }
-					tagName="span"
-				/>
-			</button>
+			<>
```
miguelpeixe, line 322: This fragment is superfluous since there's only one child element.

src/blocks/donate/streamlined/utils.js
```diff
@@ -0,0 +1,151 @@
+/**
+ * WordPress dependencies
+ */
+import { __, sprintf } from '@wordpress/i18n';
+
+const isValidEmail = string => /\S+@\S+/.test( string );
+export const validateFormData = values => {
```
miguelpeixe, line 7: Non-blocking: it would be helpful to have inline docs for these utils functions.

miguelpeixe (changes requested):
The flow is working with the regular card (`4242 4242 4242 4242`) and with GPay on a desktop Chrome.

I was not able to complete the flow with Apple or Android. While on an iPhone I got the Apple Pay flow started but immediately terminated without any messages. On an Android I got the following error after confirming the payment prompt: 

image

Maybe these errors are related to the test environment?

src/blocks/donate/edit.js
adekbadek: Done in dbc75df9d0f769155b928b651df006668aadd8eb

src/blocks/donate/streamlined/utils.js
adekbadek: Done in 4cffbdab62f50f56b6150242a256d824928a5a5c

adekbadek:
@miguelpeixe – 

> I was not able to complete the flow with Apple or Android. 

I've updated the Plugin PR to error when customer could not be created – you should now see a more meaningful Stripe error after failure.

miguelpeixe (approved):
Thank you for the revisions!

miguelpeixe (approved):
👌 All payment methods are working as expected!

PR #1045, opened 2022-02-16
fix(NextGen Gallery): disable ngg shortcode manager

### LINKED ISSUE #665: Conflict with NextGEN Gallery showing "The response is not a valid JSON response."
**Describe the bug**
Basically when we have the block "Blog Posts", it shows a generic "The response is not a valid JSON response." error as long as you have the [nextgen-gallery](https://wordpress.org/plugins/nextgen-gallery/) plugin is activated.

Reported in zen-3633315

**To Reproduce**
1. Install [NextGEN Gallery](https://wordpress.org/plugins/nextgen-gallery/)
2. Add the Blog Posts block on a page
3. If you still don't see the error, try saving/updating the page, and it should show "The response is not a valid JSON response."

**Screenshots**
https://d.pr/i/Thp1is

**Expected behavior**
It should work as expected without bringing the error "The response is not a valid JSON response."

miguelpeixe:
### All Submissions:

Fix the NextGen Gallery and Homepage Posts block conflict by disabling NextGEN's `C_NextGen_Shortcode_Manager`. The way it currently parses `the_content` conflicts with the REST API request to save a post containing a Homepage Posts block. This is due to how it uses output buffering through `ob_start()` on REST requests.

https://plugins.trac.wordpress.org/browser/nextgen-gallery/tags/3.23/non_pope/class.nextgen_shortcode_manager.php#L193

Closes #665.

Reproduce the steps mentioned in #665 with this branch and confirm the post saves without errors.

### Other information:

### DIFF
--- a/includes/class-newspack-blocks.php
+++ b/includes/class-newspack-blocks.php
@@ -25,6 +25,19 @@ class Newspack_Blocks {
 		add_filter( 'script_loader_tag', [ __CLASS__, 'mark_view_script_as_amp_plus_allowed' ], 10, 2 );
 		add_action( 'jetpack_register_gutenberg_extensions', [ __CLASS__, 'disable_jetpack_donate' ], 99 );
 		add_filter( 'the_content', [ __CLASS__, 'hide_post_content_when_iframe_block_is_fullscreen' ] );
+
+		/**
+		 * Disable NextGEN's `C_NextGen_Shortcode_Manager`.
+		 *
+		 * The way it currently parses `the_content` conflicts with the REST API
+		 * request to save a post containing a Homepage Posts block. This is due to
+		 * how it uses output buffering through `ob_start()` on REST requests.
+		 *
+		 * @link https://plugins.trac.wordpress.org/browser/nextgen-gallery/tags/3.23/non_pope/class.nextgen_shortcode_manager.php#L193.
+		 */
+		if ( ! defined( 'NGG_DISABLE_SHORTCODE_MANAGER' ) ) {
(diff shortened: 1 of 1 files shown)


dkoo (approved):
I'm not 100% clear on how this affects behavior in the conflicting plugin, but can confirm that it fixes the API request error and allows saving posts.

miguelpeixe:
It's not clear to me why NextGEN Gallery is buffering the output while filtering `the_content`. You can reproduce the same bug by adding this filter:

```php
add_filter(
	'the_content',
	function( $content ) {
		ob_start();
		return $content;
	} 
);
```

dkoo:
> It's not clear to me why NextGEN Gallery is buffering the output while filtering `the_content`.

Exactly—with your filter active, I was still able to create NextGEN galleries and saw no difference in behavior before vs. after. 🤷

miguelpeixe:
Their "shortcode manager" serves only to temporarily replace the plugin's gallery with a placeholder to protect it from other filters. Disabling the usage of the class with this constant exposes the gallery to this risk. From the class documentation:

```
 * NGG is used with a wide variety of third-party themes and plugins. Some of which, do things
 * a little unorthadox, and end up modifying the markup generated by the shortcode, which breaks
 * the display for our users.
 *
 * Rather than having to explain to our userbase, "you shouldn't use NGG with [plugin x]", we
 * decided to find a mechanism that would make our plugin immune from such manipulation from other
 * plugins, or at least reduce likelihood of such a thing happening.
```

Still, don't know why they'd need to buffer the output 😅

PR #1050, opened 2022-02-23
iFrame Block: Add alignment classes and update options

### LINKED ISSUE #1046: iFrame Block: Supports alignments, but doesn't add classes to the front-end
**Describe the bug**
The iFrame block includes the option to change the block alignment, but it doesn't appear to be adding the classes on the front-end (alignwide, alignfull), so the styles are not working there. The previews do work in the editor, though.

**To Reproduce**
1. Add a couple iframe blocks to the editor.
2. Make them align wide and align full.
3. Publish the post. 

**Expected behavior**
Though the blocks preview as wide and full in the editor, they do not appear that way on the front-end. When you inspect source, the 'alignwide' and 'alignfull' classes are not being added to the block; on testing, when they are, the block appears to work as expected.

laurelfulford:
### All Submissions:

This PR makes two changes to the iframe block: 

1. It makes sure that the alignment classes (alignwide, alignfull...) are added to the front-end of the site -- multiple alignments were available, but the styles weren't actually being applied. 
2. It pares down the alignment options from all (left, right, center, wide and full) to just wide and full. 

I am definitely open to feedback about the second change! My feeling is that the left, right and (to a lesser extent) centre alignments are less useful options for something like an embed. 

On top of that, none of them really work even with the right classes applied to the front-end: if an iframe is aligned left or right, it needs a fixed px width set so it doesn't "collapse" due to how the theme styles left and right-aligned blocks. If the iframe is centre aligned, even if it's made narrower (like 300px) it will sit to the left because the theme is apply the centre-aligned styles to the parent div, not the contents. If we do want to keep these alignments we'll need to pair this PR with one in the theme to get everything working.

Closes #1046

1. Create a new post or page and switch it to one of the one column template options. (Note: if switching to one of these templates on page creation, I'd recommend publishing and refreshing the editor, just so the right editor styles are loaded).
2. Add three iframe blocks, and make one 'wide' and one 'full' aligned.
3. Publish and view on the front-end. 
4. Note that while your wide and full alignment previewed fine in the editor, they don't work on the front-end.
6. Refresh your post/page; confirm that your alignments now work on the front end. 

### Other information:

### DIFF
--- a/src/blocks/iframe/index.js
+++ b/src/blocks/iframe/index.js
@@ -41,7 +41,7 @@ export const settings = {
 	attributes,
 	supports: {
 		html: false,
-		align: true,
+		align: [ 'wide', 'full' ],
 	},
 	edit,
 	save: () => null, // to use view.php
--- a/src/blocks/iframe/view.php
+++ b/src/blocks/iframe/view.php
@@ -50,10 +50,11 @@ function newspack_blocks_get_iframe( $args ) {
 		'height'       => '600px',
 		'width'        => '100%',
 		'isFullScreen' => false,
+		'align'        => '',
 	);
 	$args     = wp_parse_args( $args, $defaults );
 
-	return newspack_blocks_get_iframe_html( $args['mode'], $args['src'], $args['height'], $args['width'], $args['isFullScreen'] );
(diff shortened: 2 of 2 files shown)


dkoo (approved):
💯

laurelfulford:
Thanks @dkoo!

PR #1057, opened 2022-03-03
fix(author-list): handle alphabetizing last names containing spaces

dkoo:
### All Submissions:

Fixes a bug in the alphabetization of users or guest authors that have spaces in their last names.

1. Create a WP user and a guest author, each with first and last name filled out and with a space in the last name, e.g. "Firstname Lastname II". Also ensure that the last name appears in the display name of the user (the block will alphabetize by last name only if it appears visually in the display name, to avoid reader confusion).
2. Add an Author List block to a post and ensure that it renders both test authors created in step 1.
3. On `master`, observe that the authors with spaces in their names get alphabetized based on their display names, not their last names.
4. Check out this branch and confirm that the authors get alphabetized by their last names, as expected.

### Other information:

---
  - https://app.asana.com/0/0/1201745193114470

### DIFF
--- a/src/blocks/author-list/class-wp-rest-newspack-author-list-controller.php
+++ b/src/blocks/author-list/class-wp-rest-newspack-author-list-controller.php
@@ -240,7 +240,7 @@ class WP_REST_Newspack_Author_List_Controller extends WP_REST_Newspack_Authors_C
 							'id'         => intval( $guest_author->ID ),
 							'registered' => $guest_author->post_date,
 							'is_guest'   => true,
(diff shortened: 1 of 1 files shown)


miguelpeixe (approved):
At first, I thought there was something wrong with the order but then I realized that one of my guest authors had a "last name" only as display name, not an actual last name:

image

image

I guess it's okay since we can't assume spaces in the display name determine the last name for sorting.

dkoo:
Yes, if the user has no first name or last name, we alphabetize by the display name (and if no display name, the login name), and treat it as a single word even if it contains spaces.

PR #1083, opened 2022-03-30
fix: authors and guest authors in homepage posts

dkoo:
### All Submissions:

This PR supersedes #1082. Testing instructions remain the same except for an additional testing step at the end (see bolded header below). 8ddcee8 is the only new commit in this branch vs #1082, if that helps with code review.

Fixes four bugs related to author querying:

1. A SQL syntax error when filtering Homepage Posts blocks by authors and including at least one Co-Authors Plus guest author
2. When filtering Homepage Posts blocks by CAP guest authors, posts by linked user accounts are not included in the results
3. An unhandled `include` query param when fetching saved authors to display in the block sidebar controls
4. When filtering by a WP user, posts that were created by that user but are attributed to an unrelated CAP guest author are also fetched

#### SQL syntax error and linked accounts

1. On `master`, add several Homepage Posts blocks to a post and filter them by a regular WP user, a guest author with no linked account, and a guest author with a linked account. For ease of testing, use the same authors in all blocks and add separators between block instances so you can easily see where one block ends and the next begins.
2. Observe that the posts shown in the editor vs. the front-end are different, and that the front-end blocks in particular don't show all of the posts you would expect:
  - Likely only the first block instance will show any posts at all.
  - For the guest author with linked account, posts that are attributed to the linked account are not shown—only posts directly attributed to the guest author.
  - In the debug log, you'll probably see a SQL syntax error similar to this:

```
WordPress database error Not unique table/alias: 'wp_term_relationships' for query SELECT SQL_CALC_FOUND_ROWS  wp_posts.ID FROM wp_posts  LEFT JOIN wp_term_relationships ON (wp_posts.ID = wp_term_relationships.object_id)
```

3. Check out this branch and refresh both editor and front-end. Confirm that both environments show all the posts you would expect and that the posts match in editor vs. front-end. Confirm that the issues outlined in step 4 no longer apply.

#### Saved authors in block sidebar controls

1. Switch back to `master` and refresh the editor with your test blocks.
2. Select one of the blocks and expand the "Display Settings" sidebar. Observe that in the Authors field, not all of the authors you had saved before are shown here.
3. Check out this branch and refresh. Confirm that you now see all WP users and guest authors in the Authors field.

#### Unrelated CAP guest author posts

1. Log in as a WP user and publish a brand new post, then assign an unrelated/unlinked CAP guest author to the post (removing the WP user from the authors UI).
2. Switch back to `master`. In one of the test Homepage Posts blocks, filter by the WP user in step 1. Observe that in the editor and on the front-end, the block displays post(s) created by the WP user but assigned to unrelated guest authors.
3. Check out this branch, refresh, confirm that the block only shows posts that are directly attributed to the WP user.

#### Posts Carousel block

No changes were made to the Posts Carousel block, but it uses some of the same query functions and endpoints, so we should test it in the editor and on the front-end to ensure it still behaves as expected.

### NEW TEST CASE: subsequent non-author queries

8ddcee8 introduces a fix for an unsolved bug present in #1082. In that PR, any WP queries _after_ the wpuser+coauthors queries would inherit the `join` and `where` clause filtering done by those queries, even if the filter callback is removed. This fix now simplifies the use of the `posts_clauses` filter callback and preemtively strips all custom `join` and `where` modifications before each query.

1. On `hotfix/author-query-in-homepage-posts`, add two Homepage Posts blocks to a page: one with an author filter containing both WP users and CAP guest authors, and the other with no filters.
2. On the front-end, observe that the second block renders no posts. This is because it's inheriting the clause filtering from the previous block.
3. Check out this branch and confirm that all blocks render the expected posts.

### Other information:

### DIFF
--- a/includes/class-newspack-blocks.php
+++ b/includes/class-newspack-blocks.php
@@ -561,7 +561,49 @@ class Newspack_Blocks {
 					// look for authors and co-authors posts.
 					self::filter_posts_clauses_when_co_authors( $authors, $co_authors_names );
 				} else {
-					$args['author__in'] = $authors;
+					$args['author__in']   = $authors;
+					$authors_controller   = new WP_REST_Newspack_Authors_Controller();
+					$linked_guest_authors = [];
+					$author_names         = array_reduce(
+						$authors,
+						function( $acc, $author_id ) use ( $authors_controller, &$linked_guest_authors ) {
+							$author_data = get_userdata( $author_id );
+							if ( $author_data ) {
+								$acc[] = $author_data->user_login;
+
+								// Check if the user has a linked guest author, and if so, get those posts, too.
+								$linked_guest_author = $authors_controller->get_linked_guest_author( $author_data->user_data );
+								if ( $linked_guest_author ) {
+									$linked_guest_authors[] = $linked_guest_author->post_name;
+								}
+							}
+							return $acc;
+						},
+						[]
+					);
+
+					// Don't get any posts that are attributed to other CAP guest authors.
+					$args['tax_query'] = [ // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_tax_query
+						'relation' => 'OR',
+						[
+							'taxonomy' => 'author',
+							'operator' => 'NOT EXISTS',
+						],
+						[
+							'field'    => 'name',
+							'taxonomy' => 'author',
+							'terms'    => $author_names,
+						],
+					];
+
+					// But do get posts attributed to linked guest authors.
+					if ( 0 < count( $linked_guest_authors ) ) {
+						$tax_query[] = [
+							'field'    => 'name',
+							'taxonomy' => 'author',
+							'terms'    => $linked_guest_authors,
+						];
+					}
 				}
 			}
 			if ( $categories && count( $categories ) ) {
@@ -577,6 +619,7 @@ class Newspack_Blocks {
 				$args['category__not_in'] = $category_exclusions;
 			}
 		}
+
 		return $args;
 	}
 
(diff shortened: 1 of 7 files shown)


includes/class-newspack-blocks.php
```diff
@@ -468,6 +496,9 @@ public static function add_image_sizes() {
 	 * @return array
 	 */
 	public static function build_articles_query( $attributes, $block_name ) {
+		// Reset author/CAP guest author SQL statements by default.
+		self::$filter_clauses = false;
```
miguelpeixe, line 500: Great idea! I assume this is helping simplify the entire logic and deprecating `self::$newspack_blocks_posts_clauses_when_co_authors_closure` and `remove_filter_posts_clauses_when_co_authors_filter()`?

includes/class-newspack-blocks.php
```diff
@@ -15,6 +15,32 @@ class Newspack_Blocks {
 	 */
 	const DONATE_STREAMLINED_SCRIPT_HANDLE = 'newspack-blocks-donate-streamlined';
 
+	/**
+	 * Regex pattern we can use to search for and remove custom SQL statements.
+	 * Custom statements added by this class are wrapped by `capwhere` comments.
+	 */
+	const SQL_PATTERN = '/\/\* capwhere \*\/(.|\n)*\/\* \/capwhere \*\//';
```
miguelpeixe, line 22: Non-blocking nit: does `capwhere` refer to Co-Author Plus? This could change to either have a more "newspack-blocks" related name or be documented to refer to CAP-specific clauses.

miguelpeixe (approved):
I love the work done in 8ddcee8a6b955d15b9939aa98410786534291e45 👏 

Also can confirm that it fixes the described bug of unwanted inherited query clauses!

Left a tiny nit below:

includes/class-newspack-blocks.php
dkoo: Yes, I realized that adding and removing the `self::$newspack_blocks_posts_clauses_when_co_authors_closure` wasn't preventing the query modifications from persisting across queries, so really the only data that needs to be passed to the filter callback are the author IDs and co-author names.

dkoo: To further clarify, if the property is `false`, we preemptively strip any `newspack-blocks` modifications from the query clauses. We reset the property to `false` each time the `build_articles_query` method is run and only set it to contain the authors and co-authors data if we need to modify the query.

PR #1091, opened 2022-04-07
Add 3 new subscribe patterns

thomasguillot:
### All Submissions:

This PR adds 3 new subscribe patterns. These are slightly different from the previous ones. They use `templateLock: all` on the main `Group` and `lock:{remove:false}` on the heading and copy.

This means that these patterns are kind of locked yet there's still some flexibility with its customisation.

1. Switch to this branch
2. Insert the various patterns to a page and/or a prompt

### Other information:

### DIFF
--- a/includes/class-newspack-blocks-patterns.php
+++ b/includes/class-newspack-blocks-patterns.php
@@ -118,7 +118,10 @@ class Newspack_Blocks_Patterns {
 					'subscribe-3',
 					'subscribe-4',
 					'subscribe-5',
-					'subscribe-6'
+					'subscribe-6',
+					'subscribe-7',
+					'subscribe-8',
+					'subscribe-9'
 				);
 			}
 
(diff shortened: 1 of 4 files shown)


laurelfulford (approved):
These new patterns are really looking nice! 🎉 

The template locking seems to work well in my testing -- I was able to remove and edit heading, paragraph, and image blocks, and swap the image in the media & text block, but not actually delete the media & text, columns or contact blocks. 

The inability to add blocks tripped me up a bit, even though I knew the templateLock was in place -- it felt a little weird that I couldn't add a second paragraph, or re-add one once I deleted it. This functions as expected, though, and AFAIK this is as nuanced as locking gets right now. 

I don't think this is a blocker, but it might be a good thing to keep tabs on and see what the general feedback is -- I could be finding problems that won't come up in the real world 🙂 And worst-case, the lock can be removed by using the code editor (the other way is to "Ungroup" the group block, but that's not a good idea for these!).

thomasguillot:
> re-add one once I deleted it

Yep that's kind of the downside of `templateLock`. But I think this is better than allowing user to mess with the pattern 😃

PR #1095, opened 2022-04-08
feat(donate-streamlined): set user ID on Stripe client

adekbadek:
### All Submissions:

Adds current user ID to client metadata set in Stripe (when using streamlined Donate block)

_Follow instructions in https://github.com/Automattic/newspack-plugin/pull/1593_

### Other information:


miguelpeixe (approved):
Non-blocking but I think the current user ID could be assigned to a static class variable instead of globally for this case.

adekbadek:
Good idea, done in f24b7aede81dff76cc7229e22338102748052a53

PR #1102, opened 2022-04-13
feat: publish on npm when releasing

### LINKED ISSUE #849: Rename to @automattic and publish to npm

Right now external consumers that want to depend on `newspack-blocks` have to consume this package using a git url. Example:

```
dependencies:{
    newspack-blocks: "github:Automattic/newspack-blocks#v1.33.0"
}
```

Consuming it from a git url is problematic for a few reasons, mainly because there is no guarantee that the tag/branch `v1.33.0` will continue to exist or that it will always point to the same commit. Publishing it to NPM will make things easier for consumers.

Another problem is the name. Right now, the package `newspack-blocks` _already exists_ in NPM and it was removed because it contained malicious code (https://www.npmjs.com/package/newspack-blocks). Likely it has been the target of package squatting (https://www.flyingpenguin.com/?p=31813). I strongly recommend we rename it to `@automattic/ newspack-blocks`

- Rename the package to `@automattic/newspack-blocks`.
- Publish it to NPM

adekbadek:
Configures publishing the code as `@automattic/newspack-blocks` on npm.

Hard to imagine testing this other than looking at releases after this is merged – any release (prerelease, too) should trigger a release on npm 🤞 

Closes #849


dkoo (approved):
Like you say, hard to test this until we do a new release. Maybe we can build a new alpha to test this after merging?

Could we also use a strategy similar to this on Newspack Components to release to NPM automatically?

adekbadek:
> Maybe we can build a new alpha to test this after merging?

I think so! It's been a week since the alpha release but the changes on `master` since then are not _too_ big. 

> Could we also use a strategy similar to this on Newspack Components to release to NPM automatically?

Not really – most of the changes to the Plugin won't necessitate `newspack-components` update, so there will be a lot of versions with no changes. Components is quite separate from the Plugin, and follows it's own version history.

dkoo:
> Not really – most of the changes to the Plugin won't necessitate `newspack-components` update, so there will be a lot of versions with no changes. Components is quite separate from the Plugin, and follows it's own version history.

Agreed, but wondering if there's a way to trigger the NPM release only if the release tag contains changes in `assets/components`? Not a big priority since Components releases are relatively rare.

adekbadek:
Probably, but it'd be a lot of development for little gain. Somehow the version number would have to be determined, too.

PR #1114, opened 2022-04-19
Fix/post statuses in query

### LINKED ISSUE #38: Homepage blocks: How to handle previewing article changes
Right now, we're building article blocks for the homepage that will automatically display the latest X posts from a specific category.

While is is super-handy for updating the homepage without having to make manual changes, it introduces a weird challenge: as it works now, it's not possible to preview how newly published articles will look on the homepage. You can only see currently published ones. So if a new article varies wildly from ones that are already there (super long title; super tall photo [assuming we don't crop to a specific aspect ratio], lots of contributors), the homepage may not look ideal, and there's no real way to know that until you've published it. 

Here's a fake example that better illustrates what I mean -- say this is a section on the homepage:

image

It's made of up three article blocks, set in a column block. The first one shows one post from the Politics category, and the other two show 6 posts from two other categories. 

Say the next article in Politics published has no featured image, and a shorter headline and excerpt -- it could end up looking like:

image

There's a number of ways this could be fixed by changing the homepage settings (changing the number of posts in those sections, for example), but the problem is you won't know exactly what it will look like until the post is published.

In a similar vein, if an important big story is coming down the pipe, you're facing a similar issue with making sure its prominent position on the homepage will look good before the article goes live. 

@jeffersonrabb and I spoke briefly about ways to work around this, like having the ability to preview draft/scheduled posts in the homepage in the editor. Another could be having some mechanism to delay a new post appearing in some spots -- like having to republish the homepage after previewing a new article in the editor before it shows up. Both have their downsides -- this issue would be good to explore/discuss further to find a solution.

adekbadek:
### All Submissions:

Closes #38

A question – is checking for `edit_others_posts` capability a safe-enough check? 

For design review – the label which marks non-published posts: 

image

1. Insert Homepage Posts and Carousel blocks on a page
2. Ensure you have some recent draft and scheduled posts 
3. For both, observe a new panel in block settings – "Post Statuses". Set different values for this attribute, and observe the post list is updated – non-published posts should be displayed with a label
4. Publish and visit the page as the logged-in editor user – observe the posts are shown as in the editor
5. Visit the page as a non-logged-in user – observe only published posts are displayed

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/view.scss
+++ b/src/blocks/homepage-articles/view.scss
@@ -758,3 +758,17 @@ amp-script .wpnbha.has-more-button.is-error {
 	}
 }
 /* stylelint-enable */
+
+// Also used by the Carousel block.
+.newspack-preview-label {
+	position: absolute;
+	top: 5px;
+	left: 5px;
+	z-index: 1;
+	font-family: sans-serif;
+	font-size: 10px;
+	text-transform: uppercase;
+	background: #ffc000;
+	padding: 0 4px;
+	border-radius: 2px;
+}
--- a/src/blocks/homepage-articles/view.scss
+++ b/src/blocks/homepage-articles/view.scss
@@ -1,5 +1,6 @@
 @import '../../shared/sass/variables';
 @import '../../shared/sass/mixins';
+@import '../../shared/sass/preview';
 
 .wpnbha {
 	margin-bottom: 1em;
@@ -758,17 +759,3 @@ amp-script .wpnbha.has-more-button.is-error {
 	}
 }
 /* stylelint-enable */
-
-// Also used by the Carousel block.
-.newspack-preview-label {
-	position: absolute;
-	top: 5px;
-	left: 5px;
-	z-index: 1;
-	font-family: sans-serif;
-	font-size: 10px;
-	text-transform: uppercase;
-	background: #ffc000;
-	padding: 0 4px;
-	border-radius: 2px;
-}
--- a/includes/class-newspack-blocks-api.php
+++ b/includes/class-newspack-blocks-api.php
@@ -335,6 +335,7 @@ class Newspack_Blocks_API {
 				'newspack_has_custom_excerpt'     => self::newspack_blocks_has_custom_excerpt( $data ),
 				'newspack_post_format'            => self::newspack_blocks_post_format( $data ),
 				'newspack_post_sponsors'          => self::newspack_blocks_sponsor_info( $data ),
+				'post_status'                     => $post->post_status,
 				'post_type'                       => $post->post_type,
 				'post_link'                       => Newspack_Blocks::get_post_link( $post->ID ),
 			];
--- a/includes/class-newspack-blocks.php
+++ b/includes/class-newspack-blocks.php
@@ -1266,5 +1266,23 @@ class Newspack_Blocks {
 		ob_end_clean();
 		return $contents;
 	}
+
+	/**
+	 * Get post status label.
+	 */
+	public static function get_post_status_label() {
+		$post_status          = get_post_status();
+		$post_statuses_labels = [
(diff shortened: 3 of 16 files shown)


dkoo (changes requested):
This works well as described, but I'm uncertain about the idea of introducing a block attribute to show or not show preview posts.

Would it be enough and easier to understand if we simply showed all draft, scheduled, and published posts with labels for each non-published status in the block editor? On the front-end side, we could show these non-published posts if the user is logged in, has the `edit_others_posts` capability, and there's a `preview=true` param on the URL (meaning the post is visited from the "Preview" functionality in the editor). 

Or, to make it even simpler, just always show all of the draft, scheduled and published posts to logged-in users with the right capabilities. In this case if an editor really wants to see a post "as it would appear to readers" they can view it in an incognito window.

At the very least, I would advocate for a simpler interface: instead of being able to select Publish, Draft, and/or Scheduled, just make it a single toggle (off by default) to "Show draft and scheduled posts" with the same explanation instead. Right now I can't really think of a good use case for not showing published posts since these will always be shown to readers, right?

> A question – is checking for `edit_others_posts` capability a safe-enough check?

If we want to further lock it down we could add a nonce verification, but I think checking the login state plus `edit_others_posts` is probably enough.

adekbadek:
Thank you for your input, but I think it's based on assumptions about publishing workflows that we cannot make. 

> Would it be enough and easier to understand if we simply showed all draft, scheduled, and published posts with labels for each non-published status in the block editor?

> to make it even simpler, just always show all of the draft, scheduled and published posts to logged-in users 

I imagine a WP user might:

- Have a ton of insignificant drafts, and a few scheduled posts. They want to see how the block would look like after the scheduled posts are published. 
- Have a few drafts, which are planned for publication this week, and a couple of posts scheduled further into the future. They want to see the block as it would appear at the end of the week (drafts only).
- Not use scheduling and have a ton of insignificant drafts, which they don't ever want to see in the block.

This PR would support all these workflows, while the proposition in the comment would impair all of these, assuming drafts and scheduled posts are always **few** and **desired** when previewing.

I'm all for simplification, but this is an non-core feature, and one which which uses established CMS concepts, so I don't think it's too complicated.  

> Right now I can't really think of a good use case for not showing published posts since these will always be shown to readers, right?

That's right – how about removing the "published" status from the UI, so user can only select what they want to see in addition to the published posts/

dkoo:
> Thank you for your input, but I think it's based on assumptions about publishing workflows that we cannot make.

That's true—I think this is a feature which might benefit from some user feedback across sites of various sizes. I can certainly imagine the scenarios you describe but can't really say how many sites will actually encounter them.

> That's right – how about removing the "published" status from the UI, so user can only select what they want to see in addition to the published posts

I'm okay with this approach for now if you just want to get the feature out there.

adekbadek:
After an internal discussion:

- changed labels to specific posts status instead of generic "preview"
- removed "published" option from the checkbox-selection

src/blocks/homepage-articles/view.scss
```diff
@@ -758,3 +758,17 @@ amp-script .wpnbha.has-more-button.is-error {
 	}
 }
 /* stylelint-enable */
+
+// Also used by the Carousel block.
+.newspack-preview-label {
```
dkoo, line 763: These styles aren't enqueued if a post only contains a Carousel block and not the Homepage Posts block.

dkoo (changes requested):
A minor style issue with Post Carousels, otherwise looking 👍

src/blocks/homepage-articles/view.scss
adekbadek: Thanks for catching that! Fixed in a392e6ac4d78b86ca9bb58c65f71b027227cae2e

PR #1117, opened 2022-04-22
Donate block: make possible to navigate via keyboard

### LINKED ISSUE #123: Donation Block: Can't navigate via keyboard
When navigating through the page using a keyboard, you can't operate the donation block:

* The tabs don't get focus so you can't switch them
* The dollar values don't get focus so you can't switch them:

The only values that you can focus on by tabbing is the 'Donate' button, and, if already opened, the 'Other' text field.

image

It should be possible to change all of the settings using just a keyboard, and to get visual feedback on where you current are.

laurelfulford:
### All Submissions:

This PR tweaks the Donate block styles so the radio buttons that aren't visible are no longer hidden with `display: none`, but with `position: absolute; left: -99999em` to push them off screen. This way, they can be tabbed to when navigating the site with a keyboard, so it improves the block's accessibility. 

The styles also add some visual changes when a tab/tier is focused on -- I stuck with `outline` so it wouldn't change things too much (since these are also visible when you click on a tab/tier) but we might want to look at something more noticeable down the road, as this is pretty low contrast. 

Closes #123

1. Add a donate block to a post or page and publish.
2. Try to "tab" through the page (you can 'Skip to Content' at the top). When you get to the block, note that the focus skips all of the tabs and jumps immediately to the Donate Now button.
4. Try to "tab" through the page again. When you hit the block, the current tab should get focus and get a little dotted 'underline'; for the tiers, the selected tier should get a dotted outline when it gets focus:

image

image

6. You should be able to tab from the frequency tabs to the tiers to the button, but not from frequency tab to frequency tab or from tier to tier -- this is "correct" behaviour for radio groups, so hopefully intuitive to folks who navigate sites with their keyboards. For each group of radio buttons, or "level" in the donate block (the tabs, or tiers) you should be able to use your up-down or left-right arrow keys to move between options.
7. If you select 'Other' in the tier options, the text field should appear and you should be able to tab from the 'Other' tier option to the field, change the value, and tab to the Donate Now button.
8. You should be able to shift-tab to move backwards through the block.
9. You should be able to submit the Donate form by hitting 'enter' when the button is focused on. 

Please run through the different block styles (Alternative and Minimal) and Simplified Donate block using the above steps as well, and confirm that you're able to move around using just the keyboard, and that you see some kind of visual feedback on the elements that have focus.

### Other information:

### DIFF
--- a/src/blocks/donate/view.scss
+++ b/src/blocks/donate/view.scss
@@ -92,16 +92,16 @@
 		}
 
 		input[type='radio'] {
-			&:checked + .tier-label {
-				background-color: $color__primary;
-				border-color: transparent;
-				color: $color__background-body;
-			}
-
 			&:focus + .tier-label {
 				background: $color__background-screen;
 				outline: 1px dotted currentColor;
 				outline-offset: -4px;
+			}
+
+			&:checked + .tier-label {
+				background-color: $color__primary;
+				border-color: transparent;
+				color: $color__background-body;
 			}
 		}
 
--- a/src/blocks/donate/view.scss
+++ b/src/blocks/donate/view.scss
@@ -100,7 +100,7 @@
 
 			&:focus + .tier-label {
 				background: $color__background-screen;
-				outline: 1px dotted black;
+				outline: 1px dotted currentColor;
 				outline-offset: -4px;
 			}
 		}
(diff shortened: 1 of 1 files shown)


src/blocks/donate/view.scss
```diff
@@ -10,8 +10,10 @@
 	width: 100%;
 
 	input[type='radio'] {
-		display: none;
+		left: -99999em;
```
dkoo, line 13: Non-blocking: This works fine, but I notice we use the `left: -9999em; position: absolute;` trick in a few different places throughout the theme. What if we created a ["visually hidden" utility class](https://css-tricks.com/comparing-various-ways-to-hide-things-in-css/#aa-method-5-the-visually-hidden-class) like `.visually-hidden` and simply extended that class for the built-in theme elements that need it? Then, we can also apply the class to arbitrary elements (even in Custom CSS) if we need to hide it. What do you think?

dkoo (changes requested):
This works well, but I ran into one contrast issue using the default color scheme. When focusing a tier element, the text turns white but the background remains light grey:

Screen Shot 2022-05-02 at 4 05 03 PM

This affects the "default" and "alternate" styles. Suggestion: keep the text black, or turn it a different but higher-contrast color?

src/blocks/donate/view.scss
laurelfulford: We definitely could for theme-specific stuff (maybe the next time `left: -9999em; position: absolute;` gets added?). I don't think this is what you mean here, but just in case: I don't think we'd want to rely on that for a block, though -- it's not a use case we'd run into, but it would mean the block wouldn't look/function totally correctly without the theme.

laurelfulford:
> This works well, but I ran into one contrast issue using the default color scheme. When focusing a tier element, the text turns white but the background remains light grey:

Oooo, good catch @dkoo! I ended up flipping the order of the styles (moving focus in front of checked) so the current tier keeps the primary background colour, and the focus styles are just the dotted outline.

Let me know if that doesn't seem like enough, though!

src/blocks/donate/view.scss
dkoo: Ah, you're right—I was thinking of the theme because I see this pattern in a few places there. But since the two repos don't share source code we wouldn't be able to `@extend` SCSS from one to another. One possibility would be to add common/shared utility classes to the `newspack-scripts` repo, but I think that's overcomplicating things. So nevermind, for now. Thanks for pointing out the dependency issue here.

dkoo (approved):
The contrast issue looks fixed to me, now. Thanks for the update!

laurelfulford:
Thanks Derrick! 🙌

PR #1133, opened 2022-05-10
fix: remove custom column block styles

laurelfulford:
### All Submissions:

Closes # .

1.
2.
3.

### Other information:

### DIFF
--- a/src/block-styles/core/columns/editor.scss
+++ b/src/block-styles/core/columns/editor.scss
@@ -1,16 +1,16 @@
 @import '../../../shared/sass/variables';
 @import '../../../shared/sass/mixins';
 
+.editor-styles-wrapper div.wp-block-columns {
+	gap: 32px;
+
+	&.is-style-borders {
+		gap: 64px;
+	}
+}
+
 .wp-block-columns {
 	@include media( mobile ) {
-		&[class*='is-style-first-col-to'] [data-type='core/column'] {
-			margin-left: 2em;
-		}
-
-		&[class*='is-style-first-col-to'] [data-type='core/column']:nth-child( 2 ) {
-			margin-left: 0;
-		}
-
 		&.is-style-first-col-to-second [data-type='core/column']:nth-child( 2 ),
 		&.is-style-first-col-to-third [data-type='core/column']:nth-child( 2 ),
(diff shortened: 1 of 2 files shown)


dkoo (changes requested):
This one is a doozy because there are SO many variations of content that need to be handled. I ran into ~a few issues~ an issue (the others were due to me not having the right Theme branch checked out 🤦 ) with the test content you provided, confirmed in both Chrome and Firefox.

The four-columns-with-border-don't-stack-on-mobile styles appear broken on mobile. This happens with both WP 5.9.3 and 6.0. Only seems to affect this particular combination of styles with more than three columns, but the borders appear off-center at mobile breakpoints:

Screen Shot 2022-05-19 at 4 44 24 PM

laurelfulford:
Thanks @dkoo! 🙏 

> The four-columns-with-border-don't-stack-on-mobile styles appear broken on mobile

I think this might have to do with my test content, and the fact the border styles eat up a lot of space. I tried adding red borders to the `.wp-block-column` blocks on my test site, and made the paragraphs semi-transparent -- it helps show that the borders are aligned correctly, but the paragraphs are overflowing the available space:

Screen Shot 2022-05-20 at 9 37 51 AM

Though it's an edge case, this isn't great 😕 I've decreased the column gap a bit for these not-stacked columns on smaller screens in [a683470](https://github.com/Automattic/newspack-blocks/pull/1133/commits/a6834705c2b1f32347525f96db9469b0c4524559) -- I don't think it'll fix it if you switch to five columns, but at that point I think you could truly argue it's something you shouldn't be doing:

image


> What's strange is that some of these combos—like reordering columns AND retaining borders—don't seem possible to create using the UI...

This is absolutely correct -- you can't add multiple block styles to the columns. You can manually add one of the block style classes though -- and I know we've done this in the past 😬 So I included some cases like that to make sure things will still look okay.

dkoo (approved):
> I think this might have to do with my test content, and the fact the border styles eat up a lot of space. I tried adding red borders to the .wp-block-column blocks on my test site, and made the paragraphs semi-transparent -- it helps show that the borders are aligned correctly, but the paragraphs are overflowing the available space:

I see what you mean. This is an extreme edge case, anyway, and the tweak does help this particular test content look better at small viewports.

laurelfulford:
Thanks @dkoo!

PR #1142, opened 2022-05-26
Add new subscribe pattern

thomasguillot:
### All Submissions:

This is a very simple pattern, 2 columns (70 - 30) with paragraph and mailchimp block. It's intended for "above header" inline prompts.

1. Switch to this branch
2. Apply Newspack theme PR: https://github.com/Automattic/newspack-theme/pull/1818
3. Create new above header prompt
4. Insert this pattern
5. Publish

### Other information:

### DIFF
--- a/includes/class-newspack-blocks-patterns.php
+++ b/includes/class-newspack-blocks-patterns.php
@@ -121,7 +121,8 @@ class Newspack_Blocks_Patterns {
 					'subscribe-6',
 					'subscribe-7',
 					'subscribe-8',
-					'subscribe-9'
+					'subscribe-9',
+					'subscribe-10'
 				);
 			}
 
--- a/src/block-patterns/subscribe-10.php
+++ b/src/block-patterns/subscribe-10.php
@@ -4,3 +4,9 @@
  *
  * @package Newspack_Blocks
  */
+
+return array(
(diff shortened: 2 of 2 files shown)


laurelfulford (approved):
This looks good to me overall! 

I have some non-blocking feedback that probably says more about my eyesight than anything: this particular arrow symbol feels a bit small to me 😅 A big part of it is the font size, which we're stuck with; I did find some heavier options (❯ ➔ ➤) but I'm not sure they quite fit as well with the overall look. Again, definitely not a blocker, especially since you can still edit the button!

thomasguillot:
> This looks good to me overall!
> 
> I have some non-blocking feedback that probably says more about my eyesight than anything: this particular arrow symbol feels a bit small to me 😅 A big part of it is the font size, which we're stuck with; I did find some heavier options (❯ ➔ ➤) but I'm not sure they quite fit as well with the overall look. Again, definitely not a blocker, especially since you can still edit the button!

I went for the most basic arrow because I wasn't sure if all the fonts/languages were supporting all the options. Ideally, I wanted to use an SVG but it'd make things waaaaaay too overcomplicated.

laurelfulford:
> I went for the most basic arrow because I wasn't sure if all the fonts/languages were supporting all the options. Ideally, I wanted to use an SVG but it'd make things waaaaaay too overcomplicated.

Ah, that makes sense! 👍

PR #1143, opened 2022-05-26
fix: disambiguate WP users vs. guest authors with same ID

### LINKED ISSUE #1134: Author Profile: Guest author ID overrides user ID
**Describe the bug**
If an Author Profile block is embedded for a specific user ID, and a guest author exists with the same post ID, the guest author will be selected instead.

**To Reproduce**
1. Create a JN site.
2. Install the CAP and Newspack Blocks v1.49.0 plugins.
3. Create a new guest author. Their post ID will be `6`.
4. Create five new author-role users. The fifth one's user ID will be `6`.
5. Create a post with an Author Profile block. Try to embed user ID 6. Guest author 6 will be selected instead.

**Expected behavior**
The Author Profile block should be able to distinguish between post IDs and user IDs. An update to this code may be needed:

https://github.com/Automattic/newspack-blocks/blob/master/src/blocks/author-profile/view.php#L38

dkoo:
### All Submissions:

Accounts for a previously unhandled edge case where a CAP guest author and a WP user might have the same numeric ID in the DB. In this case, `master` will always choose the guest author over the WP user because we can't tell which type of author to fetch with only the ID.

This PR fixes the above issue by storing a new block attribute `isGuestAuthor` as part of the Author Profile block. If this is `true` (the default), the block's behavior will behave as it currently does (search guest authors first, and if none found, fall back to WP users). If it's `false`, it will only search WP users. This ensures that if a WP user is chosen which has the same ID as a CAP guest author, the WP user will be displayed in the editor and the front-end.

Note that this doesn't fix pre-existing instances of the Author Profile block affected by this issue, as there is no way to disambiguate the numeric IDs without the new `isGuestAuthor` attribute. These blocks will need to be manually rebuilt in the editor.

This PR also standardizes REST API parameters as snake case and properly lists and sanitizes all possible param values in the REST route registration. Otherwise, the Author List and Author Profile behavior should remain the same.

Closes #1134.

1. On a brand new site with no posts and Co-Authors Plus and Newspack Blocks (current release) installed, create enough WP users and CAP guest authors so that you have at least one guest author with the same ID as a WP user. **Hint:** if you run `wp site empty --yes` at CLI and then create a guest author, the ID will be `1` and the same as the main admin user.
2. Create some Author List blocks with various settings (for testing after the plugin upgrade).
3. Create an Author Profile block and attempt to select the WP user with an ID that overlaps with a guest author. Observe that the guest author is displayed in the editor and on the front-end despite your selection.
4. Install the Blocks plugin from this branch.
5. Recreate the Author Profile block in step 3 and confirm that it now displays the correct WP user in the editor and on the front-end.
6. Confirm that the Author List blocks still behave as before, and that both Author Profile and Author List blocks continue to behave as expected.

### Other information:

### DIFF
--- a/src/blocks/author-list/class-wp-rest-newspack-author-list-controller.php
+++ b/src/blocks/author-list/class-wp-rest-newspack-author-list-controller.php
@@ -37,19 +37,28 @@ class WP_REST_Newspack_Author_List_Controller extends WP_REST_Newspack_Authors_C
 					'methods'             => WP_REST_Server::READABLE,
 					'callback'            => [ $this, 'api_get_all_authors' ],
 					'args'                => [
-						'author_id' => [
+						'author_id'     => [
 							'sanitize_callback' => 'absint',
 						],
-						'offset'    => [
+						'author_roles'  => [
+							'sanitize_callback' => 'WP_REST_Newspack_Author_List_Controller::sanitize_array',
+						],
+						'author_types'  => [
+							'sanitize_callback' => 'WP_REST_Newspack_Author_List_Controller::sanitize_array',
+						],
+						'exclude_empty' => [
 							'sanitize_callback' => 'absint',
 						],
-						'per_page'  => [
+						'offset'        => [
 							'sanitize_callback' => 'absint',
 						],
-						'search'    => [
+						'per_page'      => [
+							'sanitize_callback' => 'absint',
+						],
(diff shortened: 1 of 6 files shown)


adekbadek (approved):
This fixes the issue. Noting that the Author List block should also be updated – added #1150

PR #1144, opened 2022-05-27
Homepage & Carousel block: skip linked images when navigating by keyboard

### LINKED ISSUE #1116: Homepage Posts block: image behind setting and keyboard navigation
Similar to #1115, this issue came out of a publisher's accessibility audit:

When you set the featured image to display 'behind' the post title in the Homepage Posts block, it kind of becomes a link that isn't visible, since it's pushed down and under the post content. It can't be clicked on as is, and can't be styled to cover everything (you end up with a weird "dead spot" in the block, as described in #501).

On top of that, when you're using a keyboard to navigate, the link is truly invisible -- you get no visual feedback when you tab to it, even an outline.

The exact feedback was:

> Invisible elements can receive focus and are visible to screen reader users at all times.

... not realizing that the link getting tabbed to contained a visible image (especially since the link isn't actually styled to contain the whole image).

Given #501 and the above, I think the best fix might be to remove the link from the featured image when this style is applied OR perhaps add some kind of visual feedback on :focus, so you can tell you're "on" the image.

laurelfulford:
### All Submissions:

This PR adds `tabindex="-1"` and `aria-hidden="true"` to the linked images in the Homepage Posts Block and Post Carousel block. This uses a similar approach to the theme's [archive and search markup](https://github.com/Automattic/newspack-theme/blob/8f47f1b03ec557279295f6a7ce6339513cf8c218/newspack-theme/inc/template-tags.php#L394). 

It's in part to fix #1116; but it also reduces the number of elements you need to tab through on each story when you're navigating the page by keyboard -- the photo in the Homepage posts block and Post Carousel have the same links as the titles in these blocks, so as is it creates an extra step you have to "get past" to navigate through each story (this was a recommendation that came out of the same accessibility review as #1116). 

Closes #1116.

1. Set up a page with a couple different Homepage posts blocks (with different image settings) and at least one Post Carousel block.
2. Navigate through the page using your keyboard; note that for each post in the blocks, the focus goes first to the image, then the title, then author before moving on to the next.
3. Apply the PR. 
4. Repeat step two and confirm that the focus skips the image in each block. 

### Other information:

### DIFF
--- a/src/blocks/carousel/view.php
+++ b/src/blocks/carousel/view.php
@@ -79,7 +79,7 @@ function newspack_blocks_render_block_carousel( $attributes ) {
 				<?php echo Newspack_Blocks::get_post_status_label(); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>
 				<figure class="post-thumbnail">
 					<?php if ( $post_link ) : ?>
-					<a href="<?php echo esc_url( $post_link ); ?>" rel="bookmark">
+					<a href="<?php echo esc_url( $post_link ); ?>" rel="bookmark" tabindex="-1" aria-hidden="true">
 					<?php endif; ?>
 						<?php if ( $has_featured_image ) : ?>
 							<?php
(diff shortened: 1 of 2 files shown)


adekbadek (approved):
👐

laurelfulford:
Thanks @adekbadek!

PR #1152, opened 2022-05-31
Sponsors plugin integration: echo closing link tags on sponsor bylines

### LINKED ISSUE #1151: Sponsors Plugin integration: Missing closing `</a>` tag on byline
**Describe the bug**
In the Sponsors byline markup, [we're missing an `echo` on a closing `</a>`](https://github.com/Automattic/newspack-blocks/blob/master/src/blocks/homepage-articles/templates/article.php#L206), which is causing some malformed HTML. This is only noticeable on non-AMP pages -- it looks like AMP is "correcting" the HTML for us. 

When AMP is not enabled, the missing `</a>` causes some issues, ranging from adding a link to the date in the block through breaking HTML that follows it.

laurelfulford:
### All Submissions:

In the Sponsors markup, the code is missing an `echo` on the closing link on the byline for both the Homepage Posts and Post Carousel blocks. I think this flew under the radar because it doesn't seem to cause issues when AMP is enabled (AMP seems to be correcting the malformed HTML), but when AMP is disabled, it causes visual issues. 

Closes #1151.

1. Start with a test site with a number of sponsored posts.
2. Set up a Post Carousel and Homepage Posts block that each display at least one of your sponsored posts that has a native sponsor; enable dates on both blocks.
3. View on the front end and disable AMP.
4. If your blocks have dates displaying, they should be incorrectly picking up the sponsor byline's link; failing that, you can also inspect the page and see that the sponsor byline link seems to appear more than once (depending on how your browser handles malformed HTML). 

image

image

6. Apply the PR.
7. Confirm that the links are now correctly formed in both the Post Carousel and Homepage Post blocks. 

### Other information:

### DIFF
--- a/src/blocks/carousel/view.php
+++ b/src/blocks/carousel/view.php
@@ -184,7 +184,7 @@ function newspack_blocks_render_block_carousel( $attributes ) {
 										}
 										echo esc_html( $byline['name'] );
 										if ( '' !== $byline['url'] ) {
-											'</a>';
+											echo '</a>';
 										}
 										echo '</span>' . esc_html( $byline['sep'] );
 									}
--- a/src/blocks/homepage-articles/templates/article.php
+++ b/src/blocks/homepage-articles/templates/article.php
@@ -203,7 +203,7 @@ call_user_func(
 							}
 							echo esc_html( $byline['name'] );
 							if ( '' !== $byline['url'] ) {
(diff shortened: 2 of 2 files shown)


dkoo (approved):
Good catch! Confirmed that this fixes the problem 👍

PR #1156, opened 2022-06-03
feat: support new sponsor options to show authors and categories

dkoo:
### All Submissions:

Implements new sponsor-related options from https://github.com/Automattic/newspack-sponsors/pull/186.

Follow testing instructions in https://github.com/Automattic/newspack-sponsors/pull/186.

### Other information:


src/blocks/homepage-articles/templates/article.php
```diff
@@ -95,20 +95,19 @@ class="<?php echo esc_attr( implode( ' ', $classes ) ); ?>"
...
 
 		<div class="entry-wrapper">
-			<?php if ( ! empty( $sponsors ) ) : ?>
-				<span class="cat-links sponsor-label">
-					<span class="flag">
-						<?php echo esc_html( Newspack_Blocks::get_sponsor_label( $sponsors ) ); ?>
-					</span>
-				</span>
-			<?php elseif ( $attributes['showCategory'] && $category ) : ?>
+			<?php if ( ! empty( $sponsors ) || ( $attributes['showCategory'] && $category ) ) : ?>
```
adekbadek, line 98: If `showCategory` is `false` for block, but a post has a sponsor, the categories will be displayed.

adekbadek (changes requested):

src/blocks/homepage-articles/templates/article.php
laurelfulford: I ran into something similar: the categories are always displayed with the sponsor flag, even when the sponsor's settings is to only show the 'flag' and not the categories.

laurelfulford:
Hi @dkoo! I took a pass at updating the styles in the blocks.

Unlike the theme, there are changes in this PR that also affect the sponsors when the authors and categories are not visible: 

* I wrapped the sponsors in a `span` with the CSS class `.entry-sponsors` to help avoid the sponsors and authors from running together -- otherwise, the sponsor logos, sponsor bylines, author avatars and author bylines were all inside of the same parent element, and it was tricky to avoid them running together.
* I made it so the sponsor logo and byline sit side-by-side. Originally they always stacked, but this seemed weird when the author is also visible; the logo and sponsor will still stack if there's not enough room.
* If authors are also showing, a second CSS class (`plus-author`) is added to the `entry-sponsors` container to allow adding a bit of space underneath the sponsors, and to allow making the sponsors text a bit smaller, to match how they're handled in single posts. 

Overall, the sponsor flag and category should sit side by side when both are visible, and should wrap over two lines if there's not a lot of space (like when the image is floated left or right). 

The Sponsors should appear above the post author when both are visible. The sponsor text should be slightly smaller when both are visible as well. 

I've attached some screenshots of different block options, including the different text alignments, image alignments, and text sizes for the Homepage Posts block, and a couple of the Post Carousel block.

<details>
<summary><strong>Homepage Posts</strong></summary>

Default text alignment:

image

Text align centre:

image

Text align right: 

image

The Sponsor text should remain the same size for font scales 4+ (like the rest of the post meta), but it should shrink with the smaller sizes along with the other post meta. For sizes 2 and 3, it should be smaller than 4 but the same size; it should go down even smaller for size 1:

image

If the sponsor flag + category wrap over two lines, they should be nicely spaced out: 

image

</details>

<details>
<summary><strong>Post Carousel</strong></summary>

image

image

</details>

dkoo:
@laurelfulford thanks again for undertaking such a big set of changes! I ran into just a few extremely nitpicky/edge case things here:

With text at right alignment, the final image has a bit of margin which makes it look not aligned with the right edge of the column:
Screen Shot 2022-06-29 at 4 01 48 PM

With the featured image thumbnails set to float on either side, the text alignment options do not affect the sponsor logos, even though they do affect the author avatars. It's somewhat hard to tell from the screenshots, but it's not because of the margins mentioned above—removing all margins in dev tools makes the logos appear flush-left.
Screen Shot 2022-06-29 at 4 03 56 PM

In the same screenshot above, you might notice there's a bit of right margin on the sponsor flag which makes it look misaligned with the column's right edge. This might be onerous to fix, it's not very noticeable, and it's likely pretty edge casey, so I'm okay with leaving it for now.

> this PR is currently failing some CircleCI tests, though I'm not sure why -- for the JS and TS tests, it's throwing newspack-scripts: not found and for the scss, it's returning stylelint: not found -- does this PR just need to be rebased, or is there something else happening?

Rebasing seems to have fixed the CI errors—thanks for highlighting!

src/blocks/homepage-articles/templates/article.php
dkoo: Whoops, good catch. b49888d should fix this.

src/blocks/homepage-articles/edit.js
```diff
@@ -145,6 +146,7 @@ class Edit extends Component {
 
 		const postTitle = this.titleForPost( post );
 		const dateFormat = __experimentalGetSettings().formats.date;
+		console.log( post );
```
adekbadek, line 149: ✂️
(I've added a rule for logging in the latest version of `newspack-scripts`)

adekbadek (approved):
Just one line to remove and we're good here.

src/blocks/homepage-articles/edit.js
dkoo: Fixed in 9107c50, thanks for catching!

dkoo:
@laurelfulford I'm going to say this is all non-blocking now, because I think this is edge-casey enough that almost no site would use this combo of settings. But noting just for the record...

The alignment of sponsor logos looks great in column view now, on the front-end!
Screen Shot 2022-07-08 at 3 21 26 PM

However, there's still a bit of extra space in the editor:
Screen Shot 2022-07-08 at 3 20 21 PM

Same deal with the floated images + text alignment. Looks great on the front-end now:
Screen Shot 2022-07-08 at 3 23 38 PM

But still looks off in the editor:
Screen Shot 2022-07-08 at 3 23 29 PM

I'd say since this only affects a specific combination of settings in the editor only, it's okay to leave for now, unless it's obvious to you what needs to be fixed. WDYT?

laurelfulford:
Thanks @dkoo! 🤦‍♀️ It looks like this was a mix of editor block styles coming from the theme (removed [here](https://github.com/Automattic/newspack-theme/pull/1833/commits/3a39e51b4c592265719524ad72234ae14f349355)), and some missing alignment styles in the blocks -- I was able to recreate some weirdness with the right alignment in the logos and category + sponsors flag on the front-end too. 

In [5b5c00e](https://github.com/Automattic/newspack-blocks/pull/1156/commits/5b5c00e5d8d27a4006cb67301f153ac5235a6e98) I added and updated the editor styles I removed from the theme, and added a right text alignment that seems to have fixed things. If you could give it one more peek, I'd appreciate it -- I think if there's anything else still kicking around, it should be safe to address in a separate PR unless it's really catastrophic 😅

dkoo:
@laurelfulford The latest changes fixed all the issues I was seeing in the editor. 🥳 I updated with `master` and fixed a couple of conflicts. @adekbadek since I opened the PR initially, do you want to do one last check before we merge, or are we good to go?

adekbadek (approved):
👌

PR #1170, opened 2022-06-14
feat: homepage posts patterns with ad unit

miguelpeixe:
### All Submissions:

Updates Homepage Posts patterns 2, 7, 14, 26 and implements new pattern 31 as described in https://github.com/Automattic/newspack-ads/issues/210.

This PR also inaugurates "placeholder blocks", a simple config to allow a more descriptive missing block. The placeholder should provide an easy button to remove the missing block, a description and a link to the plugin that provides the block:

image

Closes https://github.com/Automattic/newspack-ads/issues/210

1. With Newspack Ads active, create a new page and add patterns listed above
2. Confirm they reflect https://github.com/Automattic/newspack-ads/issues/210
3. Deactivate Newspack Ads
4. Refresh the editor
5. Confirm the Ad Unit blocks are replaced with the placeholder as pictured above

### Other information:


thomasguillot:
@miguelpeixe this is great!

I pushed some visual changes to make it look like more of a placeholder.



One thing I was thinking, the "Visit Plugin Page" button doesn't add much value here. It's potentially distracting the user if they click on it, losing their tab/editor just to see our Github page. Instead, would it be possible to Install/Activate the plugin directly from here? I know they would still need to setup Newspack Ads, etc... but at least they'd have the plugin already active.

thomasguillot:
> I thought about it and the issue with Ads is not only that you currently can't just one-click install and use it inside the block editor but also that Newspack Ads depends on Newspack Plugin. We'd have to install 2 plugins on this button action and I really don't think we should.

Is there any way for us to check if Newspack Plugin is active and Newspack Ads is simply inactive? Then we should the "Activate Plugin" button.
And if not, we display the "Visit Plugin Page" button

miguelpeixe:
> Is there any way for us to check if Newspack Plugin is active and Newspack Ads is simply inactive? Then we should the "Activate Plugin" button.
> And if not, we display the "Visit Plugin Page" button

It is a possibility, but I'd rather tackle it separately as it also depends on https://github.com/Automattic/newspack-ads/issues/444

src/setup/placeholder-blocks.js
```diff
@@ -0,0 +1,83 @@
...
+import { withDispatch } from '@wordpress/data';
+import { Icon, pullquote } from '@wordpress/icons';
+
+const placeholderBlocks = {
+	'newspack-ads/ad-unit': {
+		title: __( 'Ad Unit', 'newspack-blocks' ),
+		description: __( 'Render an ad unit from your inventory.', 'newspack-blocks' ),
+		icon: pullquote,
+		message: __( 'Place ad units inside your page by installing Newspack Ads.', 'newspack-blocks' ),
+		url: 'https://github.com/Automattic/newspack-ads',
```
adekbadek, line 17: Suggestion: Github URL won't be of much use to most WP users, might be better to point to https://newspack.pub/support/advertising/

adekbadek (approved):

PR #1234, opened 2022-08-10
feat: use reCaptcha v3 to secure Stripe donate forms

### LINKED ISSUE #1209: Donate (streamlined): Add a means of including a captcha

As it is now, the block and native checkout are vulnerable to card testing. We've seen a massive hit on one site already (120K attempts). For the standard donate block, we can add a reCaptcha to the checkout page. The reCaptcha extension doesn't work for this block. 

Include a means of adding a captcha or a check to block bots and card testing attempts. 

None.

dkoo:
### All Submissions:

Implements reCaptcha v3 to secure Stripe donation forms. Relies on reCaptcha v3 credentials entered into the UI implemented in https://github.com/Automattic/newspack-plugin/pull/1879.

If reCaptcha is enabled and credentials are stored, then all form submissions via the Stripe donate block will be tested against the reCaptcha v3 API before we attempt to process the transaction with Stripe. If the reCaptcha test returns a score lower than `0.5` (the default threshold for likely valid/human interactions), then we will fail the test and stop the transaction with an error message.

Note that reCaptcha v3 is an "invisible" form of captcha and doesn't involve any front-end user input, so the change will be totally invisible to most valid users. [More info here](https://developers.google.com/recaptcha/docs/v3).

Closes #1209.

1. Check out this branch and https://github.com/Automattic/newspack-plugin/pull/1879.
2. Do a gut check first, without saving any credentials or changing any settings: enable Stripe as your donation platform, and submit a test donation via the streamlined donate block. Confirm that the donation succeeds as expected.
3. Go to https://www.google.com/recaptcha/admin/create and create a set of reCaptcha v3 keys for your test site.
4. In the WP dashboard, go to Newspack > Reader Revenue > Stripe Settings, enable reCaptcha v3, and enter your site key and secret into the relevant fields. Save.
5. In a new front-end session, submit another test donation and confirm that it succeeds—since you're submitting via direct HTML form interaction, it should pass the reCaptcha test and the transaction should go through.
6. As a test, temporarily change the threshold value on [this line](https://github.com/Automattic/newspack-blocks/compare/release...hotfix/recaptcha-for-stripe?expand=1#diff-5349f9a82b35a56353d631adcc470a92d4a82347acddd0695b7f803cbd38f144R18) to `1.0`, which is a perfect score and nearly impossible to achieve.
7. Submit another test donation and confirm that the transaction fails with an error message.

### Other information:

### DIFF
--- a/src/blocks/donate/class-wp-rest-newspack-donate-controller.php
+++ b/src/blocks/donate/class-wp-rest-newspack-donate-controller.php
@@ -147,8 +147,8 @@ class WP_REST_Newspack_Donate_Controller extends WP_REST_Controller {
 
 			// If the reCaptcha verification score is below our threshold for valid user input.
 			if (
-			isset( $captcha_verify['score'] ) &&
-			Newspack_Blocks::DONATE_STREAMLINED_CAPTCHA_THRESHOLD > floatval( $captcha_verify['score'] )
+				isset( $captcha_verify['score'] ) &&
+				Newspack_Blocks::DONATE_STREAMLINED_CAPTCHA_THRESHOLD > floatval( $captcha_verify['score'] )
 			) {
 				return rest_ensure_response(
 					[
--- a/src/blocks/donate/class-wp-rest-newspack-donate-controller.php
+++ b/src/blocks/donate/class-wp-rest-newspack-donate-controller.php
@@ -100,11 +100,11 @@ class WP_REST_Newspack_Donate_Controller extends WP_REST_Controller {
 	 */
 	public function api_process_donation( $request ) {
 		// If reCaptcha is available, verify the user action.
-		$use_captcha     = \Newspack\Stripe_Connection::can_use_captcha();
-		$stripe_settings = \Newspack\Stripe_Connection::get_stripe_data();
-		$captcha_secret  = $stripe_settings['captchaSiteSecret'];
+		$use_captcha = \Newspack\Stripe_Connection::can_use_captcha();
 		if ( $use_captcha ) {
-			$captcha_token = $request->get_param( 'captchaToken' );
+			$stripe_settings = \Newspack\Stripe_Connection::get_stripe_data();
+			$captcha_secret  = $stripe_settings['captchaSiteSecret'];
+			$captcha_token   = $request->get_param( 'captchaToken' );
 
 			if ( ! $captcha_token ) {
 				return rest_ensure_response(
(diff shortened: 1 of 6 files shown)


src/blocks/donate/class-wp-rest-newspack-donate-controller.php
```diff
@@ -95,6 +99,65 @@ public function register_routes() {
...
+						// Translators: error message for reCaptcha.
+						'error' => sprintf( __( 'reCaptcha error: %s', 'newspack-blocks' ), $error ),
+					]
+				);
+			}
+
+			// If the reCaptcha verification score is below our threshold for valid user input.
+			if (
+			isset( $captcha_verify['score'] ) &&
+			Newspack_Blocks::DONATE_STREAMLINED_CAPTCHA_THRESHOLD > floatval( $captcha_verify['score'] )
```
miguelpeixe, line 151:
```suggestion
				isset( $captcha_verify['score'] ) &&
				Newspack_Blocks::DONATE_STREAMLINED_CAPTCHA_THRESHOLD > floatval( $captcha_verify['score'] )
```

Nit: linting

src/blocks/donate/streamlined/index.js
```diff
@@ -28,6 +28,20 @@ export const processStreamlinedElements = ( parentElement = document ) =>
 		const enableForm = () => el.classList.remove( 'stripe-payment--disabled' );
 		enableForm();
 
+		const getCaptchaToken = async reCaptchaKey => {
+			return new Promise( ( res, rej ) => {
+				const { grecaptcha } = window;
+				grecaptcha.ready( async () => {
```
miguelpeixe, line 34: Even though the library is a declared dependency, shouldn't we check `grecaptcha` is not undefined before using it?

miguelpeixe (changes requested):
Nice! 🎉 

Just a nit and a question below

src/blocks/donate/class-wp-rest-newspack-donate-controller.php
dkoo: Strange how this didn't trigger my linter—updated!

src/blocks/donate/streamlined/index.js
dkoo: Probably a good idea. I had thought maybe it was enough to declare it a dependency, but just in case: 97b7832

miguelpeixe (approved):
🙌 :shipit:

PR #1239, opened 2022-08-15
feat: handle minimum donation option in Donate block

### LINKED ISSUE #1211: Donate Block (streamlined): Include an option to set a minimum amount

We should set a minimum required amount for the payment to process. NRH sets theirs at $5 as do a few others. 

We should add a minimum that is greater than $1 and include an option for admins to change it if they so choose. 

I could see this setting either in the Stripe tab or Donations. Either makes sense.

dkoo:
### All Submissions:

Handles the new "minimum donation" setting from https://github.com/Automattic/newspack-plugin/pull/1895 in the Donate block front-end and editor.

Closes #1211 and https://github.com/Automattic/newspack-plugin/issues/1886.

See testing instructions in https://github.com/Automattic/newspack-plugin/pull/1895.

### Other information:

### DIFF
--- a/src/blocks/donate/edit.tsx
+++ b/src/blocks/donate/edit.tsx
@@ -264,7 +264,7 @@ const Edit = ( { attributes, setAttributes, className }: EditProps ) => {
 			{ label && <label htmlFor={ id }>{ label }</label> }
 			<input
 				type="number"
-				min="0"
+				min={ attributes.minimumDonation }
 				onChange={ evt =>
 					handleCustomDonationChange( {
 						value: evt.target.value,
--- a/src/blocks/donate/edit.tsx
+++ b/src/blocks/donate/edit.tsx
@@ -578,6 +578,7 @@ const Edit = ( { attributes, setAttributes, className }: EditProps ) => {
 								</div>
 							) }
 							<TextControl
+								className="components-frequency-donations__minimum-donation"
 								type="number"
 								label={ __( 'Minimum donation', 'newspack-blocks' ) }
 								min={ 1 }
--- a/src/blocks/donate/edit.tsx
+++ b/src/blocks/donate/edit.tsx
@@ -48,6 +48,7 @@ type OverridableConfiguration = {
 	disabledFrequencies: {
 		[ Key in DonationFrequencySlug as string ]: boolean;
 	};
+	minimumDonation: number;
 };
 
 type DonateBlockAttributes = OverridableConfiguration & {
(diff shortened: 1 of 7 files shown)


src/blocks/donate/streamlined/index.ts
```diff
@@ -292,7 +292,9 @@ export const processStreamlinedElements = ( parentElement = document ) =>
 			);
 
 			const formValues = utils.getDonationFormValues( formElement );
-			const validationErrors = Object.values( utils.validateFormData( formValues ) );
+			const validationErrors = Object.values(
+				utils.validateFormData( formValues, settings?.minimumDonation || 5 )
```
adekbadek, line 296: WDYT about passing all settings to validation, so in the future it could rely on more signals without refactoring?

src/blocks/donate/edit.tsx
```diff
@@ -256,15 +264,19 @@ const Edit = ( { attributes, setAttributes, className }: EditProps ) => {
...
 						value: evt.target.value,
 						frequency: frequencySlug,
 						tierIndex,
 					} )
 				}
-				value={ amounts[ frequencySlug ][ tierIndex ] }
+				value={
+					amounts[ frequencySlug ][ tierIndex ] >= attributes.minimumDonation
+						? amounts[ frequencySlug ][ tierIndex ]
+						: attributes.minimumDonation
```
adekbadek, line 278: Same as in the [other review](https://github.com/Automattic/newspack-plugin/pull/1895#discussion_r946477783), this will not allow some valid values.

adekbadek (changes requested):

src/blocks/donate/edit.tsx
dkoo: Added warnings instead of coercing the values in 269c573.

269c573 also fixes a bug where the Minimum Donation field was only shown if the "Tiered" options was checked.

src/blocks/donate/streamlined/index.ts
dkoo: Good idea—refactored in 269c573.

src/blocks/donate/edit.tsx
```diff
@@ -535,6 +577,13 @@ const Edit = ( { attributes, setAttributes, className }: EditProps ) => {
 									</div>
 								</div>
 							) }
+							<TextControl
```
adekbadek, line 580: This element would use some top-margin. If the last frequency is disabled, the elements are too close: 

image

adekbadek (approved):
Pre-approving, since only a CSS tweak is needed.

src/blocks/donate/edit.tsx
dkoo: Added some top margin in c010de8:

Screen Shot 2022-08-17 at 11 01 11 AM

PR #1248, opened 2022-08-19
fix(donate): default value not below minimum donation

miguelpeixe:
### All Submissions:

This PR sets the default value rendered on the block to never be below the configured minimum. It would be unintuitive for the donor not to be able to donate with the proposed value on the form.

Although the settings can be saved with a lower value, the displayed warning should be enough to let the user know that it may not behave as expected.

1. On the master branch, configure Reader Revenue Donations settings with a minimum donation to 5 and tiers below that value
2. Visit the page with the block and confirm these values are rendered and you are unable to donate with the default values
3. Check out this branch and confirm the values are never rendered below 5 even though you are still able to store it that way and prompted with the warning

### Other information:

### DIFF
--- a/src/blocks/donate/view.php
+++ b/src/blocks/donate/view.php
@@ -369,7 +369,7 @@ function newspack_blocks_render_block_donate( $attributes ) {
 											type='number'
 											min='<?php echo esc_attr( $configuration['minimumDonation'] ); ?>'
 											name='donation_value_<?php echo esc_attr( $frequency_slug ); ?>_untiered'
-											value='<?php echo esc_attr( $formatted_amount ); ?>'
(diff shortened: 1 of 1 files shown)


dkoo (changes requested):
This is a good enhancement! We should also apply the same check to the "tiered" inputs and their labels.

miguelpeixe:
Done in adcc7311d55ec197e4dd024bff1f35910dee5aef

dkoo (changes requested):
The same `max` comparison also needs to be applied on lines 443 and 451 (input values and labels for the tiered suggested amounts).

miguelpeixe:
Whoops! Sorry about the back and forth! Should be fixed by a4a3e6ba30373c9233027dee77392e9f8572abe0

dkoo (approved):
Works well with both tiered and untiered options, now!

PR #1253, opened 2022-08-24
feat: if donation via a prompt, add prompt ID to Stripe payment metadata

dkoo:
### All Submissions:

Detects if a donation happened from inside a Newspack Campaigns prompt, and if it did, adds the prompt's unique ID to the payment metadata sent to Stripe for processing payment. This will get picked up by https://github.com/Automattic/newspack-plugin/pull/1928 to append the prompt ID to the label attached to the GA event for the donation.

Follow testing instructions in https://github.com/Automattic/newspack-plugin/pull/1928.

### Other information:

### DIFF
--- a/src/blocks/donate/streamlined/index.ts
+++ b/src/blocks/donate/streamlined/index.ts
@@ -108,6 +108,14 @@ export const processStreamlinedElements = ( parentElement = document ) =>
 				}
 			}
 			const formValues = utils.getDonationFormValues( formElement );
+			const promptOrigin = formElement.closest( '.newspack-popup' );
+
+			// If the donation originated from a Campaigns prompt, append the prompt ID to the event label.
+			const origin =
+				promptOrigin && promptOrigin.hasAttribute( 'amp-access' )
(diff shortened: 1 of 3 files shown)


src/blocks/donate/streamlined/index.ts
```diff
@@ -108,6 +108,14 @@ export const processStreamlinedElements = ( parentElement = document ) =>
 				}
 			}
 			const formValues = utils.getDonationFormValues( formElement );
+			const promptOrigin = formElement.closest( '.newspack-popup' );
```
adekbadek, line 111: To work with both inline & overlay prompts, it should target `amp-layout`.

```suggestion
			const promptOrigin = formElement.closest( 'amp-layout.newspack-popup' );
```

In an overlay prompt, there's a child of `amp-layout` that also gets `.newspack-popup` class. Which should not be happening, I've added https://github.com/Automattic/newspack-popups/issues/945

adekbadek (changes requested):

adekbadek (approved):

PR #1322, opened 2022-12-02
HPB: filter image sizes based on the maximum image width

adekbadek:
### All Submissions:

Improves performance by removing image sizes which are too large to load from the `srcset` attribute of the `img` tags rendered by the Homepage Posts block. 

1. On `master`,
1. Create a page with some Homepage Posts blocks, make sure all articles have featured images
2. Set one of the blocks to multiple columns
3. Load the page, observe all images are downloaded in the same size (inspect the Network devtools tab)
4. Switch to this branch and rebuild the thumbnails (`wp media regenerate`)
5. Load the page, observe more appropriate image sizes are loaded (this is most visible for a block with a few columns)

### Other information:

### DIFF
--- a/includes/class-newspack-blocks.php
+++ b/includes/class-newspack-blocks.php
@@ -421,8 +421,8 @@ class Newspack_Blocks {
 					900,
 				),
 				'medium' => array(
-					800,
 					600,
+					450,
 				),
 				'small'  => array(
 					400,
@@ -439,8 +439,8 @@ class Newspack_Blocks {
 					1200,
 				),
 				'medium' => array(
+					450,
 					600,
-					800,
 				),
 				'small'  => array(
 					300,
@@ -457,8 +457,8 @@ class Newspack_Blocks {
 					1200,
 				),
 				'medium' => array(
-					800,
-					800,
+					600,
+					600,
 				),
 				'small'  => array(
 					400,
@@ -492,9 +492,9 @@ class Newspack_Blocks {
 		add_image_size( 'newspack-article-block-portrait-large', 900, 1200, true );
 		add_image_size( 'newspack-article-block-square-large', 1200, 1200, true );
 
-		add_image_size( 'newspack-article-block-landscape-medium', 800, 600, true );
-		add_image_size( 'newspack-article-block-portrait-medium', 600, 800, true );
-		add_image_size( 'newspack-article-block-square-medium', 800, 800, true );
+		add_image_size( 'newspack-article-block-landscape-medium', 600, 450, true );
+		add_image_size( 'newspack-article-block-portrait-medium', 450, 600, true );
+		add_image_size( 'newspack-article-block-square-medium', 600, 600, true );
 
 		add_image_size( 'newspack-article-block-landscape-small', 400, 300, true );
 		add_image_size( 'newspack-article-block-portrait-small', 300, 400, true );
--- a/includes/class-newspack-blocks.php
+++ b/includes/class-newspack-blocks.php
@@ -416,55 +416,67 @@ class Newspack_Blocks {
 	public static function image_size_for_orientation( $orientation = 'landscape' ) {
 		$sizes = array(
 			'landscape' => array(
-				'large'  => array(
+				'large'        => array(
 					1200,
 					900,
 				),
-				'medium' => array(
+				'medium'       => array(
+					800,
+					600,
+				),
+				'intermediate' => array(
 					600,
 					450,
 				),
-				'small'  => array(
+				'small'        => array(
 					400,
 					300,
 				),
-				'tiny'   => array(
+				'tiny'         => array(
 					200,
 					150,
 				),
 			),
 			'portrait'  => array(
-				'large'  => array(
+				'large'        => array(
 					900,
 					1200,
 				),
-				'medium' => array(
+				'medium'       => array(
+					600,
+					800,
+				),
+				'intermediate' => array(
 					450,
 					600,
 				),
-				'small'  => array(
+				'small'        => array(
 					300,
 					400,
 				),
-				'tiny'   => array(
+				'tiny'         => array(
 					150,
 					200,
 				),
 			),
 			'square'    => array(
-				'large'  => array(
+				'large'        => array(
 					1200,
 					1200,
 				),
-				'medium' => array(
+				'medium'       => array(
+					800,
+					800,
+				),
+				'intermediate' => array(
 					600,
 					600,
 				),
-				'small'  => array(
+				'small'        => array(
 					400,
 					400,
 				),
-				'tiny'   => array(
+				'tiny'         => array(
 					200,
 					200,
 				),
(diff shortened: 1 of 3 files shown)


laurelfulford:
Putting myself in the review mix mostly to look at the visuals, but I think this would also benefit from a review from another member of the product team on the code side! 🙂

laurelfulford (changes requested):
First, I think some kind of refactor of these images is long overdue! Thanks for tackling this, @adekbadek! 

I have some feedback that's mostly flagging possible visual issues -- I don't think we necessarily have to fux all of this (or even can get it working 100% in all cases), but hopefully just knowing this could happen will be useful down the road!

First, some behaviour stuff: the Homepage Posts Block [includes CSS that stretch images to fit the available space](https://github.com/Automattic/newspack-blocks/blob/master/src/blocks/homepage-articles/view.scss#L140-L143) — this was originally meant to hide cases where the original images were a bit too small, but could also be masking how this is working. 

Though it's not really the use case for this optimization, AMP stops the images from being scaled up so turning it on will show the images in their true sizes -- that could help troubleshoot where things aren’t scaling as expected.

Some things I ran into:

The code currently assumes if a HPB is in a column, the image will only need to be 600px wide; this isn't always the case. Even some of our patterns have the columns block wider; to confuse things more, you can also have a single 100% wide column in a columns block. It's not something I think anyone _intends_ to do, but I think it happens because there's not a nice way to remove a columns block from around something, but you can switch it to one column and it looks the same as if it wasn't there at all.

Here's an example of one of our Homepage Patterns + AMP, to show the image at 600px wide:

Screen Shot 2022-12-06 at 11 03 33 AM

Similarly, the image-beside styles -- the default is L (50%), but there's an XL size, too, where the image actually takes up 75% of the available space -- here's the L and XL sizes + AMP to show the actual images getting returned: 

Screen Shot 2022-12-06 at 11 09 31 AM

Wide and full width blocks can also mess with the image math, since they can get a lot larger than expected, especially on larger monitors:

Screen Shot 2022-12-06 at 11 12 51 AM

Without AMP, in all these cases the images will be stretched to fill the space, so there won't be the white space issue. However, there's going to be a range of quality loss both depending on the original image and how much it's getting scaled up.

One possible bug I ran into is that when I started with a 1200px wide image, it was still scaled down to the 'medium' size, even when not in a column/in a grid layout. Anything larger worked fine; I first noticed this with our demo content (where all images are pre-sized to 1200px wide), but I could also recreate it when I uploaded an image fresh: 
 
Screen Shot 2022-12-06 at 11 12 18 AM

And couple other odd gotchas we might want to keep on our radars: 
* With the block available on WP.com, it is in use with themes that are wider than 1200px -- the block already maxes the image size out at 1200px for any of the crops, so we're likely to have heard of some sizing issues, but the 600px mid-point could crop up more.
* We also have some sites that have scaled up their site's width ([example](https://www.bettergov.org/) and [example](https://van-magazine.com/)) using CSS. Sites that do this will already have issues with the 1200px max image size the theme and block currently load, but they could see more issues if the max image sizes are based on a 1200px theme. Block themes have the option to change site element widths through the editor UI, so it could be something that will happen more when it's possible to change without CSS.

Please let me know if you have any questions at all about any of this, or if there's anything I can do to assist!

adekbadek:
Thank you for a thorough review! 

> The code currently assumes if a HPB is in a column, the image will only need to be 600px wide; this isn't always the case. 

That was a side effect of messing with the image sizes – instead of changing the `medium` from 800px wide to 600px wide, b1a202e53e48952036677d960b0ba4ec1aacb149 adds an `intermediate` size (please rebuild the thumbnails again after pulling!)

> With the block available on WP.com, it may be used in themes that are wider than 1200px

8fc9b0e3b5446024ccf43f9f776147f22af8ffe6 adds a check for using a Newspack theme.

>  We also have some sites that have scaled up their site's width (example and example) using CSS.

7d10f8ca508ff9d18bc8d5304353802b16ad99af enables disabling this optimisation with a `NEWSPACK_DISABLE_HPB_IMAGE_OPTIMISATION` environment variable, so it can be disabled per-site.

laurelfulford (changes requested):
> Thank you for a thorough review!

Glad it’s helpful! To be honest, I’m fishing for edge cases a bit here to make sure we’re not surprised by any outcomes, so hopefully it’s not too much!

> That was a side effect of messing with the image sizes – instead of changing the medium from 800px wide to 600px wide b1a202e adds an intermediate size (please rebuild the thumbnails again after pulling!)

That makes sense! The 800x600 covers the columns defaults (up to 66.66% wide with the Newspack theme), but if the column is larger will still load images smaller than the space. 

It’s a similar issue to the XL image size when the image is aligned left or right — both I think are rare, but can happen with normal settings and could be noticeable. 

To be honest, I'm not sure if there's a good way to solve these, outside of making the image larger still (which kind of defeats the purpose!). 

> 8fc9b0e adds a check for using a Newspack theme.
> 7d10f8c enables disabling this optimisation with a `NEWSPACK_DISABLE_HPB_IMAGE_OPTIMISATION` environment variable, so it can be disabled per-site.

Thanks for both of these! I wonder if it would also make sense to add a check for AMP, since it makes the smaller images more noticeable? (They’ll still be scaled up, but I think too-small images are more noticeable than scaled up). 

I think the only thing I'm still seeing that could be a bug is when original image is exactly 1200px wide, it’s getting scaled down to 800px even when the post isn’t in a grid/column; larger images don’t seem to have this issue so I wonder if it’s just a weird rounding thing?

adekbadek:
The code in this PR only kicks in for blocks with more than one column, or media position on the left or right. I can recreate some of the reported issues on `master`, so this is not a regression.

I've added handling for block alignment (wide and full) in d3f3bd7290f66f7dc0c1cc12e26923c47b1c4f82.

laurelfulford (approved):
Thanks @adekbadek! 

I replaced a bunch of my test images with ones larger than 1200px (ruling out any weirdness from that issue), and that, paired with this PR, resolve pretty much everything I was seeing: images in wide and full width blocks look good, images in columns (regardless of numbers) look good. 

The only thing that persists is when you have the image aligned left or right at the XL size, the image itself is only 600px, though it's being scaled up even with AMP enabled. 

image

I do feel like this one more of an usual setting -- even without it not being wrapped in a columns block, the XL size is probably the least-used of the image sizes, and issue is a loss of quality over the image displaying too small even with AMP enabled. 

I'm not sure it's actually worth increasing the image size for the left/right images for this, so I'm giving it a 👍 with the caveat that we _may_ get reports about the image quality for this very specific setting, and in that case it could be worth revisiting (I think even 800 over the larger 1200 would be an improvement if it does come up, but for now that's pretty big when most of the images with this block setting would be small). 

The code does look good to me, but I think it could be good to get a more critical eye on that end of things, just to make sure I'm not missing something!

claudiulodro (approved):
Great improvement. It's not possible to catch every possible edge case, but logic looks solid and it works well in my testing.

PR #1330, opened 2022-12-12
feat(donate): additional fields on the form

adekbadek:
### All Submissions:

Adds additional fields feature to the streamlined Donate block:

image

_(address and phone fields above are added per-block)_

1. Switch `newspack-plugin` to `feat/donation-additional-fields` branch (https://github.com/Automattic/newspack-plugin/pull/2187)
1. Ensure the site is set up to use Stripe as the Reader Revenue platform and webhooks can reach the site
2. Insert a Donate block on a page. Observe new editor sidebar panel – "Additional data fields"
3. Add some fields
4. Visit the front-end and make a donation, supply values to the additional fields
5. Observe that the fields are added to the Stripe customer metadata – as separate fields, but also the whole group, JSON-encoded, under `_additional_fields` meta key

### Other information:

### DIFF
--- a/src/blocks/donate/edit/components/AdditionalFields.tsx
+++ b/src/blocks/donate/edit/components/AdditionalFields.tsx
@@ -26,13 +26,13 @@ const BASE_CSS_CLASSNAME = 'newspack-blocks-additional-fields-editor';
 
 type EditableKey = keyof Omit< AdditionalField, 'type' >;
 const FIELD_PROPS = [
-	[ 'label', __( 'Label', 'newspack' ) ],
+	[ 'label', __( 'Label', 'newspack-blocks' ) ],
 	[
 		'name',
-		__( 'Name', 'newspack' ),
+		__( 'Name', 'newspack-blocks' ),
 		__(
 			'Name of the field which will be sent to the payment procesor and other third parties. Field names must be unique.',
-			'newspack'
+			'newspack-blocks'
 		),
 	],
 ] as [ EditableKey, string, string ][];
@@ -45,7 +45,7 @@ const FieldOptions = ( { onEdit, onRemove }: { onEdit: () => void; onRemove: () 
 			<Button
 				onClick={ toggleVisible }
 				icon={ moreVertical }
-				label={ __( 'Options', 'newspack' ) }
+				label={ __( 'Options', 'newspack-blocks' ) }
 				tooltipPosition="bottom center"
 			/>
 			{ isVisible && (
@@ -56,7 +56,7 @@ const FieldOptions = ( { onEdit, onRemove }: { onEdit: () => void; onRemove: () 
 					className={ `${ BASE_CSS_CLASSNAME }__options-popover` }
 				>
 					<MenuItem onClick={ onEdit } isLink>
-						{ __( 'Edit', 'newspack' ) }
+						{ __( 'Edit', 'newspack-blocks' ) }
 					</MenuItem>
 					<MenuItem
 						isDestructive
@@ -66,7 +66,7 @@ const FieldOptions = ( { onEdit, onRemove }: { onEdit: () => void; onRemove: () 
 						} }
 						isLink
 					>
-						{ __( 'Remove', 'newspack' ) }
+						{ __( 'Remove', 'newspack-blocks' ) }
 					</MenuItem>
 				</Popover>
 			) }
@@ -109,7 +109,7 @@ const FieldEditor = ( {
 					attributes.additionalFields.filter(
 						( { name }, i ) => i !== field.fieldIndex && name === field.name
 					).length === 0;
-				return ! isValid ? __( 'Name already exists.', 'newspack' ) : '';
+				return ! isValid ? __( 'Name already exists.', 'newspack-blocks' ) : '';
 		}
 	};
 	return (
@@ -140,7 +140,7 @@ const FieldEditor = ( {
 				);
 			} ) }
 			<ToggleControl
-				label={ __( 'Required', 'newspack' ) }
+				label={ __( 'Required', 'newspack-blocks' ) }
 				checked={ field.isRequired }
 				onChange={ updateField( 'isRequired' ) }
 			/>
@@ -205,7 +205,7 @@ const AdditionalFields = ( {
 						setEditedField( newField );
 					} }
 				>
-					{ __( 'Add data field', 'newspack' ) }
+					{ __( 'Add data field', 'newspack-blocks' ) }
 				</Button>
 			</div>
 
--- a/src/blocks/donate/edit/components/AdditionalFields.tsx
+++ b/src/blocks/donate/edit/components/AdditionalFields.tsx
@@ -146,7 +146,7 @@ const FieldEditor = ( {
 			<div className={ `${ BASE_CSS_CLASSNAME }__field-edited__width` }>
 				<div>{ __( 'Field width:', 'newspack-blocks' ) }</div>
 				<ButtonGroup>
-					{ [ 100, 66, 50, 33.33 ].map( width => (
+					{ [ 100, 66.66, 50, 33.33 ].map( width => (
 						<Button
 							key={ width }
 							onClick={ () => updateField( 'width' )( width ) }
--- a/src/blocks/donate/edit/components/AdditionalFields.tsx
+++ b/src/blocks/donate/edit/components/AdditionalFields.tsx
@@ -217,9 +217,7 @@ const AdditionalFields = ( {
 										disabled={ i === attributes.additionalFields.length - 1 }
 									/>
 								</div>
-								<Button className={ `${ BASE_CSS_CLASSNAME }__field__label` } onClick={ onEdit }>
-									{ field.label }
-								</Button>
+								<span>{ field.label }</span>
 							</div>
 
 							<FieldOptions
--- a/src/blocks/donate/edit/components/AdditionalFields.tsx
+++ b/src/blocks/donate/edit/components/AdditionalFields.tsx
@@ -146,13 +146,13 @@ const FieldEditor = ( {
 			<div className={ `${ BASE_CSS_CLASSNAME }__field-edited__width` }>
 				<div>{ __( 'Field width:', 'newspack-blocks' ) }</div>
 				<ButtonGroup>
-					{ [ 100, 66, 50, 33 ].map( width => (
+					{ [ 100, 66, 50, 33.33 ].map( width => (
 						<Button
 							key={ width }
 							onClick={ () => updateField( 'width' )( width ) }
 							isPrimary={ field.width === width }
 						>
-							{ width }%
+							{ Math.round( width ) }%
 						</Button>
 					) ) }
 				</ButtonGroup>
--- a/src/blocks/donate/edit/components/AdditionalFields.tsx
+++ b/src/blocks/donate/edit/components/AdditionalFields.tsx
@@ -1,0 +1,94 @@
+/**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { TextControl, CheckboxControl, Button } from '@wordpress/components';
+
+/**
+ * Internal dependencies
+ */
+import type { AdditionalField, EditProps } from '../../types';
+
+const getUniqID = () => Math.random().toString( 36 ).substring( 2, 7 );
+
+const AdditionalFields = ( {
+	attributes,
+	setAttributes,
+}: Pick< EditProps, 'attributes' | 'setAttributes' > ) => {
+	return (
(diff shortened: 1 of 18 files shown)


src/blocks/donate/edit/components/AdditionalFields.tsx
```diff
@@ -0,0 +1,240 @@
...
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore – these *are* exported by newspack-components.
+import { Modal, Popover } from 'newspack-components';
+
+const getUniqID = () => Math.random().toString( 36 ).substring( 2, 7 );
+const BASE_CSS_CLASSNAME = 'newspack-blocks-additional-fields-editor';
+
+type EditableKey = keyof Omit< AdditionalField, 'type' >;
+const FIELD_PROPS = [
+	[ 'label', __( 'Label', 'newspack' ) ],
```
dkoo, line 29: Text domain?

```suggestion
	[ 'label', __( 'Label', 'newspack-blocks' ) ],
```

src/blocks/donate/edit/components/AdditionalFields.tsx
```diff
@@ -0,0 +1,240 @@
...
+	[
+		'name',
+		__( 'Name', 'newspack' ),
```
dkoo, line 32:
```suggestion
		__( 'Name', 'newspack-blocks' ),
```

src/blocks/donate/edit/components/AdditionalFields.tsx
```diff
@@ -0,0 +1,240 @@
...
+		__(
+			'Name of the field which will be sent to the payment procesor and other third parties. Field names must be unique.',
+			'newspack'
```
dkoo, line 35:
```suggestion
			'newspack-blocks'
```

src/blocks/donate/edit/components/AdditionalFields.tsx
```diff
@@ -0,0 +1,240 @@
...
+
+const FieldOptions = ( { onEdit, onRemove }: { onEdit: () => void; onRemove: () => void } ) => {
+	const [ isVisible, setIsVisible ] = useState( false );
+	const toggleVisible = () => setIsVisible( ! isVisible );
+	return (
+		<div>
+			<Button
+				onClick={ toggleVisible }
+				icon={ moreVertical }
+				label={ __( 'Options', 'newspack' ) }
```
dkoo, line 48:
```suggestion
				label={ __( 'Options', 'newspack-blocks' ) }
```

src/blocks/donate/edit/components/AdditionalFields.tsx
```diff
@@ -0,0 +1,240 @@
...
+			/>
+			{ isVisible && (
+				<Popover
+					position="bottom left"
+					onFocusOutside={ toggleVisible }
+					onKeyDown={ ( event: KeyboardEvent ) => ESCAPE === event.keyCode && toggleVisible }
+					className={ `${ BASE_CSS_CLASSNAME }__options-popover` }
+				>
+					<MenuItem onClick={ onEdit } isLink>
+						{ __( 'Edit', 'newspack' ) }
```
dkoo, line 59:
```suggestion
						{ __( 'Edit', 'newspack-blocks' ) }
```

src/blocks/donate/edit/components/AdditionalFields.tsx
```diff
@@ -0,0 +1,240 @@
...
+					</MenuItem>
+					<MenuItem
+						isDestructive
+						onClick={ () => {
+							onRemove();
+							toggleVisible();
+						} }
+						isLink
+					>
+						{ __( 'Remove', 'newspack' ) }
```
dkoo, line 69:
```suggestion
						{ __( 'Remove', 'newspack-blocks' ) }
```

src/blocks/donate/edit/components/AdditionalFields.tsx
```diff
@@ -0,0 +1,240 @@
...
+		closeEditor();
+	};
+	const getValidationMessage = ( key: EditableKey ) => {
+		switch ( key ) {
+			case 'name':
+				const isValid =
+					attributes.additionalFields.filter(
+						( { name }, i ) => i !== field.fieldIndex && name === field.name
+					).length === 0;
+				return ! isValid ? __( 'Name already exists.', 'newspack' ) : '';
```
dkoo, line 112:
```suggestion
				return ! isValid ? __( 'Name already exists.', 'newspack-blocks' ) : '';
```

src/blocks/donate/edit/components/AdditionalFields.tsx
```diff
@@ -0,0 +1,240 @@
...
+							</div>
+						) }
+						{ help && (
+							<div className={ `${ BASE_CSS_CLASSNAME }__field-edited__help` }>{ help }</div>
+						) }
+					</div>
+				);
+			} ) }
+			<ToggleControl
+				label={ __( 'Required', 'newspack' ) }
```
dkoo, line 143:
```suggestion
				label={ __( 'Required', 'newspack-blocks' ) }
```

src/blocks/donate/edit/components/AdditionalFields.tsx
```diff
@@ -0,0 +1,240 @@
...
+							type: 'text',
+							name: `field-${ getUniqID() }`,
+							label: `Field ${ attributes.additionalFields.length }`,
+							isRequired: false,
+							isNew: true,
+						};
+						setEditedField( newField );
+					} }
+				>
+					{ __( 'Add data field', 'newspack' ) }
```
dkoo, line 208:
```suggestion
					{ __( 'Add data field', 'newspack-blocks' ) }
```

src/blocks/donate/edit/index.tsx
```diff
@@ -30,28 +30,22 @@ import { isEmpty, pick } from 'lodash';
...
 } from '../consts';
 
-type EditProps = {
-	attributes: DonateBlockAttributes;
-	setAttributes: ( attributes: Partial< DonateBlockAttributes > ) => void;
-	className: string;
-};
-
 const TIER_LABELS = [
 	__( 'Low-tier', 'newspack' ),
```
dkoo, line 50: These two, also, even though they're already in `master`

dkoo (changes requested):
This works really well. Saw a few instances of the wrong text domain, below, and I also have few feature requests, the first more important and the others not so important (non-blocking).

1. Can we implement some way to reorder the fields? Right now the only way to reorder them is to delete and rebuild. Some up/down arrow icons on the field buttons would be enough, IMO.
2. (Non-blocking) The fields all get rendered at full-width, which can look awkward on a large screen. Maybe we could add an option to the custom fields to make them full-width or half-width? This will be a fairly common set of fields, and State and ZIP could definitely be half-width at larger viewports:

Screen Shot 2022-12-16 at 2 14 05 PM

3. (Non-blocking) A nitpick, but can we make the field names in the sidebar clickable to open the edit modal? They look like buttons, so I want to click them! 😄 

Screen Shot 2022-12-16 at 4 50 50 PM

adekbadek:
Fixed the text domain wholesale in 76c4e5a

> Can we implement some way to reorder the fields? 

Great idea! Done in 5a3fcf9

>  A nitpick, but can we make the field names in the sidebar clickable to open the edit modal? T

Done in d33238f. FYI @thomasguillot here's the updated UI:

image

> The fields all get rendered at full-width, which can look awkward on a large screen. 

Added a width setting:

image

Let me know if the styling should be tweaked @thomasguillot

adekbadek:
Update: reverted field label clickability as per @thomasguillot 's request, in b61bc3c.

src/blocks/donate/frontend/class-newspack-blocks-donate-renderer-base.php
```diff
@@ -246,6 +246,20 @@ protected static function render_streamlined_payment_ui( $attributes ) {
...
 						>
 					</div>
+					<div class="stripe-payment__row stripe-payment__row--additional-fields">
+						<?php foreach ( $attributes['additionalFields'] as $field ) : ?>
+							<input
+								data-is-additional-field
+								type="<?php echo esc_attr( $field['type'] ); ?>"
+								name="<?php echo esc_attr( $field['name'] ); ?>"
+								placeholder="<?php echo esc_attr( $field['label'] ); ?>"
+								style="width: calc(<?php echo esc_attr( $field['width'] ); ?>% - 0.5rem);"
```
dkoo, line 256: This probably only happened to me because I had fields created before you added the `width` property, but I saw a fatal at first for fields that didn't have a set width. The fatal did not happen after I deleted and recreated the fields.

```suggestion
								<?php if ( isset( $field['width'] ) ) : ?>
								style="width: calc(<?php echo esc_attr( $field['width'] ); ?>% - 0.5rem);"
								<?php endif; ?>
```

dkoo (changes requested):
Thanks for the changes! I think the widths and reordering add a lot of value to this feature.

I ran into some new minor issues with the widths:

1. The widths don't quite add up to 100%. Here's what it looks like if I set three fields with 33% width:

Screen Shot 2022-12-19 at 4 11 54 PM

2. This is another feature request, sorry. Is it possible to add non-interactive placeholder fields to the editor preview of the block? I found myself having to view the post in a preview tab and switch back and forth between the two in order to see what my changes would look like, which seems like something publisher's won't want to do.

3. Non-blocking: I found myself wanting a `25%` option for the widths, too, so I could set a row to 50%, 25%, 25%. Too much?

4. Non-blocking: This may only be required if you introduce a `25%` option, but I think the widths could potentially be applied only to larger viewports (so for mobile they'd all be 100%). We'd need some help text to explain this if we go this route.

gamebits:
Can these fields be set to invisible? For example, if there is static data the publisher wants sent to Stripe with every transaction, to distinguish Newspack donations from other Stripe activity. (Or is there another way to do that?)

src/blocks/donate/frontend/class-newspack-blocks-donate-renderer-base.php
adekbadek: Yep, that's expected. The feature is not merged yet so there's no need to handle "legacy" fields.

adekbadek:
That's an interesting feature request – can you think of any other uses besides marking transaction as Newspack-originating? We're already adding a `current_page_url` and some other metadata by default, this field could be another one of these. 

> Or is there another way to do that?

I don't think so.

adekbadek:
> The widths don't quite add up to 100%. Here's what it looks like if I set three fields with 33% width:

Fixed in 4fc0b9

> Is it possible to add non-interactive placeholder fields to the editor preview of the block? 

Yes! Great suggestion – added in c19249a8cb9a4bc7ce544475aefe8b312822c275

> Non-blocking: I found myself wanting a 25% option for the widths, too, so I could set a row to 50%, 25%, 25%. Too much?

I did ask @thomasguillot in a DM previously and he said it's too much

src/blocks/donate/edit/components/AdditionalFields.tsx
```diff
@@ -0,0 +1,282 @@
...
+			} ) }
+			<ToggleControl
+				label={ __( 'Required', 'newspack-blocks' ) }
+				checked={ field.isRequired }
+				onChange={ updateField( 'isRequired' ) }
+			/>
+			<div className={ `${ BASE_CSS_CLASSNAME }__field-edited__width` }>
+				<div>{ __( 'Field width:', 'newspack-blocks' ) }</div>
+				<ButtonGroup>
+					{ [ 100, 66, 50, 33.33 ].map( width => (
```
dkoo, line 149:
```suggestion
					{ [ 100, 66.66, 50, 33.33 ].map( width => (
```

dkoo (approved):
Thanks for the preview feature—it makes editing the block much more user-friendly! Pre-approving because there's just one small fix needed to ensure that 66% width fields are rendered at the full width (e.g. if you have a 66% and 33% field side-by-side, they should add up to 100%).

src/blocks/donate/edit/components/AdditionalFields.tsx
leogermani: > (e.g. if you have a 66% and 33% field side-by-side, they should add up to 100%).

Shouldn't it be 66.67 or 33.34 then?

dkoo: The `Math.round` [here](https://github.com/Automattic/newspack-blocks/pull/1330/files#diff-049b01968454fddef70955d3fd9e376175da7a45a99ca02cfbe7ae017c73d2ccR155) will ensure that it's rounded up to 100% when totaled.

PR #1343, opened 2023-01-10
fix: make author block font sizing more specific

laurelfulford:
### All Submissions:

Styles added for the Author Profile block are overriding custom font sizing on paragraph blocks in the editor and on the front-end.

When I originally noticed this I just assumed it was a block editor issue - I wish I had investigated sooner! 

@dkoo and @adekbadek - I'm adding you both as reviews specifically because it looks like [you co-authored the original PR](https://github.com/Automattic/newspack-blocks/pull/947) where this was added. I don't think I've broken anything, but the change to the editor styles -- specifically correcting the containing CSS class so they're getting picked up -- makes me a little nervous, I just want to make sure the other styles there are still working as expected!

See 1203536997639429-as-1203571822380539

1. Set up a page with few paragraph block with different sizing (both using the theme options and some manually entered values), and at least one Author Profile block. In the Author Profile, make sure the author's job title and employer is displayed.
2. Note that the paragraph blocks with custom font sizes are not previewing correctly.
4. Confirm that the paragraphs are now picking up the correct font sizes. 
5. Confirm that the author profile block is picking up the correct scaling - the Employer should be a bit smaller than the rest of the text, and the text should scale up and down on the front-end and in the editor when you change the Type Scale. 

### Other information:

### DIFF
--- a/src/blocks/author-profile/editor.scss
+++ b/src/blocks/author-profile/editor.scss
@@ -26,8 +26,8 @@
 			font-size: 1em;
 		}
 
-		.wp-block-newspack-blocks-author-profile__employment {
-			font-size: 0.8em;
+		&__employment {
+			font-size: 0.8em !important;
 		}
 
 		&__bio {
--- a/src/blocks/author-profile/editor.scss
+++ b/src/blocks/author-profile/editor.scss
@@ -2,7 +2,7 @@
 @use '../../shared/sass/mixins';
 
 .editor-styles-wrapper {
-	.newspack-blocks-author-profile {
+	.wp-block-newspack-blocks-author-profile {
 		.components-notice {
 			margin: 0 0 2rem 0;
 			width: 100%;
@@ -20,6 +20,14 @@
 
 		a.no-op {
 			pointer-events: none;
+		}
+
+		p {
+			font-size: 1em;
+		}
+
+		.wp-block-newspack-blocks-author-profile__employment {
+			font-size: 0.8em;
 		}
 
 		&__bio {
--- a/src/blocks/author-profile/editor.scss
+++ b/src/blocks/author-profile/editor.scss
@@ -26,10 +26,6 @@
 			font-size: 1em;
 		}
 
-		&__employment {
-			font-size: 0.8em !important;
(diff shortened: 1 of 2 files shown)


src/blocks/author-profile/editor.scss
```diff
@@ -22,6 +22,14 @@
...
 		}
 
+		p {
+			font-size: 1em;
+		}
+
+		&__employment {
+			font-size: 0.8em !important;
+		}
+
```
adekbadek, line 32: Instead of these two rules, I'd rather go for an `!important` in the `view.scss` file, to keep the font size setting in on place.

src/blocks/author-profile/view.scss
```diff
@@ -66,9 +66,11 @@
 		margin-top: 0;
 		margin-bottom: 0;
 	}
-	&__employment {
-		font-size: 0.8em !important;
+
+	& &__employment {
```
adekbadek, line 70: Is the additional `&` necessary? With an `!important` in this rule (as suggested in the other comment), it's fine with just `&__employment` specificity.

adekbadek (changes requested):

src/blocks/author-profile/editor.scss
laurelfulford: That makes sense! I was so fixed on trying to get rid of the first `!important`, I didn't think to circle back after realizing I needed it in the editor styles here. Fixed in [817f2e3](https://github.com/Automattic/newspack-blocks/pull/1343/commits/817f2e3d8898fdd1f03b3a86cf4f5bc2c1c44fff)!

src/blocks/author-profile/view.scss
laurelfulford: Thanks Adam! Should be fixed in [817f2e3](https://github.com/Automattic/newspack-blocks/pull/1343/commits/817f2e3d8898fdd1f03b3a86cf4f5bc2c1c44fff)!

adekbadek (approved):

laurelfulford:
Thanks @adekbadek!

PR #1355, opened 2023-01-24
feat(donate): checkout modal for woocommerce

miguelpeixe:
### All Submissions:

This PR implements a simplified modal WooCommerce checkout for the Donate block.

When authenticated, the billing fields are removed in order to provide a more straight-forward experience:

https://user-images.githubusercontent.com/820752/215204878-e472fa09-dc47-4f61-a277-d2f43f721df3.mov

While unauthenticated, the checkout form includes ~the fields "First Name", "Last Name" and "Email"~ all fields by default. These values are defined here and are filterable so they can be customized by a wizard setting in Newspack core (Reader Revenue wizard?) or any other plugin:

https://github.com/Automattic/newspack-blocks/blob/76612d915b817efa21837ab376215138e761da8e/src/blocks/donate/frontend/class-newspack-blocks-donate-renderer.php#L36-L53

The idea is that it can still be a full checkout form, with standard address fields and the expected friction that it entails, but preserving the navigation.

At the end of the transaction, if a new reader account is created and RAS is active, the navigation is updated so the page knows it's now an authenticated reader (without the need for a page refresh).

https://user-images.githubusercontent.com/820752/215205326-e69c869a-2f45-43c7-83bf-99150b8368c6.mov

#### Update

https://github.com/Automattic/newspack-plugin/pull/2256/commits/b5c1d4936cd0e90c47211bbeeb8f2eee8bd2d97d introduced customizable billing fields:

image

The available fields are retrieved from `WC_Checkout()->get_checkout_fields()` so any new field should be integrated with WooCommerce in order to become available. This also allows for it to be compatible with other plugins that may add new fields.

The "Email" field is the only one not editable because it's required.

1. Checkout this branch and https://github.com/Automattic/newspack-plugin/pull/2256
2. Set the Reader Revenue platform to WooCommerce with a valid payment gateway
3. Make sure you have AMP Plus (AMP plugin on with `define( 'NEWSPACK_AMP_PLUS_ENABLED', true );`)
4. Place a "Frequency" Donate block and a "Tiers" Donate block on a page
5. In a fresh session (logged out) select an option and confirm you are able to donate as shown in the videos above
6. If you have RAS enabled, confirm the "Sign In" header link changed to "My Account"
7. Click on the page and confirm you see your order
8. Back to the page with the Donate blocks, select to donate again, and confirm the billing fields are no longer showing, like in the video above
9. Switch AMP Plus off to use vanilla AMP, click to donate, and confirm you're redirected to the regular checkout

### Other information:


includes/class-newspack-blocks.php
```diff
@@ -365,9 +366,6 @@ public static function enqueue_view_assets( $type ) {
 				NEWSPACK_BLOCKS__VERSION
 			);
 		}
-		if ( static::is_amp() ) {
-			return;
-		}
```
miguelpeixe, line 370: I'm removing this check so that the `newspack-block-donate` script can get enqueued through AMP Plus. We can leave for AMP to remove the scripts instead of preemptively acting on it.

src/blocks/donate/frontend/class-newspack-blocks-donate-renderer.php
```diff
@@ -87,6 +189,12 @@ public static function render( $attributes ) {
 		}
 
 		Newspack_Blocks::enqueue_view_assets( 'donate' );
+		wp_script_add_data( 'newspack-blocks-donate', 'async', true );
+		wp_script_add_data( 'newspack-blocks-donate', 'amp-plus', true );
+
+		if ( true === $attributes['useModalCheckout'] ) {
```
dkoo, line 195: Should we make this configurable per block since it's already an attribute?

src/blocks/donate/frontend/class-newspack-blocks-donate-renderer.php
```diff
@@ -14,6 +14,108 @@
...
+	}
+
+	/**
+	 * Modify fields for modal checkout.
+	 *
+	 * @param array $fields Checkout fields.
+	 *
+	 * @return array
+	 */
+	public static function woocommerce_checkout_fields( $fields ) {
```
dkoo, line 62: Seems like it will change the billing fields on sites that are already using WooCommerce, which may not be desirable.

dkoo (changes requested):
This is working really well, does a ton to remove friction from the Woo reader checkout flow, and the code looks very clean to me! 🙌  

The only thing I'd caution is that once this is merged, it will affect any site using the Newspack Reader Revenue platform. This means:

1. Any AMP Plus or non-AMP site will start to see the modal experience as long as they meet the technical criteria.
2. Any site using Newspack as the Reader Revenue platform will get the more limited set of default fields (email + first/last name) in the checkout form, whether or not it's shown in the modal.

To avoid surprises, I'd suggest we take some measures to make the modal experience opt-in for now. This could mean:

* Adding a global option to the Reader Revenue wizard to turn on the modal experience for all blocks, disabled by default
* Setting the `useModalCheckout` attribute to `false` by default and adding UI in the block attributes to turn it on per block (could also be useful in combination with a global setting)
* (less desirable) gating it behind yet another experimental environment flag

I think ideally, we should offer some kind of UI to choose the fields that are available in the checkout form. For now, we can stick with the limited first/last name + email as long as the feature is opt-in and the limited fieldset is communicated in the UI used to opt in. The configurable fields can become a future feature request if enough publishers want the modal experience and the ability to choose custom fields (I suspect many will).

adekbadek:
> When authenticated, the billing fields are removed in order to provide a more straight-forward experience
> While unauthenticated, the checkout form includes the fields "First Name", "Last Name" and "Email" by default. 

I second @dkoo 's concerns voiced above. I think this change takes the feature too far, requiring us to put some guards in place (options, flags, etc). I think the modal checkout should be **completely unopinionated as to the checkout form** itself* – there are other plugins and settings to manipulate what's in the checkout form. 

If the change here would be _only_ pertaining to UX (modal instead of redirect), we could roll this out without introducing more settings and thus simplify things and avoid increasing the complexity. 

\* with one exception – if authenticated, the relevant fields could be prefilled

miguelpeixe:
Thank you for the review and comments!

> I think the modal checkout should be completely unopinionated as to the checkout form itself* – there are other plugins and settings to manipulate what's in the checkout form.

I don't think it should be the case for a checkout that we know to be generated by our block, which [is the case here`](https://github.com/Automattic/newspack-blocks/blob/9b5e9474bc6eebb6744e4e7c46a7781d4c1ff0c7/src/blocks/donate/frontend/class-newspack-blocks-donate-renderer.php#L63-L68). This template and field definition is not applied to other WC checkouts.

We need room to adapt the checkout to provide a good experience. The standard fields result in a poor modal and should be avoided.

> ideally, we should offer some kind of UI to choose the fields that are available in the checkout form

I agree that the wizard should provide the setting to choose the fields available in this form. I'll start working on it now.

miguelpeixe:
https://github.com/Automattic/newspack-plugin/pull/2256/commits/b5c1d4936cd0e90c47211bbeeb8f2eee8bd2d97d introduces customizable billing fields through the Reader Revenue wizard. The PR description has been updated.

src/blocks/donate/frontend/class-newspack-blocks-donate-renderer.php
dkoo: Yes, my concern is that some publishers (mainly non-profits) need the addresses and phone numbers for donors, so we shouldn't push anything that would cause donations to be collected without that data, unknowingly. A global opt-in switch plus configurable fields would mitigate this concern, as discussed.

dkoo:
@miguelpeixe as we discussed separately, I think we need a single global toggle in the Reader Revenue wizard to opt into the whole feature set. Once the feature is toggled on, we can show the UI to choose billing fields and default to the more limited set, as publishers who need the address and phone number will have an opportunity to choose them as soon as they opt into the feature.

adekbadek:
> my concern is that some publishers (mainly non-profits) need the addresses and phone numbers for donors, so we shouldn't push anything that would cause donations to be collected without that data, unknowingly.

> I think we need a single global toggle

With the addition of customizable billing fields, I think the best rollout path would be to make them all enabled by default. Instead of a global-feature-toggle, the publisher would opt-in to the more simplified checkout by toggling off the fields which they don't need. 

This way we don't increase the settings surface (another toggle!) and UI complexity.

miguelpeixe:
> This way we don't increase the settings surface (another toggle!) and UI complexity.

I also think that having an all-fields default is better than adding another toggle to the UI.

miguelpeixe:
https://github.com/Automattic/newspack-plugin/pull/2256/commits/9526e0632f4dd07c72426adacde6faca32d468c3 adds a custom thank you message for donations and f46300b0815eb43bd36bf549269bb459f1d4d56 tweaks the return URL after checkout so it preserves the order information:

image

miguelpeixe:
A few more tweaks on 65603be3627c2dc19b096e5c3d7da0d82b7795d5:

- The template is simplified and uses the default form (either from theme or WC)
- Selected billing fields should now be displayed when unfilled (for authenticated users)
- Email is always rendered for authenticated users

src/blocks/donate/frontend/class-newspack-blocks-donate-renderer.php
dkoo: This has been resolved by making our default fields match WooCommerce's, so no changes in billing fields should occur when the feature is enabled.

dkoo (changes requested):
> The template is simplified and uses the default form (either from theme or WC) instead of replicating the templates unnecessarily

Perhaps related to this change, the thank you page now shows the full site layout including header, nav, sidebars, footer, etc. inside the modal:

Screen Shot 2023-02-02 at 3 29 23 PM

miguelpeixe:
Thanks, @dkoo !

Indeed, the change in the templates caused this issue, which I believe to have fixed with 58b551e3af25e58e74bb06058b02431104f1a690. Maybe you tested without the latest pull?

dkoo:
> Indeed, the change in the templates caused this issue, which I believe to have fixed with https://github.com/Automattic/newspack-blocks/commit/58b551e3af25e58e74bb06058b02431104f1a690. Maybe you tested without the latest pull?

It's working now after `npm ci`ing. Thanks!

dkoo (approved):

dkoo:
Approved from an engineering standpoint, but we should probably await a design review before merging.

thomasguillot:
@miguelpeixe I pushed some (small) changes.
Ideally we would want to split this modal into different steps because it's super tall on the screen (and even taller on mobile).
Do you think we could do something like this?
Screenshot 2023-02-06 at 16 58 26

miguelpeixe:
Love the idea, @thomasguillot! Thank you for the style tweaks as well.

It's implemented in f3718af9be0f3e3a3825d98f7162b8cefae401db:

| Step 1 | Step 2 |
| --- | --- |
| image | image |

In case the customer already has the required billing fields filled, step 1 is automatically skipped when opening the modal. They have the ability to click "Edit" and go back to step 1.

thomasguillot:
Screenshot 2023-02-15 at 11 08 13

@miguelpeixe are you able to remove the wp-admin bar from the modal when a user is logged in?

miguelpeixe:
Good call! Done in c923aeda2be1e067254c6f0162fdbae2e702f8ad

miguelpeixe:
Another fix: the 2-step logic was not considering anonymous users so they would stay forever stuck on the "Billing fields" step. d3be4a36cba96e18188984b73e6b33beadb7f34b fixes the logic so logged-in and anonymous users can go through the 2-step approach.

src/blocks/donate/templates/checkout-form.php
```diff
@@ -0,0 +1,112 @@
...
+				<?php do_action( 'woocommerce_checkout_billing' ); ?>
+				<button type="submit" class="button alt wp-element-button"><?php esc_html_e( 'Continue', 'newspack-blocks' ); ?></button>
+			</div>
+		<?php else : ?>
+			<div class="checkout-billing checkout-billing-summary">
+				<div class="checkout-billing-summary__grid">
+					<h3><?php esc_html_e( 'Billing details', 'newspack-blocks' ); ?></h3>
+					<a href="<?php echo esc_url( add_query_arg( 'edit_billing', 1 ) ); ?>" class="edit-billing-link"><?php esc_html_e( 'Edit', 'newspack-blocks' ); ?></a>
+				</div>
+				<p><?php echo WC()->countries->get_formatted_address( $form_billing_fields ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?></p>
```
dkoo, line 98: Nit: this doesn't seem to output the email address, but we might want to display that, too, right?

src/blocks/donate/templates/checkout-form.php
```diff
@@ -0,0 +1,112 @@
...
+		<?php if ( $edit_billing ) : ?>
+			<div class="checkout-billing">
+				<?php do_action( 'woocommerce_checkout_billing' ); ?>
+				<button type="submit" class="button alt wp-element-button"><?php esc_html_e( 'Continue', 'newspack-blocks' ); ?></button>
+			</div>
+		<?php else : ?>
+			<div class="checkout-billing checkout-billing-summary">
+				<div class="checkout-billing-summary__grid">
+					<h3><?php esc_html_e( 'Billing details', 'newspack-blocks' ); ?></h3>
+					<a href="<?php echo esc_url( add_query_arg( 'edit_billing', 1 ) ); ?>" class="edit-billing-link"><?php esc_html_e( 'Edit', 'newspack-blocks' ); ?></a>
```
dkoo, line 96: When clicking "Edit", any fields I had filled out in the previous step but which haven't yet been saved to my user account (e.g. before making the purchase) are emptied out again.  Is there any way to pre-fill these based on the prior user input? Otherwise "edit" isn't really accurate...it's more "redo" 😄

dkoo: Another observation here. I say "observation" because I'm not sure we need to do anything about it. But if you make a donation from a logged-in state with an account that already has an email address, it's now possible to click "edit" and change the email address to something else. That new email address will be associated with the order/subscription, but won't change the email address associated with the account itself. I wonder if that might cause confusion for readers who think they're updating the email address associated with their account (which we don't allow for RAS readers).

I suppose if it's possible to `disable` the email input field in this scenario, that might be safest, but it's a behavior that's allowed by WooCommerce, so I'm not sure.

dkoo (changes requested):
The two-step flow is pretty slick for checkouts that have a lot of fields, but part of me misses the really low-friction modal experience when there are only a couple of fields. Maybe something to revisit in a future PR if lots of publishers limit the field set.

I also saw some behavior resulting from the new "Edit" capability, which isn't really buggy behavior but could be minorly annoying to readers. I noted them below, but they're minor, so if anything will take a lot of effort I'd say they shouldn't be blockers.

thomasguillot:
> The two-step flow is pretty slick for checkouts that have a lot of fields, but part of me misses the really low-friction modal experience when there are only a couple of fields. Maybe something to revisit in a future PR if lots of publishers limit the field set.

This is true and I can see a case for it.
I wonder if we could have an option in Newspack Plugin to enable/disable the 2-step flow

dkoo:
>  I wonder if we could have an option in Newspack Plugin to enable/disable the 2-step flow

Agreed, that seems to be the best way to accommodate both cases. However, I don't think it's a blocker for this PR.

src/blocks/donate/templates/checkout-form.php
miguelpeixe: > That new email address will be associated with the order/subscription, but won't change the email address associated with the account itself.

I'm assuming this is expected behavior because that's how WC handles it as well. If you change the billing email address on checkout it will not have any effect outside of that order.

src/blocks/donate/templates/checkout-form.php
miguelpeixe: Good call! Done in fff44dce5c904df5069ed35a77f04e8c5468b1b5

src/blocks/donate/templates/checkout-form.php
dkoo: > I'm assuming this is expected behavior because that's how WC handles it as well. If you change the billing email address on checkout it will not have any effect outside of that order.

I figured it was expected behavior in Woo. It's just now more likely to happen due to our modal checkout flow giving users ability to edit the email field more easily.

miguelpeixe:
I agree the 2-step flow can be optional and tackled in a separate PR!

Other than the requested tweaks, a couple of others:

 - 2c224b9dda46f2b7b1990feb0fd837b9bc363f98: the modal was being attached on every donate block (SDB). This ensures that only `wc` blocks get the modal feature.
 - 4334d6e85e9e33246fcd2dea78dd7e0f9573d67f: if a reader registers through other means (not WC), their email is not automatically added as the billing email address. This changes it for the modal checkout in order to skip the 1st step of the flow

src/blocks/donate/templates/checkout-form.php
miguelpeixe: We can enforce it to be `readonly`, but the original WC form makes the email input editable as well.

dkoo (approved):
Thanks for the revisions on this—it's looking great!

PR #1363, opened 2023-02-01
fix: remove mark background in Newspack blocks

### LINKED ISSUE #1285: Donate Block: Highlight text styles on button add yellow background
**Describe the bug**
This is a bit of an edge case but: if you use the inline 'Highlight' option to change the text colour in the Donate Block button, the text will have a yellow background on the front-end due to the `mark` tags the block editor uses to add inline styles.

In other blocks, an inline transparent background is added to the `mark` tag when Highlight is used. 

A light yellow background is added to `mark` by the theme, but a brighter yellow is typically the browser default for the tag -- so I opted to file the issue here instead of in the theme repo. 

**To Reproduce**
1. Add a Donate block.
2. Select the text and open the little 'down arrow' more options menu:

image

3. Click 'Highlight' and change the text colour -- either to something in the editor palette, or something custom: 

image

5. Publish and view on the front-end.

Ran into this on a publisher's site when they wanted to change the text from black to white -- they were using one of those in-between button colours where either would technically have sufficient contrast, but the colour code picked black (this happens in the theme as well! The contrast logic is a little too black and white... 😄 ). 

Edited to add: It could be worth adding a text colour option to the block, too, instead of relying on the contrast logic --- it's a case where it's a more hands-on change (vs. changing ALL the buttons across the site) and there's a good handful of publishers we have where we're changing the button text to be the opposite of what the theme's doing, because they're in the same middle ground. It also syncs up the behaviour/options with a standard button block.

**Screenshots**
Donate block with the button text changed using the 'Highlight' option:

image

Styles added to the button:

image

Example of the same kind of styles applied to a paragraph block -- a transparent background is added with the text style:

image

laurelfulford:
### All Submissions:

This is a long explanation for a tiny code change:

**Context:** The RichText fields in Gutenberg allow some minor text formatting -- bold, italic, link, highlights (text and background colours). The highlights use the `mark` tag. 

The `mark` tag typically has some standard styling in HTML (bright yellow by default in most browsers; kind of a mellow-er yellow in our theme). 

In most cases, Gutenberg works around that by adding `style="background-color: rgba(0, 0, 0, 0)"` to `mark` tags that don't have a background colour. 

Publishers have very occasionally been using the highlight option to change the section title of the Homepage Posts block, or the button text colour in the Donate block. 

**The issue:** In both blocks, we're sanitizing the strings on the PHP side using `wp_kses_post`, which is stripping out `rgba()` (more info [here](https://core.trac.wordpress.org/ticket/24157)). This causes the `mark` tag to display the yellow background. Though most Gutenberg blocks appear not to have this issue, the occasional one that uses `wp_kses_post` for sanitizing text -- like the search block's button -- display the same issue.

**The fix:** The work around here is kind of hacky, but quick: rather than trying to get `rgba()` allowed in the blocks, or rewriting the sanitization, I opted to override the `mark` background colour when the HTML tag is used in these blocks. This was added to the blocks rather than the theme because this would be an issue with any theme (unless that theme is overriding the browser default and removing the `mark` background). 

I'm definitely open to feedback on this approach!

Closes #1285; see 1203695565716937-as-1203423790806450

1. Add a Homepage Posts block and give it a 'Section title'.
2. Highlight some of the section title and using the caret dropdown in the block toolbar, select Highlight:

image

3. Change the text colour.
4. Repeat steps 1-3, but change the text background instead -- this is just to confirm this still works.
5. Add a Donate block.
6. Highlight some of the Donate Now button text, and give it a text colour.
7. Highlight some of the Thank you text, and give it a text colour.
8. Repeat steps 5-7, but instead of giving each section of text a background colour.
9. Publish and view on the front-end.
10. Note the light yellow highlight behind the text colour that was changed; your background colour should be applied as expected. 

image

image

11. Apply the PR and run `npm run build`.
12. Confirm the light yellow highlight is gone from your text colour highlights without impacting your background colour highlights.

image

image

### Other information:

### DIFF
--- a/src/blocks/donate/styles/view.scss
+++ b/src/blocks/donate/styles/view.scss
@@ -17,4 +17,8 @@
 			color: colors.$color__text-main;
 		}
 	}
+
+	mark {
+		background-color: transparent;
+	}
 }
--- a/src/blocks/homepage-articles/view.scss
+++ b/src/blocks/homepage-articles/view.scss
@@ -50,6 +50,10 @@
 		font-size: variables.$font__size-sm;
 		margin-bottom: 0.5em;
 		width: 100%; // make sure this isn't caught up in the flex styles.
+
+		mark {
+			background-color: transparent;
+		}
 	}
 
 	/* Column styles */


dkoo (approved):
Nice, simple fix for what seems like a pretty vexing issue. 👍

laurelfulford:
Thanks @dkoo! Yeah, I fell down a bit of a rabbit hole with this one... 😅

PR #1377, opened 2023-02-21
Homepage Post patterns: remove unneeded group & columns blocks

laurelfulford:
### All Submissions:

I removed the containing Group block from all patterns but 16, 18, and 20, since the group blocks are used to add a background colour. From those, I just removed the custom CSS classes for consistency. 

For patterns For 4, 11, 13, 23, 27 and 28, I removed the Columns block and replaced it with the Homepage Posts grid option.

There are still some patterns that have columns blocks, but to me they read more intentional (eg. it felt like the different columns were intended to act as different topic sections). That being said, I wonder if we should add some section header placeholders to help add some more context around how they should be used?

See 1203869877018965-as-1203928233002231

1. Apply the PR. 
2. Check the Newspack Homepage Patterns by adding each to the editor; confirm:
   * That they still match the screenshots 
   * That they all can be added without error
   * That none but 16, 18, or 20 include a group block.

### Other information:


adekbadek (approved):
> That being said, I wonder if we should add some section header placeholders to help add some more context around how they should be used?

I think that's a good idea, it would help discoverability of the section header feature.

> That they still match the screenshots

Just to be clear, there are no screenshots – the previews are dynamically generated by Gutenberg. Or am I misunderstanding something?

laurelfulford:
Thanks @adekbadek! 

> I think that's a good idea, it would help discoverability of the section header feature.

Cool -- I think so too! I've spun up a new issue for myself here: 1200550061930446-as-1204034280172662

> Just to be clear, there are no screenshots...

Oh god 🤦‍♀️ I _knew_ that, too -- I was watching them load, HTML then CSS, live in the patterns sidebar -- and I was still checking to make sure the inserted pattern matched the preview.

PR #1390, opened 2023-03-07
help.newspack.com doc links

gamebits:
Updates all newspack.pub documentation references in code to help.newspack.com.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,4 +40,4 @@ To get started with a new block:
 
 ### Usage
 
-[End-user documentation can be found here](https://newspack.pub/support/blocks/homepage-posts/).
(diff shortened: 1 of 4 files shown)


leogermani (approved):
I've checked that the new URLs exist and they look good.

Not related to this PR but do we really want these URLs to be i18n ? And if so, the text domain is missing from the `__()` calls...

I think we could drop the usage of i18n there, since we don't have a standardized way of offering different URLs for different languages, nor we have the documentation in different language...

miguelpeixe:
The diff breaks lint as well. The linter is probably requiring a line break due to the increase in the number of characters.

leogermani:
> The diff breaks lint as well. The linter is probably requiring a line break due to the increase in the number of characters.

I went ahead and fixed it

miguelpeixe (approved):
🙌

PR #1422, opened 2023-04-07
fix: bugs for Post Carousel

dkoo:
### All Submissions:

Fixes an editor bug and a front-end bug.

Closes `1204166503564710/1204293714861357`.

Editor bug:

1. Add a Post Carousel to a post in the editor. Change the aspect ratio and/or slides per view options from the defaults.
2. Update the query params for the carousel block—e.g. add a category or tag. On `master`, observe that the block preview updates with the new posts, but the carousel itself becomes non-functional and doesn't show your aspect ratio or slides per view settings until you change another attribute. This is because the Swiper library operates outside of React and doesn't trigger a component re-render by itself.
3. Check out this branch, repeat, confirm that the carousel remains interactive after updating the post query. The fix ties the initialization of a Swiper instance to state data so that it triggers a re-render every time Swiper initializes a new instance.

Front-end bug:

1. Add a Post Carousel to a Campaigns prompt. Change the aspect ratio and/or slides per view options from the defaults. (Not required, but makes it easier to see when the carousel is initialized.)
2. View the prompt on the front-end on `master`. Observe that the carousel never becomes interactive and your aspect ratio/slides per view options are never reflected. This is because Swiper executes immediately on `domReady` but won't init elements that aren't visible in the DOM, and the prompts are hidden on `domReady` until the segmentation logic completes asynchronously.
3. Check out this branch and repeat step 2. This time confirm that the carousel initializes as soon as the prompt becomes visible. The fix ties the Swiper init method to an `IntersectionObserver` for each block instance so that the carousel isn't initialized until/unless it becomes at least 25% visible.

### Other information:

### DIFF
--- a/src/blocks/carousel/edit.js
+++ b/src/blocks/carousel/edit.js
@@ -99,6 +99,7 @@ class Edit extends Component {
 				if ( latestPosts && this.swiperInstance.realIndex < latestPosts.length ) {
 					initialSlide = this.swiperInstance.realIndex;
 				}
+				this.setState( { swiperInitialized: false } );
 				this.swiperInstance.destroy( true, true );
 			}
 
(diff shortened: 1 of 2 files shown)


laurelfulford (approved):
Both changes look good to me, @dkoo!

The carousel's style settings no longer change when when the query is updated in the editor. And the carousel appears to function as normal again when embedded in an campaign -- I tried both inline and overlay, and things like the slides-per-"page" setting and the previous/next controls worked fine. 🚢

Issue #858, opened 2021-09-03
Homepage Posts/Post Carousel: Allow filtering by custom taxonomy terms


dkoo:
**Is your feature request related to a problem? Please describe.**
The Homepage Posts and Post Carousel blocks allow editors to choose additional post types to display in block content:

<img width="176" alt="Screen Shot 2021-09-03 at 3 23 12 PM" src="https://user-images.githubusercontent.com/2230142/132066594-2b3c047e-cfa6-40fb-aa3b-b19cd0816d98.png">

However, the filtering options currently only support filtering by post categories and tags (Homepage Posts also allows excluding by those same taxonomies):

<img width="265" alt="Screen Shot 2021-09-03 at 3 23 40 PM" src="https://user-images.githubusercontent.com/2230142/132066605-fd16cff9-a063-4e67-bd89-40f160fe03c8.png">

This results in a situation where it's possible to choose a CPT to be shown in these blocks, but if the CPT doesn't use post categories or tags, it's not possible to filter the query further by the taxonomies the CPT actually does use. A concrete example is the recent GDG launch, where they needed to be able to show Event post types in a Post Carousel block and filter them by a "featured" category. However, the Event CPT on that site uses its own custom taxonomy, not regular post categories or tags, so this is currently not possible.

**Describe the solution you'd like**
Ideally, we should be able to filter the block's query by any taxonomies associated with the CPTs selected under Post Types. I'm not sure what this should look like in the block UI—since it's kind of an edge case, maybe we could combine all custom taxonomies into a single Autocomplete field? This might result in a confusing or difficult to use UI, though, especially if terms from multiple custom taxonomies have similar names. cc @thomasguillot

**Describe alternatives you've considered**
Open to other ideas on how to handle this, but I can definitely see this as being a useful feature, especially since we provide the `newspack_blocks` post type support handle to let other plugin devs add support for Homepage Posts and Post Carousel blocks to any arbitrary post type.

sophiegyo:
I was testing this on a simple free WordPress.com site and experienced it too, it simply wasn't possible to filter by portfolio or testmonial taxonomies. If this could be allowed it would be super appreciated :)

dkoo:
Thanks for the feedback on this! It is definitely on our radar. :)

PR #1427, opened 2023-04-11
Add brand query support to the blocks

laurelfulford:
### All Submissions:

[Edited to add all of this after @leogermani's review! Basically copied over some comments about this PR from Slack 🙂 ]

This PR adds the 'Brands' taxonomy as an option for querying to the Homepage Posts and Post Carousel blocks. 

See 1204179187018802-as-1204179187293071

This is a pretty literal set up, just adding the grabbing the Brands taxonomy. We do have an existing issue for allowing more [general custom taxonomy support](https://github.com/Automattic/newspack-blocks/issues/858); this could be a good opportunity to add that by abstracting these changes.  

2. Confirm there are no issues/errors on the front-end or in the editor when you add a homepage posts block.
3. Install and activate the [Multi-Branded Site](https://github.com/automattic/newspack-multibranded-site) plugin
4. Set up some brands and apply them to stories
5. Add a homepage posts block to a page; confirm you now have a 'Brands' field in with the Category and Tags query options.
6. Use it to pick one of your brands.
7. Confirm that the block is showing the expect stories in the editor and on the front-end. 
8. Repeat the same steps for the Post Carousel block.
9. Deactivate the Multi-Branded Site plugin, and confirm there are no issues with the blocks in the editor and the front-end, that they no longer query the Brands or throw errors.
10. Confirm that the Brands code is loading when it should be - I did leave some const definitions in place even when the plugin isn't active; I wasn't sure if they should always be wrapped in a check for the plugin or if it was ok to leave them as is! 

### Other information:

### DIFF
--- a/includes/class-newspack-blocks.php
+++ b/includes/class-newspack-blocks.php
@@ -587,11 +587,13 @@ class Newspack_Blocks {
 		$authors             = isset( $attributes['authors'] ) ? $attributes['authors'] : array();
 		$categories          = isset( $attributes['categories'] ) ? $attributes['categories'] : array();
 		$tags                = isset( $attributes['tags'] ) ? $attributes['tags'] : array();
+		$brands              = isset( $attributes['brands'] ) ? $attributes['brands'] : array();
 		$tag_exclusions      = isset( $attributes['tagExclusions'] ) ? $attributes['tagExclusions'] : array();
 		$category_exclusions = isset( $attributes['categoryExclusions'] ) ? $attributes['categoryExclusions'] : array();
 		$specific_posts      = isset( $attributes['specificPosts'] ) ? $attributes['specificPosts'] : array();
 		$posts_to_show       = intval( $attributes['postsToShow'] );
 		$specific_mode       = isset( $attributes['specificMode'] ) ? intval( $attributes['specificMode'] ) : false;
+
 		$args                = array(
 			'post_type'           => $post_type,
 			'post_status'         => $included_post_statuses,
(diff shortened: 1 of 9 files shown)


leogermani (approved):
@laurelfulford Thanks for this.

I confirm this works as expected and I think it's fine to have brands hardcoded for now. If we decide to abstract this to other custom taxonomies we can work on the top of what you've done.

I also confirmed the block still works with the multibranded plugin deactivated

The nitpick I have is that I think it would be better to have the brands as a list of checkboxes instead of an autocomplete... but only if it's an easy change... for now this does what we need.

Maybe @miguelpeixe or @dkoo  can also share their thoughts. If it's best to do this in an abstract way for any custom taxonomy.

(the blocks plugin could offer a filter with an array of additional taxonomies that should be added)

laurelfulford:
Thanks @leogermani! I've marked the PR as ready for review and have added some information to the description (including a link to the custom taxonomy support feature request) 🙂 

I'll follow up with @miguelpeixe and @dkoo next week re: adding more general support! I'd also be interested for more feedback on the field format -- it matches how we're handling the other queries (category and tags), so I think it may be better this way just for consistency, but I can also see how it'd be handy to actually be able to see the list vs. having to know enough of the different brands off the top of your head.

miguelpeixe:
I agree with Leo, we don't need to address the broader custom taxonomy support in this PR.

We could create a separate task for this, which could also better handle the UI for each taxonomy type. List with a checkbox for hierarchical and autocomplete search for non-hierarchical. This would be aligned with WP's approach to this and how the brand taxonomy is constructed.

PR #1454, opened 2023-05-19
Add custom taxonomies support 

leogermani:
### All Submissions:

Replaces the hard-coded support to the `brands` taxonomy with a generic support to any registered custom taxonomy.

1. On a site without multi-branded plugin, confirm the homepage block and the carousel block still work as expected
2. Activate the multi-branded plugin
3. Confirm you now see the option to filter by brands in both blocks
4. Confirm the filtering works as expected
5. Confirm that the post shows the correct posts in the front end
6. Confirm the articles have a corresponding `$taxonomy-$term-slug` class in the HTML element
7. Confirm the content of the block is updated on the fly when you make changes
8. Register another custom taxonomy and make sure it works as expected:
7; Also confirm that if the custom taxonomy is not public or if `show_in_rest` is set to `false` it won't show up.

```PHP
add_action( 'init', '__register_custom_tax' );
function __register_custom_tax() {
register_taxonomy( 'genre', 'book', array(
       'public' => true,
       'show_in_rest' => true,
    ) );
}

```

### Other information:

### DIFF
--- a/includes/class-newspack-blocks.php
+++ b/includes/class-newspack-blocks.php
@@ -209,16 +209,22 @@ class Newspack_Blocks {
 	 */
 	public static function get_custom_taxonomies() {
 		$custom_taxonomies = array_map(
-			function( $tax_slug ) {
-				$tax = get_taxonomy( $tax_slug );
-				if ( ! empty( array_intersect( [ 'post', 'page' ], $tax->object_type ) ) && ! $tax->_builtin && $tax->show_in_rest ) {
+			function( $tax ) {
+				if ( ! empty( array_intersect( [ 'post', 'page' ], $tax->object_type ) ) ) {
 					return [
-						'slug'  => $tax_slug,
+						'slug'  => $tax->name,
 						'label' => $tax->label,
 					];
 				};
 			},
-			get_taxonomies( [ 'public' => true ] )
+			get_taxonomies(
+				[
+					'public'       => true,
+					'_builtin'     => false,
+					'show_in_rest' => true,
+				],
+				'objects'
+			)
 		);
 		$custom_taxonomies = array_values(
 			array_filter(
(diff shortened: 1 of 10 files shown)


includes/class-newspack-blocks.php
```diff
@@ -202,6 +202,43 @@ public static function enqueue_placeholder_blocks_assets() {
...
+			function( $tax_slug ) {
+				$tax = get_taxonomy( $tax_slug );
+				if ( ! empty( array_intersect( [ 'post', 'page' ], $tax->object_type ) ) && ! $tax->_builtin && $tax->show_in_rest ) {
+					return [
+						'slug'  => $tax_slug,
+						'label' => $tax->label,
+					];
+				};
+			},
+			get_taxonomies( [ 'public' => true ] )
```
miguelpeixe, line 221: Can't the args passed to `get_taxonomies()` include the `_builtin` and `show_in_rest` options? You can also [return objects by setting the second parameter as `true`](https://developer.wordpress.org/reference/functions/get_taxonomies/#parameters) so you don't have to `get_taxonomy()` inside the array map.

includes/class-newspack-blocks.php
```diff
@@ -202,6 +202,43 @@ public static function enqueue_placeholder_blocks_assets() {
...
+				}
+			)
+		);
+
+		/**
+		 * Filters the custom taxonomies that will be available in the Home Page block.
+		 *
+		 * By default, on the top of category and tags, will display any public taxonomy applied to post or pages
+		 *
+		 * @param $custom_taxonomies array Array of custom taxonomies where each taxonomy is an array with slug and label keys.
```
miguelpeixe, line 237:
```suggestion
		 * @param array $custom_taxonomies Array of custom taxonomies where each taxonomy is an array with slug and label keys.
```

Small typo

miguelpeixe (changes requested):
Works great! Just some small fixes below.

I noticed you're preventing non-post/page taxonomies in the mix, but the block supports post type selection. I wonder if it should return all public taxonomies and inside the editor decide which options to render based on the selected post types for the block.

includes/class-newspack-blocks.php
leogermani: Thanks, fixed in https://github.com/Automattic/newspack-blocks/pull/1454/commits/ffe1780a5651f457b7f3036f38fca3e174e1639e

leogermani:
> I wonder if it should return all public taxonomies and inside the editor decide which options to render based on the selected post types for the block

I think this could be a nice improvement, but for now this will cover most of the cases I believe.

miguelpeixe (approved):
Thank you for the revision! It's working beautifully 🙌

PR #1486, opened 2023-06-26
feat: add updated translation files

laurelfulford:
### All Submissions:

This PR is a long-overdue update to the plugin's POT file, and the PO, MO and translation JSON files for the supported languages in the blocks. 

See 1200550061930446-as-1204849686104010

This PR adds a lot of new files (a new JSON file for each new JS file), but no additional translations. I think the best way to review is to review the changed code, and do a few spot checks with the blocks, but I'm also open to any other testing you can do!

1. Apply the PR. 
2. Switch your test site to one of the languages that have been translated in the plugin (German (de_DE), Spanish (es_ES), Belgian French (fr_BE), or Portuguese (pt_PT)).
3. Add a couple blocks to the editor to make sure they're getting translated. This will only be the case for some of the older blocks -- the newer ones haven't been translated at all yet -- and some strings will be missing. A good couple options are the Donate Block, and the Homepage Posts block.
4. Publish your page, and view in the editor and on the front-end:   
    * For the Donate block, all of the tabs and 'Donation Amount' should be translated; 'Your contribution...' and 'Donate Now' will not be due to limitations in pulling these strings from the block.json to translate. The upside is they can be translated by editing the block, and that this doesn't introduce a regression -- the translation already stopped working in `master`. 
    * For the Homepage Post block, the byline string (by) should be translated, and the date should be automatically localized. 
5. Repeat this for at least one more language. 
6. Manually review changes in the PR. There are a LOT of additions that can be ignored -- the big one to check for is things that have been removed:
     * Some changes are expected -- some strings have changed, or have been removed outright (like the aformentioned 'Donate' ones). 
     * In some cases, some strings appear to have been removed from the JSON files - a couple examples include the 'keywords', or the block description. If you manually check the block though, the translations still work -- one example is the Homepage Post block description in Belgian French. I'm not sure why this is happening but am more concerned that the translations continue to work as they do on `master` instead of how they look in the code -- especially on the front-end. 

Please ping me if I can help answer any questions, or verify any testing! This one is a bit of a monster; I don't think it's necessary to manually check every changed string, only verify that the odd one still works as expected. 

### Other information:


dkoo (approved):
Seems to be working on my end—I didn't test exhaustively, but I see the front-end strings being translated...

laurelfulford:
Thanks @dkoo! 🙌

PR #1516, opened 2023-08-23
fix: support account creation on modal checkout

dkoo:
### All Submissions:

Fixes the modal checkout flow when WooCommerce settings are set to allow account creation and password setting during checkout.

Also adds a few accessibility and style improvements to the modal checkout flow:

- Allows submitting part 1 (billing details) of the form with and "enter" keypress
- Shows the spinner when submitting part 1 of the form, to provide visual feedback that the form is loading another part
- Fully hides the order review table if we're not supposed to see it (if there are no coupons or taxes to calculate), so that we don't see it flash on screen before Woo hides it via JS

Some caveats/discussion points:

- Enabling this account creation flow during checkout allows readers to bypass the RAS account verification flow, as the reader account will be generated with a password but without verifying ownership of the email address. Should we be concerned about this?
- The account creation fields will warn you about weak passwords, but this doesn't prevent you from submitting the form and creating an account with a weak password. This seems to be WooCommerce behavior, but it caught me a little off guard.

Closes `1200550061930446/1205326345667190`.

1. On `master`, in WooCommerce settings > Accounts & Privacy, enable the "Allow customers to create an account during checkout" option and disable the "When creating an account, automatically generate an account username for the customer based on their name, surname or email" and "When creating an account, send the new user a link to set their password" options. Like this:

Screenshot 2023-08-23 at 5 49 38 PM

2. Start a donation with the modal checkout and observe that the "Create an account" section with username and password fields appear in the first part of the modal checkout flow.
3. Fill out all fields and click "Continue" to proceed to part 2.
4. Complete part 2 of the checkout flow and observe errors preventing you from submitting the form.
5. Check out this branch and repeat steps 2-4. This time confirm that the "Create an account" section and fields appear on part 2 of the checkout flow, and that you're able to complete the transaction.
6. After completing the transaction, start a new session and confirm you're able to log into the account using the email + password you set during step 5, and that the display name matches what you entered during checkout.

### Other information:

### DIFF
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -198,13 +198,13 @@ final class Modal_Checkout {
 					<div class="newspack-blocks-variation-modal__selection" data-product-id="<?php echo esc_attr( $product_id ); ?>">
 						<h3><?php echo esc_html( $product->get_name() ); ?></h3>
 						<p><?php esc_html_e( 'Select an option below:', 'newspack-blocks' ); ?></p>
-					<?php
-					$variations = $product->get_available_variations( 'objects' );
-					foreach ( $variations as $variation ) {
-						$name        = $variation->get_formatted_variation_attributes( true );
-						$price       = $variation->get_price_html();
-						$description = $variation->get_variation_description();
-						?>
+						<?php
+						$variations = $product->get_available_variations( 'objects' );
+						foreach ( $variations as $variation ) {
+							$name        = $variation->get_formatted_variation_attributes( true );
+							$price       = $variation->get_price_html();
+							$description = $variation->get_variation_description();
+							?>
 							<form>
 								<input type="hidden" name="newspack_checkout" value="1" />
 								<input type="hidden" name="product_id" value="<?php echo esc_attr( $variation->get_id() ); ?>" />
@@ -219,8 +219,8 @@ final class Modal_Checkout {
 								</button>
 							</form>
 							<?php
-					}
-					?>
+						}
+						?>
 					</div>
 				</div>
 			</div>
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -213,7 +213,7 @@ final class Modal_Checkout {
 										<span class="price"><?php echo $price; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?></span>
 										<span class="variation_name"><?php echo esc_html( $name ); ?></span>
 									</span>
-								<?php if ( ! empty( $description ) ) : ?>
+									<?php if ( ! empty( $description ) ) : ?>
 										<span class="description"><?php echo esc_html( $description ); ?></span>
 									<?php endif; ?>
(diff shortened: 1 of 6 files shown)


yogeshbeniwal:
> Enabling this account creation flow during checkout allows readers to bypass the RAS account verification flow, as the reader account will be generated with a password but without verifying ownership of the email address. Should we be concerned about this?

Should be ok. new account is created rather than pre-existing account. Adding account verification here will add friction to conversion.

RAS email verification link will be sent in parallel to enable change on my account page etc?

miguelpeixe:
> Enabling this account creation flow during checkout allows readers to bypass the RAS account verification flow, as the reader account will be generated with a password but without verifying ownership of the email address. Should we be concerned about this?

This is the existing behavior for accounts created through Woo's default customer creation flow. The only verification that remains required, and will not be affected by this change, is the one for newsletters management on the My Account page. We will continue to not allow someone to manage an email newsletter subscription of a 3rd party without verifying that they own the address.

includes/class-modal-checkout.php
```diff
@@ -198,13 +198,13 @@ public static function render_variation_selection() {
 					<div class="newspack-blocks-variation-modal__selection" data-product-id="<?php echo esc_attr( $product_id ); ?>">
 						<h3><?php echo esc_html( $product->get_name() ); ?></h3>
 						<p><?php esc_html_e( 'Select an option below:', 'newspack-blocks' ); ?></p>
-						<?php
-						$variations = $product->get_available_variations( 'objects' );
-						foreach ( $variations as $variation ) {
-							$name        = $variation->get_formatted_variation_attributes( true );
-							$price       = $variation->get_price_html();
-							$description = $variation->get_variation_description();
-							?>
```
miguelpeixe, line 207: NIT: I'm confused by the linting change here. Is this the recommended practice?

miguelpeixe (approved):
Great enhancements! It's working as described 🎉 

Left a minor linting question below.

includes/class-modal-checkout.php
dkoo: I'm not sure how that happened. fc0c80f reverts this change, but PHPCS complained about different whitespacing issues, so there is still some diff compared with `master`.

dkoo: Not sure why PHPCS is wigging out, but I reverted those changes in f834b9 and there's now no whitespacing diff in this changeset.

dkoo:
> This is the existing behavior for accounts created through Woo's default customer creation flow. The only verification that remains required, and will not be affected by this change, is the one for newsletters management on the My Account page. We will continue to not allow someone to manage an email newsletter subscription of a 3rd party without verifying that they own the address.

Thanks for confirming!

> RAS email verification link will be sent in parallel to enable change on my account page etc?

This PR doesn't change anything in when/how the RAS verification emails are sent.

dkoo:
Oops, realized this is already approved. Since the only changes since the approval are whitespacing-related, I'll just go ahead and merge.

PR #1521, opened 2023-08-30
feat: add option for a button after checkout

leogermani:
### All Submissions:

Implements an optional link to click after the transaction is complete on the modal checkout.

The button supports linking to the page referrer (which falls back to the current page URL if the referrer is empty) or a custom URL.

image

1. Checkout this branch and https://github.com/Automattic/newspack-theme/pull/2180
2. Add a checkout button to a page
3. Select the option to go to the Referrer after purchase and add a button label
4. Add a link to this page in the site menu
5. Visit an article
6. Go to the page via the menu link
7. Click the button and buy the product
8. Confirm the button shows up in the thank you message and link back to the article you were seeing
9. In a new tab, copy and paste the page URL directly into the browser so you get no referrer
10. Make the purchase again and confirm the button links to the current page
11. Edit the button and choose to go to a custom URL
12. Repeat the purchase and confirm the button now takes you to that URL
13. Edit the button again and choose not to show the button
14. Purchase again and confirm there is no button in the thank you message

### Other information:

### DIFF
--- a/src/blocks/checkout-button/edit.js
+++ b/src/blocks/checkout-button/edit.js
@@ -158,7 +158,7 @@ function RedirectAfterSuccess( props ) {
 	return (
 		<>
 			<SelectControl
-				label={ __( 'Redirect after success', 'newspack-blocks' ) }
+				label={ __( 'Post-Checkout Button', 'newspack-blocks' ) }
 				help={ __(
 					'Select whether the user should be presented with a button to navigate after a successful purchase.',
 					'newspack-blocks'
--- a/src/blocks/checkout-button/edit.js
+++ b/src/blocks/checkout-button/edit.js
@@ -160,7 +160,7 @@ function RedirectAfterSuccess( props ) {
 			<SelectControl
 				label={ __( 'Redirect after success', 'newspack-blocks' ) }
 				help={ __(
-					'Select whether the user should be presented with a button to navigate after a successfull purchase.',
+					'Select whether the user should be presented with a button to navigate after a successful purchase.',
 					'newspack-blocks'
 				) }
 				value={ attributes.afterSuccessBehavior }
--- a/src/blocks/checkout-button/edit.js
+++ b/src/blocks/checkout-button/edit.js
@@ -167,7 +167,7 @@ function RedirectAfterSuccess( props ) {
 				options={ [
 					{ label: __( 'Do not show a button', 'newspack-blocks' ), value: '' },
 					{ label: __( 'Go to a custom URL', 'newspack-blocks' ), value: 'custom' },
-					{ label: __( 'Go to the referrer page', 'newspack-blocks' ), value: 'referrer' },
+					{ label: __( 'Go to the previous page', 'newspack-blocks' ), value: 'referrer' },
 				] }
 				onChange={ value => {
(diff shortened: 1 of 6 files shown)


src/blocks/checkout-button/edit.js
```diff
@@ -153,6 +153,48 @@ function ProductControl( props ) {
...
+				label={ __( 'Redirect after success', 'newspack-blocks' ) }
+				help={ __(
+					'Select whether the user should be presented with a button to navigate after a successful purchase.',
+					'newspack-blocks'
+				) }
+				value={ attributes.afterSuccessBehavior }
+				options={ [
+					{ label: __( 'Do not show a button', 'newspack-blocks' ), value: '' },
+					{ label: __( 'Go to a custom URL', 'newspack-blocks' ), value: 'custom' },
+					{ label: __( 'Go to the referrer page', 'newspack-blocks' ), value: 'referrer' },
```
adekbadek, line 170: Suggestion: "previous page" might be more understandable for non-technical people.

adekbadek (changes requested):
When the custom URL field is never set, the button will not be displayed. To reproduce:

1. Insert the checkout button on a page, choose "Go to referrer page" option
2. Observe that the button is not rendered after checkout
3. Back in the editor, choose "Go to a custom URL" option and fill in a URL. Save and edit again, choosing "Go to referrer page". Save again.
4. Observe the button is rendered after checkout

miguelpeixe:
Do you think we should render the button linked to the current page when the URL is not filled?

adekbadek:
I think that without the URL set, the button should just close the modal. But the issue I'm describing is about something else: according to the testing instructions, if the URL is not filled and "Go to referrer page" is set, the button should be rendered and link to the referrer. And it's not, under conditions described in my review.

src/blocks/checkout-button/edit.js
miguelpeixe: Updated in 4325f94b4f83c785cc64bd00c164b8338eae766f

miguelpeixe:
Ah, sorry for the confusion. I was able to reproduce the issue.

4c514ae272b2f8fe437d9d0b591bec8d9836081d changes the block save to render the URL input if the `afterSuccessBehavior` is set, fixing the issue. I'm not rendering the inputs at all times because it causes a markup mismatch for existing blocks, requiring "attempt block recovery".

src/blocks/checkout-button/edit.js
```diff
@@ -153,6 +153,48 @@ function ProductControl( props ) {
 	);
 }
 
+function RedirectAfterSuccess( props ) {
+	const { attributes, setAttributes } = props;
+	return (
+		<>
+			<SelectControl
+				label={ __( 'Redirect after success', 'newspack-blocks' ) }
```
adekbadek, line 161: Suggestion: this wording makes me think that there will be an automatic page redirection after the purchase is completed. Shouldn't this label inform the user about the post-checkout _button_ instead?

adekbadek (approved):
Left a wording suggestion.

src/blocks/checkout-button/edit.js
miguelpeixe: Updated in 518e0c32d3b971b4ad144560a2f457a71423f4d6

PR #1534, opened 2023-09-19
fix: undefined variable error

dkoo:
### All Submissions:

Fixes a well-hidden PHP error due to an undefined variable. It looks like PHP in some cases is silently casting this undefined `$children` variable to an array, suppressing the warning unless you invoke the variable before the first `$children[] =` definition.

1. On `release`, create two categories that are children of another category and assign to some posts. Add `error_log( print_r( $children, true ) );` to right before [this line](https://github.com/Automattic/newspack-blocks/blob/master/includes/class-newspack-blocks.php#L674).
2. On another post, add a Homepage Posts block with the "include subcategories" option on and assign it to the parent category. Observe a PHP warning: `PHP Warning:  Undefined variable $children in /newspack-repos/newspack-blocks/includes/class-newspack-blocks.php on line 673`
3. Check out this branch, make sure the `error_log` is still the first line inside the `foreach ( $categories as $parent ) {
` block, and confirm that the PHP warning doesn't happen, and that the block still shows all posts in both parent and child categories (but only if the "include subcategories" option is on). Confirm on the front-end as well.

### Other information:

### DIFF
--- a/includes/class-newspack-blocks.php
+++ b/includes/class-newspack-blocks.php
@@ -670,9 +670,10 @@ class Newspack_Blocks {
 			}
 			if ( $categories && count( $categories ) ) {
 				if ( 1 === $include_subcategories ) {
+					$children = [];
 					foreach ( $categories as $parent ) {
-						$children[] = get_categories( array( 'child_of' => $parent ) );
-						foreach ( $children[0] as $child ) {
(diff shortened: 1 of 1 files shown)


laurelfulford (approved):
Confirmed I got the warning following the initial set of steps on `release`, and that the warning stopped (and was replaced with noting `$children` was an array in the error log) after I applied this update!

PR #1543, opened 2023-10-02
feat(homepage-posts): support deduplication toggling

### LINKED ISSUE #258: Article Block: Ability to show duplicate posts
Right now, when displayed on the front end, the Article block will strip out any posts that have already appeared earlier on the page to avoid duplication.

However, there are cases where the duplication would be welcome -- for example, a special feature section on the homepage may want to be 'exempt' from the no-duplication rule, so it can show the posts its set to show, even if they do appear earlier on the page. 

This will help prevent unexpected gaps, too, especially in cases where you're cherry-picking specific stories that may be showing up in another section when first published.

miguelpeixe:
### All Submissions:

1205451924709229-as-1205631504130139

Implements a toggle to allow a Homepage Posts block to be excluded to deduplication logic.

**Without deduplication, the block will not account for previously queried posts and will also not have their queried posts as part of the exclusion for the following blocks.**

image

---

| Default deduplication | With "Block 2" opted-out |
| --- | --- |
| image | image |

Closes #258.

1. Check out this branch and draft a new page with 3 Homepage Posts block, similar to the image above
2. Toggle the "Use deduplication logic" off in the second block and confirm the content is duplicate from the previous block and the third block renders the content that was previously being rendered in the edited block (same as the image above)
3. Save, visit the page, and confirm the same behavior persists
4. Edit the page and change the second block to "Choose Specific Posts"
5. Select a post from the first block and confirm the post renders without interfering with the previous block
6. Save, visit the page, and confirm the same behavior persists

### Other information:

### DIFF
--- a/includes/class-newspack-blocks.php
+++ b/includes/class-newspack-blocks.php
@@ -671,8 +671,14 @@ class Newspack_Blocks {
 		} else {
 			$args['posts_per_page'] = $posts_to_show;
 
-			$show_rendered_posts = apply_filters( 'newspack_blocks_homepage_shown_rendered_posts', false );
-			if ( $show_rendered_posts ) {
+			/**
+			 * Filters whether to use deduplication while rendering this block.
+			 *
+			 * @param bool   $deduplicate Whether to deduplicate.
+			 * @param array  $attributes  The block attributes.
+			 */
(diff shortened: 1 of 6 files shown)


miguelpeixe:
Thanks, @dkoo! I'll for [this one](https://github.com/Automattic/newspack-popups/pull/1228) to be approved before merging so it doesn't risk breaking campaigns.

PR #1485, opened 2023-06-26
feat(modal-checkout): render order details under different conditions

miguelpeixe:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-theme/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Implements a dynamic check on whether to render the order details for a modal checkout cart.

The following should render the order details table:
 - Cart has an applied coupon
 - Taxes are enabled and set to display prices excluding taxes (#1480)
 - Cart has more than one item

### How to test the changes in this Pull Request:

1. Repeat steps from #1480
2. Disable tax rates and calculations on WC settings (or switch the "display prices" in the "tax" tab to "including taxes")
3. Enable coupons and create a coupon
4. Click on the checkout button and confirm the order details do not render
5. Apply the coupon and confirm the order details rendered with coupon details as the coupon is applied
6. Go through the entire checkout process and confirm the order is placed

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1547, opened 2023-10-04
fix: allow order details wrapper to render

miguelpeixe:
### All Submissions:

#1516 introduced a regression by hiding the order details wrapper in the template. Since #1485 this has been [handled dynamically](https://github.com/Automattic/newspack-blocks/blob/d1dfb2ae700dd8ceea001d2c0c71690b405b99f3/includes/class-modal-checkout.php#L516-L535), due to the dynamic nature of WC cart template fragments. Adding or removing coupons changes whether the order details should be rendered or not.

This PR removes the "hidden" class from the template so the fragment hook can determine whether to render details.

1. While on the release branch, checkout a product with the checkout button block and apply a coupon
2. Confirm the coupon doesn't seem applied
3. Check out this branch, try again, and confirm the UI updates and the coupon is applied

### Other information:

### DIFF
--- a/src/modal-checkout/templates/checkout-form.php
+++ b/src/modal-checkout/templates/checkout-form.php
@@ -76,7 +76,7 @@ if ( ! $checkout->is_registration_enabled() && $checkout->is_registration_requir
 	}
 	?>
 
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Works, thanks for catching!

PR #1548, opened 2023-10-09
Inline HPB styles in a style tag

adekbadek:
### All Submissions:

Inlines the critical parts of CSS needed by the Homepage Posts Block, in order to get rid of [layout shift](https://web.dev/cls/) happening when the file-based CSS stylesheet is loaded. 

To demonstrate, here's a recording of a (throttled) page load on `master`:

And here's how it looks like on this branch:

If this strategy proves effective in production environment, it can be enhanced by moving more of the static styles from `view.css` to the dynamically created inline stylesheet. 

1. Create a page with some instances of the Homepage Posts Block, with different type scale settings and images shown
2. Load the front-end, observe the styles are applied just like on `master`, only faster*

\* use DevTools' Network speed throttling to observe it better

### Other information:

### DIFF
--- a/includes/class-newspack-blocks-api.php
+++ b/includes/class-newspack-blocks-api.php
@@ -404,7 +404,12 @@ class Newspack_Blocks_API {
 	 * @return WP_REST_Response.
 	 */
 	public static function css_endpoint() {
-		return newspack_blocks_get_homepage_articles_css_string( [ 'typeScale' => range( 1, 10 ) ] );
+		return newspack_blocks_get_homepage_articles_css_string(
+			[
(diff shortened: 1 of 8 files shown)


miguelpeixe (changes requested):
I was able to reproduce the improvements in CLS but the editor is throwing an error on saves and autosaves:

image

Looks like the inline CSS is being added to the save return payload:

image

adekbadek:
Good catch! Fixed in f97e39a37afe0b44cf740713c11cbe05d64124ef

miguelpeixe (approved):
Thanks for revising! Working as described

PR #1550, opened 2023-10-10
fix: match thankyou template when using existing customer email

dkoo:
### All Submissions:

Adjusts some logic in the WooCommerce checkout flow when using the modal checkout. In the scenario where a donation is made while logged out but using an email address that's already associated with a custom account, we don't want to show the login prompt on the thank you step—instead, we want to show the order summary like we do for new email addresses or when already logged in.

1. On `release`, while not logged in, make a donation using an email address that's already associated with an existing reader account.
2. On the thank you step after the successful transaction, observe that instead of an order summary, there's a message to "Please log in to your account to view this order" and a login form.
3. Check out this branch, repeat with a new donation and the same email address. This time confirm that the thank you page shows an order summary and no message or login form, similarly to what's shown when donating using a new email address.

### Other information:

### DIFF
--- a/src/modal-checkout/templates/form-login.php
+++ b/src/modal-checkout/templates/form-login.php
@@ -1,16 +1,6 @@
 <?php
 /**
- * Login form. WooCommerce by default doesn't allow order details to be shown
- * if the order is completed using a customer email address that doesn't match
- * the currently logged in user.
- *
- * Details: https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/shortcodes/class-wc-shortcode-checkout.php#L302-L321
- *
- * For Newspack sites, we don't want to emphasize user account flows outside of
- * RAS. This custom login template replaces the login form that appears on the
- * order-received.php template with an order details summary so the experience
- * matches whether or not the email address used is already associated with an
- * existing customer account.
+ * Login form.
  *
  * @see https://docs.woocommerce.com/document/template-structure/
  * @package WooCommerce\Templates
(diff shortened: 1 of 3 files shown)


src/modal-checkout/templates/form-login.php
```diff
@@ -0,0 +1,67 @@
...
+ * @version 8.1.0
+ *
+ * @var WC_Order $order
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+$order = isset( $_GET['order_id'] ) ? \wc_get_order( \absint( \wp_unslash( $_GET['order_id'] ) ) ) : false; // phpcs:ignore WordPress.Security.NonceVerification.Recommended, WordPress.WP.GlobalVariablesOverride.Prohibited
```
dkoo, line 26: I'm not sure why this is throwing a PHPCS error, because `$order` is not a global var. Without `WordPress.WP.GlobalVariablesOverride.Prohibited` the following error is thrown:

```
FILE: src/modal-checkout/templates/form-login.php
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 26 | ERROR | Overriding WordPress globals is prohibited. Found
    |       | assignment to $order
    |       | (WordPress.WP.GlobalVariablesOverride.Prohibited)
----------------------------------------------------------------------
```

dkoo: Ah, it was actually a global var because of the template structure. 1adcf62 moves everything to a function so its vars don't conflict with the global scope.

miguelpeixe (approved):
Tests well!

dkoo:
4d20fee adds an extra security layer using Woo's `key` hash which ties the order details summary to the specific transaction flow. This should prevent CSRF requests from being able to expose customer and order details using just an order ID.

miguelpeixe (approved):

PR #1563, opened 2023-10-19
Chore: update phpcs to handle PHP 8

adekbadek:
### All Submissions:

Updates the lint-related Composer packages to handle PHP 8. 

After upgrading to PHP 8, running the linting commands (`./vendor/bin/phpcs`, `./vendor/bin/phpcbf`) was erroring out. This was resolved by updating these packages. Along with the updates, some code had to be refactored – some of it automatically with `phpcbf`, some manually.

1. Smoke test code changes, everything should behave as before
2. Ensure the lint-php job is passing on CI
3. For a good measure, run the linting commands locally to see if there are no issues with local development

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/view.php
+++ b/src/blocks/homepage-articles/view.php
@@ -204,9 +204,9 @@ function newspack_blocks_render_block_homepage_articles( $attributes ) {
 				echo Newspack_Blocks::template_inc( // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
 					__DIR__ . '/templates/articles-list.php',
 					[
-						'articles_rest_url' => $articles_rest_url,
-						'article_query'     => $article_query,
-						'attributes'        => $attributes,
+						'articles_rest_url' => $articles_rest_url, // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+						'article_query'     => $article_query, // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+						'attributes'        => $attributes, // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
 					]
 				);
 				?>
--- a/src/blocks/homepage-articles/view.php
+++ b/src/blocks/homepage-articles/view.php
@@ -308,7 +308,7 @@ function newspack_blocks_format_byline( $author_info ) {
 		array_reduce(
 			$author_info,
 			function ( $accumulator, $author ) use ( $author_info, &$index ) {
-				$index ++;
+				$index++;
 				$penultimate = count( $author_info ) - 2;
 
 				$get_author_posts_url = get_author_posts_url( $author->ID );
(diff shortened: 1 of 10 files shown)


src/blocks/homepage-articles/view.php
```diff
@@ -308,7 +308,7 @@ function newspack_blocks_format_byline( $author_info ) {
 		array_reduce(
 			$author_info,
 			function ( $accumulator, $author ) use ( $author_info, &$index ) {
-				$index ++;
+				$index++;
```
miguelpeixe, line 311: Why use pre-increment in the other cases and post-increment in this case?

dkoo: Agreed, and I'm not familiar with the difference between `++$i` vs. `$i++`

dkoo:
Tests are all running well and passing for me locally in a PHP 8.0 environment. Has anyone been able to test on PHP 8.1 yet?

adekbadek:
I'm running 8.2 locally, tests are passing.

src/blocks/homepage-articles/view.php
adekbadek: The difference (didn't know that before either!) is that post-de/increment _first returns the value, then increments it_ ([source](https://www.php.net/manual/en/language.operators.increment.php)). [According to the creators of PHPCSStandards](https://github.com/PHPCSStandards/PHPCSExtra/tree/746c3190ba8eb2f212087c947ba75f4f5b9a58d5#universaloperatorsdisallowstandalonepostincrementdecrement-wrench-bar_chart-books):

> Using pre-in/decrement is more in line with the principle of least astonishment and prevents bugs when code gets moved around at a later point in time.

This is [only applied](https://github.com/PHPCSStandards/PHPCSExtra/blob/746c3190ba/Universal/Sniffs/Operators/DisallowStandalonePostIncrementDecrementSniff.php#L94-L97) however if it is a **"stand-alone" statement**.

A statement is "stand-alone" if it does not require other statements to be executed ([source](https://brainly.com/question/37422224)). Since this statement is inside of a statement (that's how I understand (the condition)[https://github.com/PHPCSStandards/PHPCSExtra/blob/746c3190ba/Universal/Sniffs/Operators/DisallowStandalonePostIncrementDecrementSniff.php#L94-L97] on which it is allowed), it's not stand-alone, and so post-increment is allowed. Similarly on line 173, incrementing `$page` is in a statement, and _could_ be post-incremented. 

All that said, this is not intuitive to me. We can switch off this rule if it creates more confusion than clarity. WDYT?

dkoo: Thank you for looking into this and explaining! I think we should switch off the rule and continue using post-increment as we have been, since it makes the most sense for us and probably also for other engineers who write code in languages other than PHP.

adekbadek: Done in b6b144d744e98920675715dd275812c42f6b289f!

dkoo (approved):
Sorry for the delay on this. It took me a while to find the time to upgrade my local environment to PHP 8.1. No major issues running the new dependencies/rules.

adekbadek:
Added exclusion of `Generic.Formatting.MultipleStatementAlignment.NotSameWarning` and `WordPress.PHP.YodaConditions.NotYoda` rules, as discussed on Slack (p1699431739808529-slack-C04UKFHPB6U).

PR #1566, opened 2023-10-20
Fix Homepage Posts Block to support primary category theme mod setting

ronchambers:
Fixed Homepage Posts Block to adhere to the existing Customizer Theme Mod boolean 'post_primary_category'. This customizer boolean is used by the Newspack Theme to show either the WPSEO/Yoast primary category or all categories in archive and post templates.

Now that this same boolean is checked in the Homepage Posts Block, the block output will better match the theme category output.

Newspack Theme > Customizer > Template Settings > Post Settings:

image

Homepage Posts Block will now check this setting.

This pull request essentially copies 2 lines of code from Newspack Theme (condensed into 1): [get_theme_mod](https://github.com/Automattic/newspack-theme/blob/89df5c7d446698b314ba9493e64d1eab7af045fa/newspack-theme/inc/template-tags.php#L233) and [if check](https://github.com/Automattic/newspack-theme/blob/89df5c7d446698b314ba9493e64d1eab7af045fa/newspack-theme/inc/template-tags.php#L236)

### DIFF
--- a/includes/class-newspack-blocks-api.php
+++ b/includes/class-newspack-blocks-api.php
@@ -140,7 +140,7 @@ class Newspack_Blocks_API {
 		$category = false;
 
 		// Use Yoast primary category if set.
-		if ( class_exists( 'WPSEO_Primary_Term' ) ) {
+		if ( class_exists( 'WPSEO_Primary_Term' ) && get_theme_mod( 'post_primary_category', true ) ) {
 			$primary_term = new WPSEO_Primary_Term( 'category', $object['id'] );
 			$category_id  = $primary_term->get_primary_term();
 			if ( $category_id ) {
--- a/src/blocks/homepage-articles/view.php
+++ b/src/blocks/homepage-articles/view.php
@@ -345,7 +345,7 @@ function newspack_blocks_format_byline( $author_info ) {
 function newspack_blocks_format_categories( $post_id ) {
 	$category = false;
 	// Use Yoast primary category if set.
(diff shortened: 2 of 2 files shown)


adekbadek (commented):
I'm not sure about this change. The theme mod is for displaying Yoast's primary category on in **single post view**, extending this to the HP block seems like a stretch. IMO it makes more sense to make it a block attribute.  

I'll defer to @laurelfulford with the decision, though.

laurelfulford:
Thanks for the ping on this, @adekbadek, and for submitting this change, @ronchambers! 

I agree that it should be an option in the block -- on top of what @adekbadek said, if we use an existing control it means this change would affect all sites using that control without their say. A block-based control will also work with other themes.

I did a little testing; right now, it still only shows one category because [that's the non-Yoast fallback behaviour in the block](https://github.com/Automattic/newspack-blocks/pull/1566/files#diff-08ca22a99213615eda27f7d1733026453974879ed9799ae54b3e0a9833e84f10R358). When Yoast is disabled, only the first category in the array (literally the first category checked when building the post) is shown instead.

When Yoast is active, it looks like the Primary Category is made the first one in the array (TIL!). It kind of makes sense (the primary category should be first), but it means the blocks still show the same single category. Instead of using the existing fallback, I think we'll need a third if/else option to loop through all of them.

Just let me know if you have any questions at all about the above!

adekbadek:
Closing for now.

PR #1561, opened 2023-10-18
Modal checkout – post-checkout newsletter signup

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Adds post-checkout newsletter signup options:

image

image

### How to test the changes in this Pull Request:

1. Switch `newspack-plugin` to branch from https://github.com/Automattic/newspack-plugin/pull/2707
1. Switch `newspack-theme` to branch from https://github.com/Automattic/newspack-theme/pull/2188 
1. Ensure that the Newsletters provider is configured and some lists are active in Engagement wizard -> Newsletters
1. With "Newspack" set as the Reader Revenue platform, make a donation using the checkout modal
3. Observe that after the checkout is successful, the chosen newsletter lists are presented
4. Select some of them and click "Continue", observe a success message
5. Verify that the email used has been added to the list(s)
6. Set a `NEWSPACK_DISABLE_POST_CHECKOUT_NEWSLETTER_SIGNUP` environment variable to `true` and observe that now the signup lists are not displayed after the checkout 

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1571, opened 2023-10-24
Changes to button-after-success feature 

adekbadek:
### All Submissions:

1. Implements the post-checkout button (https://github.com/Automattic/newspack-blocks/pull/1521) for the Donate block. 
2. Changes how the post-checkout button works:
    1. It will be displayed at all times, instead of optionally
    2. By default, the post-checkout button will say "Continue browsing" and close the checkout modal on clicking
    3. Two other options are, as before, to go to previous page (the `referrer`) or go to custom URL

1. Switch `newspack-plugin` to `feat/donate-post-checkout-button` (https://github.com/Automattic/newspack-plugin/pull/2719)
4. Add a Donate block to a page, configure it to display a button in the "After Purchase" sidebar settings panel
5. Complete a transaction, observe the button is displayed (as in https://github.com/Automattic/newspack-blocks/pull/1521/)
6. Test the post-checkout button flow in various configurations:
    1. With and without the newsletters (`NEWSPACK_ENABLE_POST_CHECKOUT_NEWSLETTER_SIGNUP`, see https://github.com/Automattic/newspack-blocks/pull/1561)
    2. Initiated from a Checkout Button block
    3. Initiated from a Checkout Button block, with a variable product

### Other information:

### DIFF
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -729,18 +729,22 @@ final class Modal_Checkout {
 		}
 		$signup_data = self::get_newsletter_signup_data();
 		if ( false !== $signup_data ) {
-			$result = \Newspack_Newsletters_Subscription::add_contact(
-				[
-					'email'    => $signup_data['email'],
-					'metadata' => [
-						'current_page_url'                => home_url( add_query_arg( array(), \wp_get_referer() ) ),
-						'newsletters_subscription_method' => 'post-checkout',
+			if ( empty( $signup_data['lists'] ) ) {
+				return new \WP_Error( 'no-lists', __( 'No lists selected.', 'newspack-blocks' ) );
+			} else {
+				$result = \Newspack_Newsletters_Subscription::add_contact(
+					[
+						'email'    => $signup_data['email'],
+						'metadata' => [
+							'current_page_url' => home_url( add_query_arg( array(), \wp_get_referer() ) ),
+							'newsletters_subscription_method' => 'post-checkout',
+						],
 					],
-				],
-				$signup_data['lists']
-			);
-			if ( \is_wp_error( $result ) ) {
-				return $result;
+					$signup_data['lists']
+				);
+				if ( \is_wp_error( $result ) ) {
+					return $result;
+				}
 			}
 			return true;
 		}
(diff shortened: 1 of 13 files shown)


includes/class-modal-checkout.php
```diff
@@ -562,28 +585,43 @@ public static function order_review_fragments( $fragments ) {
...
-			empty( $_REQUEST['after_success_behavior'] ) ||
-			empty( $_REQUEST['after_success_url'] )
-		) {
+		if ( empty( $_REQUEST['modal_checkout'] ) ) {
 			return;
 		}
 
-		$button_label = ! empty( $_REQUEST['after_success_button_label'] ) ? sanitize_text_field( wp_unslash( $_REQUEST['after_success_button_label'] ) ) : __( 'Continue browsing', 'newspack-blocks' );
+		$button_label = ! empty( $_REQUEST['after_success_button_label'] ) ? sanitize_text_field( urldecode( wp_unslash( $_REQUEST['after_success_button_label'] ) ) ) : __( 'Continue browsing', 'newspack-blocks' ); // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+		$url          = ! empty( $_REQUEST['after_success_url'] ) ? sanitize_text_field( urldecode( wp_unslash( $_REQUEST['after_success_url'] ) ) ) : ''; // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
```
dkoo, line 614: Nit: should this be `esc_url`?

```suggestion
		$url          = ! empty( $_REQUEST['after_success_url'] ) ? esc_url( urldecode( wp_unslash( $_REQUEST['after_success_url'] ) ) ) : ''; // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
```

dkoo (changes requested):
Almost there, but I ran into an instance where the after-checkout modal is blank after clicking "Continue":

Screenshot 2023-10-26 at 4 41 35 AM

To replicate, create a Checkout Button block pointing to a simple non-subscription product, complete the purchase, and then deselect all newsletter options before clicking "Continue". The final screen will be blank as above if not opting into any newsletter option.

includes/class-modal-checkout.php
adekbadek: Actually escaping is done in the output code a few lines below, so that's redundant. Changed in a19bd5a265c41e1aa17e801e0bace8e228281a86

adekbadek:
> The final screen will be blank as above if not opting into any newsletter option.

Thanks for catching that! Ideally the newsletter signup form should be validated (`required` field), but there's no pure HTML solution for this, so this will need a bit of JS. In the interest of time I'll tackle that in a separate PR.

dkoo:
> Ideally the newsletter signup form should be validated (required field)

I don't agree, if I'm understanding you correctly. Readers should be able to continue without selecting any newsletter options, right?

dkoo (changes requested):
If no lists are selected, we now show a simple error message:

Screenshot 2023-10-27 at 5 10 22 AM

Instead, I think we should show the usual success message. We cannot require readers to opt into newsletters in this scenario, and must provide a way for them to proceed without opt-in.

adekbadek:
Done in d94d8ebd33db9bbdc2ae578207ffe39fe904a5d4. If no lists were selected "Signup successfull!" heading will be omitted.

dkoo (approved):
This works for me, for now! Thanks for the latest updates.

PR #1581, opened 2023-11-03
feat(modal-checkout): support shipping

dkoo:
### All Submissions:

Adds support for shipping info during the modal checkout flow, to bring our modal checkout to closer parity with the default Woo checkout flow.

1. Check out this branch.
2. In WP admin > WooCommerce > Settings > Shipping, add at least one shipping zone and one shipping method to the zone (preferably one that costs some amount of money).
3. In Reader Revenue > Donations enable all of the billing fields.
4. Publish at least one product that is NOT virtual. Ensure that it has some available inventory.
5. Create a Checkout Button block pointing to your non-virtual product.
6. As a new reader, purchase the product using the Checkout Button. Confirm that the modal now shows an order summary that shows the total cost of the purchase including shipping fees, and a toggle to ship to a different address.

Screenshot 2023-11-03 at 9 27 39 AM

7. Complete the purchase without touching the toggle. In WP admin > Users, find the customer's WP user account and confirm that the shipping address fields are automatically filled to match the billing address fields.
8. In another new session, purchase the same product again. This time toggle ON the "ship to a different address?" option and fill in a different address in the fields that are unhidden. Make sure you fill in the optional order notes.
9. On the payment modal screen, confirm that the separate shipping address is shown underneath the billing address, and that you can go back to edit them (unfortunately the "ship to a different address?" option doesn't seem to be saved if you go back to edit, but the address fields themselves are, so this doesn't seem like a blocker to me).
10. Also go to the order for this last purchase and confirm that the "Customer provided note" field shows the text you entered into the order notes field during checkout.
11. Complete the transaction. In WP admin > Users, find the customer's WP user account and confirm that the shipping address fields are filled in and match the address you inputted in the second purchase.
12. Check out https://github.com/Automattic/newspack-plugin/pull/2733.
13. Navigate to or log into My Account. Confirm that there's an "Addresses" menu item and that clicking on it allows you to see/edit the billing and shipping address info you've entered previously.
14. Smoke test a regular donation and a purchase of a virtual product. Both should work as before with no changes. 
15. In Newspack > Reader Revenue disable all billing fields except for First/Last Name and email. Complete a new donation and purchase of a virtual product and confirm that My Account doesn't show the Addresses menu item (because your account has no address info).

### Other information:

### DIFF
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -601,12 +601,17 @@ final class Modal_Checkout {
 		if ( ! self::is_modal_checkout() ) {
 			return $fields;
 		}
+
 		/**
 		 * Temporarily use the same fields as the donation checkout.
 		 *
 		 * This should soon be replaced with a logic that allows the customization
 		 * at the Checkout Button Block level.
 		 */
+		$cart = \WC()->cart;
+		if ( $cart->needs_shipping_address() ) { // Don't modify fields if shipping is required.
+			return $fields;
+		}
 		$billing_fields = apply_filters( 'newspack_blocks_donate_billing_fields_keys', [] );
 		if ( empty( $billing_fields ) ) {
 			return $fields;
@@ -642,19 +647,15 @@ final class Modal_Checkout {
 	 */
 	public static function get_prefilled_fields() {
 		$checkout        = \WC()->checkout();
-		$cart            = \WC()->cart;
-		$checkout_fields = [ 'billing' => $checkout->get_checkout_fields( 'billing' ) ];
+		$checkout_fields = $checkout->get_checkout_fields();
 		$customer        = new \WC_Customer( get_current_user_id() );
 		$customer_fields = $customer->get_data();
-
-		if ( $cart->needs_shipping_address() ) {
-			$checkout_fields['shipping'] = $checkout->get_checkout_fields( 'shipping' );
-		}
 
 		// If the user is logged in and there's no billing email, use the user's email.
 		if ( is_user_logged_in() && empty( $customer_fields['email'] ) ) {
 			$customer_fields['email'] = $customer->get_email();
 		}
+
 		$valid_request    = self::validate_edit_billing_request();
 		$prefilled_fields = [];
 		foreach ( $checkout_fields as $type => $fields ) {
@@ -750,7 +751,10 @@ final class Modal_Checkout {
 			return true;
 		}
 		if ( $cart->needs_shipping_address() ) {
-			return true;
+			$totals = $cart->get_totals();
+			if ( (float) $totals['total'] !== (float) $totals['subtotal'] ) {
+				return true;
+			}
 		}
 		return false;
 	}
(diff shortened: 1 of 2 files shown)


leogermani (changes requested):
I'm getting a "First name is a required field" error when trying to make the purchase. The checkout works fine for donation products...

Captura de tela de 2023-11-07 12-03-07

I also see this table summary at the top. Not sure if that's expected or if this might be related to some other option in my site (I remember seeing this when playing with taxes settings)

Captura de tela de 2023-11-07 12-07-15

dkoo:
> I'm getting a "First name is a required field" error when trying to make the purchase. The checkout works fine for donation products...

@leogermani what billing fields do you have enabled for your site in Reader Revenue > Donations?

leogermani:
> > I'm getting a "First name is a required field" error when trying to make the purchase. The checkout works fine for donation products...
> 
> @leogermani what billing fields do you have enabled for your site in Reader Revenue > Donations?

All of them

dkoo:
> I'm getting a "First name is a required field" error when trying to make the purchase. The checkout works fine for donation products...

I'm not able to replicate this—does it keep happening if you try in a fresh session? If so, could you provide steps to replicate?

> I also see this table summary at the top. Not sure if that's expected or if this might be related to some other option in my site (I remember seeing this when playing with taxes settings)

Yes, this is intentional because shipping usually involves some kind of shipping fee (configurable for each shipping method/zone). The summary will show the subtotal + any applicable shipping fees. I think it's worth showing the summary even if the site offers free shipping, so that the customer knows exactly what the order total will be.

Screenshot 2023-11-20 at 4 07 58 PM

dkoo:
Note: rebased to reflect the new project-level scope.

dkoo:
@leogermani dda1ef0 fixes the "required fields" validation to only require shipping fields if the "ship to a different address" option is enabled.

includes/class-modal-checkout.php
```diff
@@ -609,26 +633,54 @@ public static function get_prefilled_fields() {
 	 */
 	public static function has_filled_required_fields() {
 		$checkout        = \WC()->checkout();
-		$fields          = $checkout->get_checkout_fields( 'billing' );
-		$required        = array_filter(
-			$fields,
-			function( $field ) {
-				return isset( $field['required'] ) && $field['required'];
+		$cart            = \WC()->cart;
+		$checkout_fields = [ 'billing' => $checkout->get_checkout_fields( $type ) ];
```
leogermani, line 637: `$type` is undefined here

leogermani (changes requested):
When I click the checkout button, I'm now always taken to the second step of the checkout modal, as if I had already entered my billing and shipping info (like when I'm logged in), but it's all empty. I can click Edit and then start from scratch. Then it works.

I think this is why I'm seeing this warning:

```
foreach() argument must be of type array|object, bool given in /newspack-repos/newspack-blocks/src/modal-checkout/templates/checkout-form.php on line 140
```

There's also a Warning of `$type` being undefined. I left a comment in the code.


> Confirm that there's an "Addresses" menu item

There is not. And there's a "Billing" item with a button that takes me to Stripe. Doesn't look right.

image

includes/class-modal-checkout.php
dkoo: Fixed in b4e7f28, but I'm thinking we should hold on this until Miguel finishes the refactor to a single template for the checkout modal since that might cause a lot of this changeset to be revisited.

dkoo:
Closing this in favor of #1626.

PR #1587, opened 2023-11-08
fix: flush WC template cache on template updates

adekbadek:
### All Submissions:

Ensures the WC template cache is cleared (once) when the templates are updated.

⚠️ This change should be merged into the current alpha, too. 

The issue with cached templates was only encountered on live sites, we could not reproduce it locally, on JN, or even staging. 

### Other information:

### DIFF
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -464,6 +464,14 @@ final class Modal_Checkout {
 			return $located;
 		}
 
+		// Flush WC template cache when the templates version changes.
+		// Increase this number after making changes to the templates.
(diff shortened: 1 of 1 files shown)


leogermani:
I suggest we close this one in favor of https://github.com/Automattic/newspack-theme/pull/2201

The issue has nothing to do with this plugin. The modal checkout templates are not affected by the cache because they are inserted via a filter that happens after the cache is read. No need to ever clear the cache here.

dkoo:
@leogermani thanks for digging into this and explaining what's going on. Closing this PR as suggested.

PR #1591, opened 2023-11-09
feat(modal-checkout): update "place order" text to "donate now"

adekbadek:
### All Submissions:

Updates the "Place order" text in modal checkout to "Donate now".

⚠️  This is an alpha-hotfix, merging to `alpha`.

1. Test with a one-time and recurring donation via modal checkout – the CTA on the second step should always say "Donate now"
2. Use the regular checkout, confirm the corresponding button says "Place order" for regular items and "Sign up now" for subscription items

### Other information:

### DIFF
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -48,6 +48,9 @@ final class Modal_Checkout {
 		add_filter( 'newspack_donations_cart_item_data', [ __CLASS__, 'amend_cart_item_data' ] );
 		add_filter( 'newspack_recaptcha_verify_captcha', [ __CLASS__, 'recaptcha_verify_captcha' ], 10, 2 );
 		add_filter( 'woocommerce_enqueue_styles', [ __CLASS__, 'dequeue_woocommerce_styles' ] );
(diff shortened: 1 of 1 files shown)


includes/class-modal-checkout.php
```diff
@@ -898,7 +901,23 @@ public static function recaptcha_verify_captcha( $should_verify, $url ) {
...
+	/**
+	 * Set the "Place order" button text.
+	 *
+	 * @param string $text The button text.
+	 */
+	public static function order_button_text( $text ) {
+		if ( ! self::is_modal_checkout() ) {
+			return $text;
+		}
+		return __( 'Donate now', 'newspack-blocks' );
```
miguelpeixe, line 920: The modal checkout is also used for other products, especially membership subscriptions. This change should only impact donation carts.

miguelpeixe (changes requested):

includes/class-modal-checkout.php
dkoo: Jumping in this is an urgent request. 00daac9 adds the condition so that the "Donate Now" text is only displayed for donation transactions.

adekbadek: Thanks Derrick!

PR #1601, opened 2023-11-22
fix(modal-checkout): prevent initial render of details table

adekbadek:
### All Submissions:

Prevents initial render of the order details table.  

See 1200133283036252-as-1205933702651194

1. On `master`, make a donation using the modal checkout – observe that the order details table is visible for a moment right after the modal content loads
2. Switch to this branch, observe the issue solved 
3. [Enable coupons](https://woo.com/document/coupon-management/) and create a coupon. Apply it during the checkout and observe the order details table appears only after the coupon is applied

### Other information:


includes/class-modal-checkout.php
```diff
@@ -489,6 +489,12 @@ public static function wc_get_template( $located, $template_name ) {
 			}
 		}
 
+		// This is for the initial display – the markup will be refetched on cart updates (e.g. applying a coupon).
+		// Then it'd be handled by the `woocommerce_update_order_review_fragments` filter.
+		if ( 'checkout/review-order.php' === $template_name && ! self::should_show_order_details() ) {
+			$located = NEWSPACK_BLOCKS__PLUGIN_DIR . 'src/modal-checkout/templates/empty-order-details.php';
```
miguelpeixe, line 495:
```suggestion
			ob_start();
			echo '<table class="shop_table woocommerce-checkout-review-order-table empty"></table>';
			return ob_end_flush();
```

Non-blocking: we can remove the need for a separate file for the empty table template by using an output buffer.

miguelpeixe (approved):
Working as described, but I propose a slightly different approach below.

adekbadek:
Thanks @miguelpeixe !

PR #1603, opened 2023-11-22
feat: support subscriptions gifting

dkoo:
### All Submissions:

Adds support for the [WooCommerce Subscriptions Gifting extension](https://github.com/woocommerce/woocommerce-subscriptions-gifting) to our modal checkout flow.

Because of limitations in how the WCSG extension handles its own form fields and validation, this PR implements custom form fields for inputting gift recipient info during checkout and validating that input via WCSG's own validation methods.

1. Install and activate the [WooCommerce Subscriptions Gifting extension](https://github.com/woocommerce/woocommerce-subscriptions-gifting)
2. Check out this branch and start a new subscription transaction (donation or non-donation via Checkout Button).
3. Confirm that the gift subscription options are available in the first modal screen:

Screenshot 2024-01-04 at 12 39 06 PM

4. Confirm that checking the "This donation/transaction is a gift" box toggles the recipient email field.
5. UNCHECK the box and confirm that you can complete the transaction without entering anything into the email field.
6. Start a new transaction and CHECK the box. Confirm that you can't proceed if you don't enter an email address or if you enter a non-email address string into the gift recipient field.

Screenshot 2024-01-04 at 12 39 22 PM

7. Confirm that you can't proceed if you enter the same email address as the billing email address into the gift recipient field.

Screenshot 2024-01-04 at 12 39 31 PM

8. Enter a valid email address into the gift recipient field and confirm that you see it in the billing details summary on the next modal screen:

Screenshot 2024-01-04 at 12 42 42 PM

9. Complete the transaction and confirm that the subscription is listed with both buyer and recipient info in WP admin, and that the recipient receives an email address informing them of their new gift subscription.

Screenshot 2024-01-04 at 12 54 40 PM
Screenshot 2024-01-04 at 12 54 15 PM

10. Smoke test a transaction through the regular Shop/Cart/Checkout pages and confirm that the gift options work there too (they should be untouched from the base Subscriptions Gifting plugin experience).

### Other information:


includes/class-modal-checkout.php
```diff
@@ -466,10 +469,28 @@ public static function woocommerce_get_return_url( $url, $order ) {
...
+		// WooCommerce Subscriptions Gifting extension.
+		if (
+			'html-add-recipient.php' === $template_name &&
+			(
+				// Don't show on cart.
+				\is_cart() ||
+				(
+					// Only show on checkout if not modal, or if newspack_display_gifting_information template arg is passed.
+					\is_checkout() &&
+					self::is_modal_checkout() &&
```
adekbadek, line 486: Nit: the whole if statement could be moved below the `if ( ! self::is_modal_checkout() )` statement below, so this condition would not be necessary, and it would be clear that any modifications to the templates are only done in the context of the modal checkout.

includes/class-modal-checkout.php
```diff
@@ -466,10 +469,28 @@ public static function woocommerce_get_return_url( $url, $order ) {
...
+					empty( $args['newspack_display_gifting_information'] )
+				)
+			)
+		) {
+			$located = NEWSPACK_BLOCKS__PLUGIN_DIR . 'src/modal-checkout/templates/html-add-recipient.php';
```
adekbadek, line 491: As @miguelpeixe pointed out in  https://github.com/Automattic/newspack-blocks/pull/1601#discussion_r1402090229, you don't need the template file – (empty) markup can be returned from  this function.

adekbadek (changes requested):
On the frontend side it worked as expected, but the resulting subscription is not marked as being gifted in the admin:

image

includes/class-modal-checkout.php
adekbadek: Oops, scratch that. Just noticed that this solution causes a PHP Notice to be logged: 

```
PHP Notice:  Function wc_get_template was called incorrectly.
```

Nevermind, and I'll redo my change.

dkoo:
> On the frontend side it worked as expected, but the resulting subscription is not marked as being gifted in the admin:

Hmm, after buying a lot of gift subscriptions, I'm seeing inconsistent results. I think there's some kind of race condition happening here. This PR may be better to hold and revisit once we have the single-screen modal checkout template.

includes/class-modal-checkout.php
dkoo: 30b20bc prefixes the name of the empty template with `empty-` to make its purpose a bit clearer.

dkoo:
@adekbadek @miguelpeixe this has been fixed and the testing instructions updated. Please test and review again!

adekbadek (approved):
Works as described!

PR #1611, opened 2023-11-27
fix(homepage-posts): placement for inline styles

miguelpeixe:
### All Submissions:

The inline styles for the Homepage Posts block are not rendered right after `.entry-content` as documented in the code. They are rendered alongside the first rendered HPP block. This can cause containers that leverage pseudo classes like `:first-child` or `:last-child` to not behave as expected, since the style tag can get picked up.

| Before | After |
| --- | --- |
| image | image |

This PR changes the placement for the inline CSS so it's printed inside the block container. The improved CLS implemented in the original PR (#1548) should be preserved.

1. Confirm the bug by adding a 2 column block with a HPP block on each column and visiting the front-end (see image above)
2. Check out this branch and confirm there's no longer a top margin on the first block
3. Open DevTools, change to mobile viewport and apply network throttling to slow 3g (see #1548 for further details if needed)
4. Refresh the page and confirm the slowly rendered layout doesn't shift

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/view.php
+++ b/src/blocks/homepage-articles/view.php
@@ -229,7 +229,7 @@ function newspack_blocks_render_block_homepage_articles( $attributes ) {
 	$block_name = apply_filters( 'newspack_blocks_block_name', 'newspack-blocks/homepage-articles' );
 
 	// Gather all Homepage Articles blocks on the page and output only the needed CSS.
-	// This CSS will be printed right after .entry-content.
+	// This CSS will be printed along with the first found block markup.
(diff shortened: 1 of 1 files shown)


laurelfulford (approved):
Thanks @miguelpeixe!! 🙌 

This fixes the issue for me! I also checked through our theme and regular block styles, it looks like we don't rely on this div's position in any of our styles so adding another tag before shouldn't cause any problems unless there's a super-weird edge case out there!

miguelpeixe:
Thanks, Laurel!

PR #1615, opened 2023-11-29
feat(modal-checkout): allow anonymous purchase for registered email

miguelpeixe:
### All Submissions:

Moves a logic that is [currently placed behind RAS](https://github.com/Automattic/newspack-plugin/blob/fcad059abbe0314ba8a6f3eaaf92d5c96350e248/includes/reader-activation/class-reader-activation.php#L89-L90) but is designed for the modal checkout.

If an anonymous purchase is made for a registered email, it should be attached to that account regardless of them being authenticated.

1. Checkout this branch and https://github.com/Automattic/newspack-plugin/pull/2781
2. Make sure RAS is toggled off (`$ wp option delete newspack_reader_activation_enabled`)
3. Disable the "Allow customers to create an account during checkout" option from the "Accounts & Privacy" section of WC:
image

4. Go through the regular checkout flow anonymously and use an existing email
5. Confirm the order goes through and in the dashboard the order is not assigned to any user (shown as "guest" customer)
6. Also anonymously, go through the modal checkout flow (either through the Checkout Button block or Donate block) using an existing email
7. Confirm the order has been placed and assigned to the user
8. Visit `/my-account` and confirm you're still logged out

### Other information:

### DIFF
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -51,6 +51,10 @@ final class Modal_Checkout {
 		add_filter( 'wcs_place_subscription_order_text', [ __CLASS__, 'order_button_text' ], 1 );
 		add_filter( 'woocommerce_order_button_text', [ __CLASS__, 'order_button_text' ] );
 		add_filter( 'option_woocommerce_subscriptions_order_button_text', [ __CLASS__, 'order_button_text' ] );
+
+		/** Custom handling for registered users. */
+		add_filter( 'woocommerce_checkout_customer_id', [ __CLASS__, 'associate_existing_user' ] );
+		add_filter( 'woocommerce_checkout_posted_data', [ __CLASS__, 'skip_account_creation' ], 11 );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


adekbadek (commented):
> Confirm you're prompted to log in, as it's available through this checkout UI

Can't get this part to work, tried with different configurations of the guest checkout options. Tested by buying a subscription in a an anonymous session, using an existing user's email. What is the correct configuration on WC guest checkout for this to work? And at which point should the login prompt appear?

image

includes/class-modal-checkout.php
```diff
@@ -973,6 +977,50 @@ public static function order_button_text( $text ) {
 		return $text;
 	}
 
+	/**
+	 * If a reader tries to make a recurring donation with an email address that
```
leogermani, line 981:
```suggestion
	 * If a reader tries to make a purchase with an email address that
```

leogermani (commented):
It works, I left one minor fix in the docs

But I'm still struggling to understand why this should be a modal checkout thing and not a RAS.

On one hand, I understand that the modal checkout should be Woocommerce friendly and compatible with its features

On the other hand, I thought anything that would automatically register a user in WP should only happen under RAS (or if explicitly enabled in woocommerce in the cases it does it)

miguelpeixe:
> But I'm still struggling to understand why this should be a modal checkout thing and not a RAS.

The modal checkout doesn't currently support creating an account mid-checkout and we have publishers outside of RAS using it. This creates a scenario in which existing, but logged-out, customers get stuck while making a purchase.

Context: p1701216656275889-slack-C0390J7K0KY

>  I thought anything that would automatically register a user in WP should only happen under RAS

That is still the case. These hooks are not to create an account. It attaches the order to the existing email if it exists.

leogermani (approved):
thanks for the context

PR #1617, opened 2023-12-01
Donate Block: Update selected tier behaviour

laurelfulford:
### All Submissions:

Queuing this up for release early next week! 

This PR updates the logic used to toggle the tab status (active or inactive) in the tiers view of the Donate block.

Originally the tabs, when one was selected, would all toggle the active CSS class -- meaning either one or two would appear active, and they wouldn't necessarily line up with the actual active tab. This was only on the front-end -- it worked as expected in the editor. This PR updates it to check for current status and update; if there's a more elegant way to do this, I'm all ears! 

See 1200550061930446-as-1206029652902709

1. Add a Donate block to your test set, and change the Layout to Tiers; make sure the block is showing all three tabs (One time, Monthly, Annually). Publish you page.
2. View on front end, and click between the different tabs; note that the selected state appears to alternate between One Time, and Monthly+Annually, regardless which you click. You can tell which one is truly active by the "per X" or "once" text after each dollar amount:

before

5. Repeat step two; confirm that only the clicked tab appears selected.

after

### Other information:

### DIFF
--- a/src/blocks/donate/tiers-based/view.ts
+++ b/src/blocks/donate/tiers-based/view.ts
@@ -52,9 +52,11 @@ export default ( parentEl: HTMLElement ) => {
 
 			// Toggle the frequency buttons' active states.
 			frequencyButtonsEls.forEach( el => {
-				selectedFrequency === el.getAttribute( 'data-frequency-slug')
-					? el.classList.add( BUTTON_ACTIVE_CLASSNAME )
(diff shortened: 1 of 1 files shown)


dkoo (approved):
This fixes the issue for me!

laurelfulford:
Awesome, thanks @dkoo! Setting myself a reminder to merge Monday AM!

PR #1621, opened 2023-12-05
fix(homepage-posts): handle empty block when printing inline styles

adekbadek:
### All Submissions:

If the HP block is empty, and it happens to be the first one on the page, the inline styles won't be printed. 

1. Set up a page with at least two HP blocks – the first one should be empty*
2. For the second block, choose a large type scale
3. On `release`, observe that the inline styles are not applied – the type scale is not reflected on the frontend 
4. Switch to this branch, observe the styles are applied

\* e.g. set it to display posts from a category with no posts

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/view.php
+++ b/src/blocks/homepage-articles/view.php
@@ -352,13 +352,14 @@ function newspack_blocks_render_block_homepage_articles( $attributes ) {
 
 	ob_start();
 
+	echo $inline_style_html; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+
 	if ( $article_query->have_posts() ) :
 		?>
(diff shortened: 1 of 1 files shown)


miguelpeixe (changes requested):
It was moved to inside the container in #1611 so it doesn't interfere with `:first-child` CSS selectors. This change will reintroduce the issue.

How about using a class variable to check if the inline CSS was inserted, and if not, the next HPP block can print it?

adekbadek:
Got it, thanks for catching that! 0820a010bf04d49519c7d295d7bc5eb3b31ccba8 resolves the issue in a better way.

miguelpeixe (approved):
Good solution!

PR #1626, opened 2023-12-09
feat(ras-acc): support shipping, custom fields, and order notes in modal checkout

dkoo:
### All Submissions:

Adds support for conditionally showing the order summary for a shippable product only if shipping might incur additional fees on top of the of the order subtotal.

Also adds support for the new features introduced in https://github.com/Automattic/newspack-plugin/pull/2800 to show different billing fields for donations vs. non-donation transactions, and to show order notes in either case.

Check out this branch and https://github.com/Automattic/newspack-plugin/pull/2800.

#### Conditional order summary

1. In **WooCommerce > Settings > Shipping**, make sure you have an active shipping zone and that the shipping zone has two shipping methods: one should be "Free shipping" with no fees, and the other should be "Flat rate" with a shipping fee attached. Enable just "Free shipping" for now.
2. Make sure you have a shippable product published. Assign the product to a Checkout Button block.
3. As a new reader, purchase the shippable product via the Checkout Button block and proceed to the second modal screen (payment details). Confirm that no order summary is shown, because the "Free shipping" method is active.
4. Back in settings, turn on the fee-based shipping method.
5. Repeat step 3, but this time confirm the second modal screen contains an order summary showing an option to select a shipping method, and that updating the shipping method to the one that incurs will also update the order total:

Screenshot 2023-12-08 at 6 34 27 PM

6. Back in settings, turn off "Free shipping" and repeat once more. This time confirm that the order summary is shown and that the shipping fee is automatically appended to the total (as it's now the only available shipping method).

#### Custom billing fields and order notes

See testing instructions in https://github.com/Automattic/newspack-plugin/pull/2800.

### Other information:


dkoo:
I'm not sure why the SCSS lint is failing...those lines are unchanged and also have `/* stylelint-disable-line */`. I would ignore this warning for now.

leogermani:
It didn't quite work for me.

I set up the testing slightly different.

I have two shipping zones, each of them with only one method available
* US -> flat rate
* Brazil -> free shipping

When I set the shipping address to Brazil, I still see the summary with "Frete grátis" (free shipping" there.

image

When I set it to the US, I also see the summary, as expected.

dkoo:
Ah, that's a good catch. a4ca300 fixes the logic so that it only considers shipping methods that apply to the current cart transaction.

leogermani (approved):

PR #1629, opened 2023-12-13
feat(modal-checkout): custom coupon form

miguelpeixe:
### All Submissions:

Requires #1602.

Implements a custom coupon form handling so it doesn't inherit Woo's default UI handling of the form. The custom handling is implemented by changing the class name of the form.

image

image

1. Check out this branch and make sure you have coupons enabled and at least 1 enabled coupon code
2. Go through the modal checkout flow through a Checkout Button block (make sure it's not a donation product, which has coupons disabled)
3. Enter an invalid coupon code and confirm you get the error message as above and the input is focused
4. Enter a valid coupon code and confirm the checkout is updated to include the discount
5. Click to remove the coupon code in the order review table and confirm the coupon is removed and the coupon form input is focused

### Other information:

### DIFF
--- a/src/modal-checkout/checkout.scss
+++ b/src/modal-checkout/checkout.scss
@@ -255,6 +255,13 @@
 				flex-grow: 1;
 			}
 		}
+		p.result {
+			margin: 0;
+			font-size: 14px;
+			&.error {
+				color: #cc1818;
+			}
+		}
 	}
 	#checkout_details { /* stylelint-disable-line */
 		display: flex;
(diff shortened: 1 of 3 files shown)


leogermani (approved):
:chefkiss:

Do you think it's ok to modify the class name and potentially lose support for other plugins customizations on this form? I think it's fine, just confirming

miguelpeixe:
> Do you think it's ok to modify the class name and potentially lose support for other plugins customizations on this form? I think it's fine, just confirming

It's something to keep an eye out for. I'm not aware of plugins that replace the coupon behavior.

It's worth noting that the code that this takes inspiration from, to be as close to the original behavior as possible, is 9 years old and is not exactly "hookable" to be extended. We'd be using such a hook if there was any 😅 . So I'd say it's safe to assume it shouldn't conflict with anything.

PR #1632, opened 2023-12-18
fix: avoid duplicity with linked guest authors

leogermani:
### All Submissions:

Fixes an issue with CAPs linked authors in the Hompepage posts block

1. Create a user and a guest author that share the same name
2. Create one post authored by each of these users
3. Edit the guest author and link them to the user with the same name
4. On `master` add a new Homepage block to a page and search for the author by it name in the "authors" filter box
5. Confirm you get two results
6. Checkout this branch
7. Confirm you now get only one result
8. Confirm the posts are queried correctly and you get both posts rendered in the block

### Other information:

---
  - https://app.asana.com/0/0/1204817890713184

### DIFF
--- a/src/blocks/author-profile/class-wp-rest-newspack-authors-controller.php
+++ b/src/blocks/author-profile/class-wp-rest-newspack-authors-controller.php
@@ -84,6 +84,8 @@ class WP_REST_Newspack_Authors_Controller extends WP_REST_Controller {
 		// Total number of users and guest authors.
 		$guest_author_total = 0;
 		$user_total         = 0;
+		$guest_authors = [];
+		$linked_guest_authors = [];
 
 		// Get Co-authors guest authors.
 		if ( $is_guest_author ) {
@@ -106,6 +108,14 @@ class WP_REST_Newspack_Authors_Controller extends WP_REST_Controller {
 
 			$guest_authors      = get_posts( $guest_author_args );
 			$guest_author_total = count( $guest_authors );
+		}
+
+		foreach ( $guest_authors as $ga ) {
+			$linked_guest_author = get_post_meta( $ga->ID, 'cap-linked_account', true );
+
+			if ( $linked_guest_author ) {
+				$linked_guest_authors[] = $linked_guest_author;
+			}
 		}
 
 		$users = [];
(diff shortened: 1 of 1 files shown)


miguelpeixe (changes requested):
I was not able to reproduce the issue this is trying to fix. My steps on the master branch were:

- Created a user "Jane Doe"
- Created a guest user "Jane Doe" 
- Created a "Jane Doe Post" and selected both users as author
- Edited guest user "Jane Doe" to map to the user with the same name
- Created a homepage posts block to filter only "Jane Doe" and this is what I got:

image

I edited the created post and was surprised to see that the post no longer had 2 authors. I suspect CAP resolved that when the GA was mapped.

Looking into the original issue in Asana, I believe the problem might be related to how the [`AutocompleteTokenField`](https://github.com/Automattic/newspack-blocks/blob/a0040b43a3f97c889e5bb5b5b96f07777b52a670/src/components/query-controls.js#L278-L284) behaves.

[The component "tokens" are the input labels](https://github.com/Automattic/newspack-blocks/blob/a0040b43a3f97c889e5bb5b5b96f07777b52a670/src/components/autocomplete-tokenfield.js#L155-L163), so you are not able to select multiple items that share the same name. I believe this should be resolved in 2 ways (not exclusively):
- Append a unique value to the input label, e.g. `Jane Doe (#{id})`
- Prevent mapped users from rendering as an option, so you get only one option

leogermani:
> Created a homepage posts block to filter only "Jane Doe" and this is what I got:

What should happen is that you will see both Janes in the dropdown. If they are not linked, that's expected, but if they are linked you should see only one.

Good idea adding the ID in the label, I'll have a look. But let me know if you can reproduce the above - and confirm the fix

miguelpeixe:
I just realized that I completely misinterpreted the PR 😅 

I thought it was duplicating posts and not the authors' names. I now understand what to test.

miguelpeixe (approved):
Now that I understand what to test I can confirm it's working as described!

The addition to the unique value on the label would be nice but not a blocker for this.

dkoo:
Screenshot 2023-12-22 at 11 00 06 AM

Those sneaky guest authors... 🕵️

leogermani:
> Screenshot 2023-12-22 at 11 00 06 AM
> 
> Those sneaky guest authors... 🕵️

oops, sorry for my english. Miguel got it because we speak the same language :)

miguelpeixe:
I'd never question the use of "duplicity" to mean duplicate 😆

dkoo:
I got what you meant, too. I actually had to look it up because I wasn't sure if my hunch was correct. 🤣

PR #1645, opened 2024-01-19
feat(reader-revenue): make NYP and Stripe Gateway optional

dkoo:
### All Submissions:

Makes Stripe Gateway and Name Your Price extensions optional.

See https://github.com/Automattic/newspack-plugin/pull/2866 for testing instructions.

### Other information:


src/blocks/donate/frontend/class-newspack-blocks-donate-renderer-base.php
```diff
@@ -33,8 +50,10 @@ public static function get_configuration( $attributes ) {
 
 		$configuration['defaultFrequency'] = $attributes['defaultFrequency'];
 
+		$is_manual = Newspack_Blocks::can_use_name_your_price() && $attributes['manual'];
```
miguelpeixe, line 53: This line is throwing `Warning: Undefined array key "manual"` because it's not an attribute that has a default value. It can be undefined.

src/blocks/donate/frontend/class-newspack-blocks-donate-renderer-base.php
```diff
@@ -33,8 +50,10 @@ public static function get_configuration( $attributes ) {
...
+
 		/* If block is in "manual" mode, override certain state properties with values stored in attributes */
-		if ( $attributes['manual'] ?? false ) {
+		if ( $is_manual ?? false ) {
```
miguelpeixe, line 56:
```suggestion
		if ( $is_manual ) {
```

The `?? false` is superfluous if we're not checking if `manual` is set in the associative array.

miguelpeixe (changes requested):
The fallback is working great! I have a few comments inline.

Non-blocking, but I miss having a greyed-out style here:

image

Even with the message, I still want to click the button.

src/blocks/donate/frontend/class-newspack-blocks-donate-renderer-base.php
dkoo: Fixed in a985f00

src/blocks/donate/edit/FrequencyBasedLayout.tsx
```diff
@@ -89,38 +89,75 @@ const FrequencyBasedLayout = ( props: { isTiered: boolean } & ComponentProps ) =
...
+											name={ `donation_value_${ frequencySlug }` }
+											defaultChecked={ true }
+										/>
+										<label
+											className="tier-select-label tier-label"
+											htmlFor={ `newspack-${ frequencySlug }-${ uid }-untiered-input` }
+										>
+											{ settings.currencySymbol +
+												untieredAmount.toFixed( 2 ).replace( /\.?0*$/, '' ) +
+												getFrequencyLabel( frequencySlug ) }
```
chickenn00dle, line 153: I wonder if this might cause an issue for languages where the amount should not precede the per month/year text? Might be a better idea to have the `getFrequencyLabel` method accept the amount and return a translated string which includes the amount in it?

src/blocks/donate/frontend/class-newspack-blocks-donate-renderer-base.php
```diff
@@ -16,6 +16,23 @@ abstract class Newspack_Blocks_Donate_Renderer_Base {
 
 	private static $configurations_cache = []; // phpcs:ignore Squiz.Commenting.VariableComment.Missing
 
+	/**
+	 * Get frequency label.
+	 *
+	 * @param string $frequency_slug Frequency slug.
+	 * @param bool   $hide_once_label Whether to hide the "once" label.
+	 */
+	public static function get_frequency_label( $frequency_slug, $hide_once_label = false ) {
```
chickenn00dle, line 25: Same comment as above re including amount in the translations here

src/blocks/donate/frontend/class-newspack-blocks-donate-renderer-base.php
```diff
@@ -33,8 +50,10 @@ public static function get_configuration( $attributes ) {
 
 		$configuration['defaultFrequency'] = $attributes['defaultFrequency'];
 
+		$is_manual = Newspack_Blocks::can_use_name_your_price() && ! empty( $attributes['manual'] );
```
chickenn00dle, line 53: nitpick: same comment re creating variable only used in one place

includes/class-modal-checkout.php
```diff
@@ -219,20 +219,18 @@ function ( $item ) {
...
-				if ( empty( $price ) ) {
-					$price = \WC_Name_Your_Price_Helpers::get_suggested_price( $product_id );
-				}
-				$min_price = \WC_Name_Your_Price_Helpers::get_minimum_price( $product_id );
-				$max_price = \WC_Name_Your_Price_Helpers::get_maximum_price( $product_id );
-				$price     = ! empty( $max_price ) ? min( $price, $max_price ) : $price;
-				$price     = ! empty( $min_price ) ? max( $price, $min_price ) : $price;
-
-				$cart_item_data['nyp'] = (float) \WC_Name_Your_Price_Helpers::standardize_number( $price );
+		$is_product_nyp = \Newspack_Blocks::can_use_name_your_price() ? \WC_Name_Your_Price_Helpers::is_nyp( $product_id ) : false;
```
chickenn00dle, line 222: nitpick: is it necessary to create a variable here considering this is only used in one place?

chickenn00dle (commented):
Nice work @dkoo!

Tested and everything works as expected. I left a few comments in the code, mostly nitpicks that can be ignored, but one question regarding translations that should be answered before moving forward with this.

src/blocks/donate/edit/FrequencyBasedLayout.tsx
dkoo: That's a good point, although I had some trouble conceptualizing how we'd translate such a string when it has to include HTML tags for styling, too. 3701068 is my best attempt—how's it look?

src/blocks/donate/frontend/class-newspack-blocks-donate-renderer-base.php
dkoo: Nope, not necessary—updated in 42e145b.

includes/class-modal-checkout.php
dkoo: Nope, not necessary—updated in 42e145b.

src/blocks/donate/frontend/class-newspack-blocks-donate-renderer-base.php
dkoo: See https://github.com/Automattic/newspack-blocks/pull/1645#discussion_r1470424015

src/blocks/donate/frontend/class-newspack-blocks-donate-renderer-base.php
```diff
@@ -16,6 +16,54 @@ abstract class Newspack_Blocks_Donate_Renderer_Base {
...
+		* Example string: <h3 class="wpbnbd__tiers__amount__number">$100</h3>
+		*                 <span class="wpbnbd__tiers__amount__frequency">per month</span>
+		*/
+		$formatted_amount = sprintf(
+			_x( // phpcs:ignore WordPress.WP.I18n.MissingTranslatorsComment
+				'%1$s %2$s',
+				'%1$s is the amount (with currency symbol). %2$s is the frequency of the donation. Inludes embedded HTML tags for styling.',
+				'newspack-blocks'
+			),
+			'<h3 class="wpbnbd__tiers__amount__number">' . $currency_symbol . $amount . '</h3>',
```
chickenn00dle, line 60: I think this pattern for translating is fine, but I think we will still run into issues with languages where currencies and amounts might need to be swapped. Can we use something like the [wc_price](https://github.com/woocommerce/woocommerce/blob/5d7f6acbcb387f1d51d51305bf949d07fa3c4b08/includes/wc-formatting-functions.php#L561) helper function for this?

chickenn00dle: Alternatively, I'm not sure how tied we are to the specific wording here, but WC Subscriptions has done alot of the heavy lifting for formatting price strings, which can get super tricky. Can we rely on something like [wcs_price_string](https://github.com/Automattic/woocommerce-subscriptions-core/blob/81d809a476e87c260492d4cc0413818d85e123cc/includes/wcs-formatting-functions.php#L33)?

src/blocks/donate/edit/FrequencyBasedLayout.tsx
```diff
@@ -96,31 +96,63 @@ const FrequencyBasedLayout = ( props: { isTiered: boolean } & ComponentProps ) =
...
+											name={ `donation_value_${ frequencySlug }` }
+											defaultChecked={ true }
+										/>
+										<label
+											className="tier-select-label tier-label"
+											htmlFor={ `newspack-${ frequencySlug }-${ uid }-untiered-input` }
+										>
+											<div
+												dangerouslySetInnerHTML={ {
+													__html: getFrequencyLabel(
```
chickenn00dle, line 142: I wonder if we can have this string passed from the backend via `newspack_blocks_data` since the logic for creating the string already lives there? Then we can remove the duplicate logic from the frontend.

src/blocks/donate/edit/FrequencyBasedLayout.tsx
chickenn00dle: Thanks @dkoo! I left some additional review comments for this 👍

dkoo: Great suggestions here! Thank you for the guidance on this part—admittedly our work around making sure things are translatable has been lacking. 😅

src/blocks/donate/edit/FrequencyBasedLayout.tsx
dkoo: Done in 5f9479a. Because the block needs to be able to dynamically render values from attributes, this may look a little janky with the string replacement, but I'm open to better suggestions.

src/blocks/donate/frontend/class-newspack-blocks-donate-renderer-base.php
dkoo: 5f9479a implements `wcs_price_string` with a fallback to `wc_price` if not available (if that's not available, the site has bigger problems).

src/blocks/donate/edit/FrequencyBasedLayout.tsx
chickenn00dle: Ah. I see what you mean. It definitely does seem a bit overly complex. I'm not sure of a cleaner way to handle this with the existing design for the donate block. My suggestion would be to remove the `per X`/`once` from the amount strings altogether as this is implied based on the category tab selection at the top (one-time/monthly/annually). But I think this might be outside of the scope of this PR.

I think what we have here is fine, although there is an edge case where the publisher has an empty value set, for example while editing an amount. The block shows `NaN`:

Screenshot 2024-02-02 at 10 24 48

chickenn00dle (changes requested):
Sorry for the delay on getting back to this one @dkoo. Almost there. Just one final issue commented here: https://github.com/Automattic/newspack-blocks/pull/1645#discussion_r1476211787

src/blocks/donate/edit/FrequencyBasedLayout.tsx
dkoo: 5b8470c forces the value to be `0` in this case. Maybe still not ideal, but better than `$NaN`?

chickenn00dle: Yeah. I think it makes sense to set this to 0 when the input is empty. One other issue I noticed just now while retesting this is if the input is empty, we need to make sure the below $X validation still functions and prevents us from saving.

Here it is with an input value less than 5 for me:
Screenshot 2024-02-06 at 11 37 44

And while the input is empty:
Screenshot 2024-02-06 at 11 37 51

I think instead of forcing how the null value is rendered in the preview, we can force null values to be stored as 0 somewhere around here: https://github.com/Automattic/newspack-blocks/blob/5b8470cd35fdc1d24113ebb60ebeb220dbbc622c/src/blocks/donate/edit/index.tsx#L109

chickenn00dle: I also want to note while I was retesting this, I did some more thinking about the complexity of generating the translation string with dynamic pricing and HTML elements, and realized we can probably reduce the complexity by removing the `span` wrappers from the currency symbol and price, and instead placing them only around the frequency string.

We can achieve the same style results via CSS this way by setting the `.wpbnbd__tiers__amount__value` text to have the bolded and enlarged size and set the frequency span to have the unbolded smaller size.

This is how it is rendered right now:
```
<span class="wpbnbd__tiers__amount__value">
    <span class="woocommerce-Price-amount amount">
        <bdi>
            <span class="woocommerce-Price-currencySymbol">$</span>5
        </bdi>
    </span>
    once
</span>
```

We can probably achieve the same result with this:
```
<span class="wpbnbd__tiers__amount__value">
    $5
    <span class="some-class-for-frequency-label">
        once
    </span>
</span>
```

I'm okay if we tackle this another time though since this PR has become quite involved.

dkoo: > I also want to note while I was retesting this, I did some more thinking about the complexity of generating the translation string with dynamic pricing and HTML elements, and realized we can probably reduce the complexity by removing the `span` wrappers from the currency symbol and price, and instead placing them only around the frequency string.

That's a good call, but yes, let's leave for another PR.

> I think instead of forcing how the null value is rendered in the preview, we can force null values to be stored as 0 somewhere around here: 

At first I tried forcing empty input values to be stored as `0`, but this results in the input itself showing `0` when you delete the values in the input, which is potentially confusing to the user. Agree that we should still validate against the minimum donation value, though.

dkoo: > Agree that we should still validate against the minimum donation value, though.

Implemented in 9f041db

chickenn00dle (approved):
Nice! thanks for dealing with all the back and forth on this one!

PR #1649, opened 2024-01-23
feat(modal-checkout): update thank you page with new newspack ui

chickenn00dle:
### All Submissions:

Part of `1206197591054328/1206405274735557/f`

This applies the newspack ui to the modal checkout thank you page and prepares the rest of the modal checkout to be updated to the new ui. As a consequence this breaks the styling of the modal checkout and billing forms a bit, but I will work on those in a follow up PR so this one doesn't get too big:

Checkout form is expected to look like this:
Screenshot 2024-01-30 at 11 45 15

Billing info form is expected to look like this:
Screenshot 2024-01-30 at 11 45 25

1. With newspack plugin on `feat/update-checkout-block-thankyou-ui` branch, trigger modal checkout via donate button
2. Complete checkout until you see the thank you page
3. Confirm the thankyou modal looks as expected with the udpated ui

Screenshot 2024-01-25 at 14 31 24

4. Switch to `master` on newspack plugin and repeat the above steps
5. Confirm nothing appears to be broken (The thank you page should not have the new styles applied and just be a generic transaction summary)

### Other information:


includes/class-modal-checkout.php
```diff
@@ -289,19 +290,20 @@ public static function render_modal_markup() {
...
 						<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" role="img" aria-hidden="true" focusable="false">
 							<path d="M19 6.41L17.59 5 12 10.59 6.41 5 5 6.41 10.59 12 5 17.59 6.41 19 12 13.41 17.59 19 19 17.59 13.41 12 19 6.41z"/>
 						</svg>
-					</a>
+					</button>
 				</header>
-				<section class="newspack-blocks-modal__content">
-					<div class="newspack-blocks-modal__spinner">
+				<section class="newspack-ui__modal__content">
+					<div class="newspack-ui__spinner">
```
chickenn00dle, line 306: I couldn't find a place for a loading state in the main newspack ui changes, so moved the logic for this in https://github.com/Automattic/newspack-plugin/pull/2878. So that will need to be checked out to ensure the spinner correctly functions while testing.

src/modal-checkout/modal.js
```diff
@@ -154,30 +166,32 @@ domReady( () => {
...
 				if ( window.newspackReaderActivation?.overlays ) {
 					modalCheckout.overlayId = window.newspackReaderActivation?.overlays.add();
 				}
 				iframeResizeObserver = new ResizeObserver( entries => {
 					if ( ! entries || ! entries.length ) {
 						return;
 					}
-					const contentRect = entries[ 0 ].contentRect;
-					if ( contentRect ) {
-						modalContent.style.height = contentRect.top + contentRect.bottom + 'px';
```
chickenn00dle, line 175: I was running into a strange bug where the iframe body was not the same height as the iframe document itself which was causing some issues with scrollbars. I got around this by getting the height of the iframe via `offsetHeight` instead which seems to have addressed the problem. Not sure if anyone has seen this issue before or if this is the best approach.

src/modal-checkout/modal.js
```diff
@@ -154,30 +166,32 @@ domReady( () => {
...
+
+					const iframeHeight = iframe.contentDocument.documentElement.offsetHeight;
+					if ( iframeHeight ) {
+						// Account for newspack ui modal content padding.
+						modalContent.style.height = iframeHeight + 48 + 'px';
```
chickenn00dle, line 187: Just noting that to get the iframe to load without a scollbar and to get the modal to load the iframe with the correct height, I had to account for the padding of the modal content here.

src/modal-checkout/modal.js
miguelpeixe: This is happening because the `__content` is applying a padding that, in the case of the modal checkout, is expected to live inside the iframe body ([here](https://github.com/Automattic/newspack-blocks/pull/1649/files#diff-93359786372d1c41af813037af697f33c6a6da8d4a48622a3c18d8e6f26f1cd7L2)), so the `top` and `bottom` values it picks up is not including the padding generated by parent element.

I'm not sure which strategy is best, but I remember having issues with the `offsetHeight` providing a consistent value across different stages of checkout/iframe changes.

laurelfulford: I've suggested that we add the `newspack-ui__modal__content` CSS class to the `<body>` tag inside of the iframe further up in the review. This ends up adding more padding inside of the iframe, but I think it could help with this: we could use CSS to get rid of the problem spacing around the iframe, so the padding's all handled inside. 

Something like this in Newspack UI could work if we can all iframes should act like this: 

```
.newspack-ui__modal__content iframe {
	width: calc( 100% + var( --newspack-ui-spacer-8, 48px ) );
	max-width: unset;
	height: calc( 100% + var( --newspack-ui-spacer-8, 48px ) );
	border: none;
	margin: calc( -1 * var( --newspack-ui-spacer-5, 24px ) );
}
```

... or if it's specific to the modal checkout, we can add it to the `iframe[name="newspack_modal_checkout_iframe"]` selector in the blocks instead. 

I did a rough test with the original JS and it seems to work! Let me know what you think 🙂

src/modal-checkout/templates/thankyou.php
```diff
@@ -47,68 +47,41 @@ function newspack_blocks_replace_login_with_order_summary() {
...
-					<strong><?php echo $order->get_order_number(); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?></strong>
-				</li>
-			</ul>
-		<?php else : ?>
-			<p class="woocommerce-notice woocommerce-notice--error woocommerce-thankyou-order-failed">
-				<?php esc_html_e( 'Unfortunately your order cannot be processed. Please attempt your purchase again.', 'newspack-blocks' ); ?>
+	<?php if ( $is_success ) : ?>
+		<div class="newspack-ui__box newspack-ui__box--success newspack-ui__box__text-center">
+			<span class="newspack-ui__icon newspack-ui__icon--success">
+				<?php // TODO: How should we handle sharing icons across plugins? ?>
```
laurelfulford, line 56: Ooooh, good point! 🤔 

I think at least these SVGs are just decorative; we might be able to use only CSS for them -- something like: 

```
.newspack-ui__icon:after {
    background-color: currentcolor;
    content: '';
    display: block;
    height: 24px;
    mask-size: 24px;
    width: 24px;
}

.newspack-ui__icon--success::after {
	mask: url('../path/to/svg/asset.svg') 0 0 no-repeat;
}
```

We'd still need to duplicate must-have SVGs (like the close icon), so this might not be an ideal replacement!

includes/class-modal-checkout.php
```diff
@@ -287,21 +288,27 @@ public static function render_modal_markup() {
 		*
 		* @param string $title The title.
 		*/
-		$title = apply_filters( 'newspack_blocks_modal_checkout_title', __( 'Complete your transaction', 'newspack-blocks' ) );
+		$title        = apply_filters( 'newspack_blocks_modal_checkout_title', __( 'Complete your transaction', 'newspack-blocks' ) );
+		$class_prefix = 'newspack-blocks';
+
+		if ( class_exists( '\Newspack\Newspack_UI' ) ) {
+			$class_prefix = 'newspack-ui';
+		}
```
laurelfulford, line 296: This is a nice approach to swapping the classes!

src/modal-checkout/modal.js
```diff
@@ -129,7 +139,9 @@ domReady( () => {
 			form.addEventListener( 'submit', ev => {
 				const formData = new FormData( form );
 				// Clear any open variation modal.
-				variationModals.forEach( variationModal => variationModal.classList.remove( 'open' ) );
+				variationModals.forEach( variationModal => {
```
laurelfulford, line 142: The variations modal isn't quite marked up like the modal checkout ([variations](https://github.com/Automattic/newspack-blocks/pull/1649/files#diff-74a92f16c32d2a94751337aaff7d24de10881675f81809bae9d09e79cc47e10eL350) vs. [modal checkout](https://github.com/Automattic/newspack-blocks/pull/1649/files#diff-74a92f16c32d2a94751337aaff7d24de10881675f81809bae9d09e79cc47e10eR298-R299)), so changing the JS for it causes the Checkout Button to stop opening if it's for a variable product.

(To recreate, you need to have a variable product in Woo, and set up the Checkout Button block to use it, and check "Allow the reader to select the variation before checkout" -- here's [a little more about the block](https://help.newspack.com/publishing-and-appearance/blocks/checkout-button-block/) and [that specific feature](https://help.newspack.com/publishing-and-appearance/blocks/checkout-button-block/#customize-product-variations)!).

When I roll back to the epic/ras-acc branch the button still works. I think it should ultimately all be updated (with the same markup for the modal container and overlay), but that can be done separately 🙂 (if you like, I'm happy to take a swing at it!).

laurelfulford (changes requested):
Thanks for digging into this, @raz! 🙂 I think it's on the right track overall! I left some comments/feedback -- please let me know if you have any questions at all about what I wrote!

src/modal-checkout/modal.js
chickenn00dle: Nice! Thanks for the clarity and suggestions y'all 🙌 

I was able to apply these styles and revert the iframe resizing logic in https://github.com/Automattic/newspack-blocks/pull/1649/commits/c8b032947bf6f540c64a67d7eaf8bf32ac89c320

includes/class-modal-checkout.php
```diff
@@ -491,10 +505,10 @@ public static function get_checkout_template( $template ) {
 			<link rel="profile" href="https://gmpg.org/xfn/11" />
 			<?php wp_head(); ?>
 		</head>
-		<body id="newspack_modal_checkout">
+		<body class="newspack-ui" id="newspack_modal_checkout">
```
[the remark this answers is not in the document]
chickenn00dle: Ah. Makes sense! I've added the modal specific class in https://github.com/Automattic/newspack-blocks/pull/1649/commits/c8b032947bf6f540c64a67d7eaf8bf32ac89c320

src/modal-checkout/templates/thankyou.php
chickenn00dle: Gotcha! I'm gonna leave this one as is for now and maybe we can tackle this in a follow-up PR?

src/modal-checkout/modal.scss
```diff
@@ -1,189 +1,213 @@
...
-	0% {
-		transform: rotate( 0deg );
-	}
-	50% {
-		transform: rotate( 180deg );
-	}
-	100% {
-		transform: rotate( 360deg );
+// Checkout modal specific styles.
+.newspack-ui {
```
chickenn00dle, line 5: I decided to keep the iframe specific styles here rather than in the main newspack ui library. My thought is we should let each iframe dictate how it should be styled rather than enforce styles top down. That said, I don't feel too strongly about this, and am happy to move this over to the newspack ui library anyway.

adekbadek:
@chickenn00dle – 

> As a consequence this breaks the styling of the modal checkout form a bit, but I will work on that in a follow up PR so this one doesn't get too big.

Can you please add a list of these breakages to PR description? I've seen this one when testing, not sure if it's expected:

image

includes/class-modal-checkout.php
```diff
@@ -1209,5 +1229,12 @@ public static function jetpack_active_modules( $modules ) {
 		}
 		return $modules;
 	}
+
+	/**
+	 * Get the relevant class prefix (newspack-blocks or newspack-ui)depending on whether Newpack plugin is active.
```
adekbadek, line 1234:
```suggestion
	 * Get the relevant class prefix (newspack-blocks or newspack-ui) depending on whether Newpack plugin is active.
```

src/modal-checkout/modal.js
```diff
@@ -25,47 +27,54 @@ function domReady( callback ) {
 	document.addEventListener( 'DOMContentLoaded', callback );
 }
 
-const triggers =
-	'.wpbnbd.wpbnbd--platform-wc,.wp-block-newspack-blocks-checkout-button,.newspack-blocks-variation-modal';
+const CLASSNAME_BASE = newspackBlocksModal.has_newspack_ui ? 'newspack-ui' : 'newspack-blocks';
```
adekbadek, line 30: To avoid duplication, I'd suggest passing the value of `Modal_Checkout::get_class_prefix()` directly.

adekbadek (commented):
LGTM, but I'll defer to @laurelfulford as the primary reviewer.

includes/class-modal-checkout.php
```diff
@@ -54,6 +54,9 @@ public static function init() {
 		add_action( 'woocommerce_checkout_before_customer_details', [ __CLASS__, 'render_before_customer_details' ] );
 		add_filter( 'woocommerce_enable_order_notes_field', [ __CLASS__, 'enable_order_notes_field' ] );
 		add_action( 'woocommerce_checkout_process', [ __CLASS__, 'wcsg_apply_gift_subscription' ] );
+		add_filter( 'woocommerce_order_received_verify_known_shoppers', '__return_false' );
+		// TODO: Remove once we apply auth flow to checkout modal.
+		add_filter( 'newspack_reader_activation_should_render_auth', '__return_false' );
```
chickenn00dle, line 59: Noting that I've added this with the latest merge of master into epic/ras-acc since we have another task to integrate the auth flow into checkout.

includes/class-modal-checkout.php
chickenn00dle: Good catch! Fixed in https://github.com/Automattic/newspack-blocks/pull/1649/commits/6fa14705166d8070e6f3139fe2f6cb6ce61c7c37

src/modal-checkout/modal.js
chickenn00dle: Good call! Updated in https://github.com/Automattic/newspack-blocks/pull/1649/commits/e188f8ff43b1bc5855c511963dabee8cbe65c7f8

chickenn00dle:
> Can you please add a list of these breakages to PR description? I've seen this one when testing, not sure if it's expected

Sure thing @adekbadek! 

It is expected. This is the checkout form which this PR only changes the modal wrapper for. I've updated the description to show the parts of the checkout flow which are expected not to be fully styled as of this PR.

src/modal-checkout/templates/thankyou.php
laurelfulford: That makes sense to me! 👍

src/modal-checkout/modal.js
laurelfulford: This looks good overall! 

I did notice the spinner's no longer displaying -- it's using the Newspack UI classes but spinner styles don't exist there. 

Similar to [this comment](https://github.com/Automattic/newspack-blocks/pull/1649/files#r1467956484), this could be something that just lives in the checkout modal, I think the CSS classes just need to be updated in the markup to not switch to the `newspack-ui__` classes, and the CSS itself might need a bit of a massage so the structure still works with how the modal might be structured with the `newspack-ui` classes 🙂 

image

laurelfulford (changes requested):
Apologies for taking a bit to get back to this, @chickenn00dle! I added some comments, but I may have missed where some of this is being addressed elsewhere! If that's the case, please disregard -- ditto if any of this would be better to handle in separate PRs 🙂 

Just let me know if you have any questions at all!

includes/class-modal-checkout.php
chickenn00dle: > It looks like our Sign Up / Sign In button styles are included in the auth.css file this enqueues -- not sure if just deleting the other files would be better, or maybe something else?

Ah! Good catch. I should note that I am only temporarily bypassing the auth flow for this and the other modal checkout ui update PR. But will be incorporating auth into modal checkout with 0/1206274567818686/1205668937699524. Not sure if it's okay to allow these buttons to appear broken in epic/ras-acc for a short time only?

src/modal-checkout/modal.js
chickenn00dle: For now, I've gone ahead and added the existing variation styles to newspack-ui in https://github.com/Automattic/newspack-plugin/pull/2878, but if they need to be updated, then it'd be great if you could take a crack at it!

src/modal-checkout/modal.js
chickenn00dle: > I did notice the spinner's no longer displaying -- it's using the Newspack UI classes but spinner styles don't exist there.

I've actually moved the styles for the spinner over to newspack-ui in Automattic/newspack-plugin#2878. Sorry if it wasn't clear, but you'll need to check that branch out in the main Newspack plugin before testing this to see the spinner in action!

chickenn00dle:
Thanks for having another look @laurelfulford!

I've commented on your feedback. I'm happy to move styles for things into this PR but wanted to make sure you were aware of all of my updates before making any changes.

includes/class-modal-checkout.php
laurelfulford: Ah gotcha! 

> Not sure if it's okay to allow these buttons to appear broken in epic/ras-acc for a short time only?

I think that's probably fine -- it looks like some other stuff will be in flux as well!

src/modal-checkout/modal.js
laurelfulford: Oh nice -- thanks @chickenn00dle! 🙌

src/modal-checkout/modal.js
laurelfulford: Ah thanks! I'm afraid I haven't been following things as closely as I should this week. When I have that branch checked out too, the spinner is there! 🙂

laurelfulford (approved):
Thanks for all the exhaustive back and forth on this, @chickenn00dle! I think stuff is looking good overall -- thank you for helping me get up to speed where the other pieces are! 

I think this piece is good to go, and probably will probably make things easier for the other PRs to have it merged 😅

adekbadek:
Note: https://github.com/Automattic/newspack-blocks/pull/1653 updates the styling of the checkout modal to fix the issues introduced in this PR.

PR #1650, opened 2024-01-23
feat(modal-checkout): add filter for order button text

dkoo:
### All Submissions:

Adds a filter to allow customizing the text for the button to complete an order in modal checkout.

1. Check out this branch.
2. As a reader, start the following transactions via modal checkout (Donate block or Checkout Button block for non-donations):
  - One-time donation
  - Recurring donation
  - One-time non-donation purchase
  - Subscription non-donation purchase

3. Confirm that the order button text displays as expected, unchanged from behavior in `master` (for any Newspack donation products: "Donate now", or whatever the default button label is for non-donation products).
4. Add some custom code to your test site in a temporary plugin or in your theme's functions.php:

```php
/**
 * Customize the order button text.
 *
 * @param string $text The order button text.
 * @param bool   $is_donation True if the current cart contains a Newspack donation product.
 * @param bool   $is_subscription True if the current cart contains a subscription product.
 * @return string The new text.
 */
function newspack_test_order_button_text( $text, $is_donation, $is_subscription ) {
	if ( $is_donation ) {
		$text = $is_subscription ? 'Custom text for recurring donations' : 'Custom text for one-time donations';
	} else {
		$text = $is_subscription ? 'Custom text for non-donation subscriptions' : 'Custom text for one-time non-donation purchases';
	}

	return $text;
}
\add_filter( 'newspack_blocks_modal_checkout_order_button_text', 'newspack_test_order_button_text', 10, 3 );

```

5. Repeat step 2, and confirm that your custom text is shown as expected in each case.

Screenshot 2024-01-23 at 3 06 38 PM

### Other information:


miguelpeixe (commented):
In the RAS-ACC project the logic for order button text has been decoupled so the Donation related text [lives in the `class-donations.php` of `newspack-plugin`](https://github.com/Automattic/newspack-plugin/blob/dd4f3595fdcc438b8835b2bd56e16688b6c48c25/includes/class-donations.php#L1039-L1049) and the modal checkout related change [lives in the `class-modal-checkout.php` of `newspack-blocks`](https://github.com/Automattic/newspack-blocks/blob/f5c707e0e48a0f04f4d7f4e64eee0b34539c01a8/includes/class-modal-checkout.php#L1041-L1059). This will cause a conflict with the proposed filter when the project epic branch merges into master, since there's no longer donation related code in the modal checkout.

Although I don't mind the new filter, these can already be changed without additional code on our end by adding a filter the same way `order_button_text` works.

dkoo:
@miguelpeixe thanks for highlighting. And true that the text can already be customized without a new filter, but I think a new filter could make the customization that much easier for non-engineers now, and also make it easier to implement a UI setting if we want to in the future. Will reopen this against the epic.

PR #1667, opened 2024-02-07
fix: properly handle states when Country field is disabled

leogermani:
### All Submissions:

Fixes an issue in the checkout modal when the "Country/Region" field is disabled, but the State field is enabled.

In those cases, in the first step of the modal, users will see the States selectbox and select a state, but in the second step, the hidden select field will not have the correct states loaded.

The end result is that the donation submission will either fail or assign the user to a wrong state (AF, which is the Afghanistan code, the first option in the select).

1. On `trunk` go to Newspack > Reader Revenue. Enable the state and Address field, but disable "Country"
2. As an anonymous reader, try to make a donation. Confirm you can select the state.
3. On the second step of the donation modal, inspect the modal and confirm you see the hidden states SELECT listing countries instead of states

Captura de tela de 2024-02-07 13-40-09

4. Try to submit the donation
5. Confirm the donation either fails, or if it succeeds, confirm that the order is not assigned to the state you chose.
6. Checkout this branch
7. Reproduce the steps above and confirm that the hidden select shows the correct states and that you can make the donation
8. Confirm the order has the correct state
9. Change your store default country and confirm things still work
10. try to break it :) 

### Other information:

### DIFF
--- a/src/modal-checkout/templates/checkout-form.php
+++ b/src/modal-checkout/templates/checkout-form.php
@@ -116,7 +116,11 @@ if ( ! $checkout->is_registration_enabled() && $checkout->is_registration_requir
 			</div>
 			<?php foreach ( $form_billing_fields as $key => $value ) : ?>
 				<?php if ( 'state' === $key ) : ?>
-					<?php $wc_states = WC()->countries->get_states( $form_billing_fields['country'] ); ?>
+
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Fixes the issue!

PR #1671, opened 2024-02-14
fix(hpp): filter invalid autocomplete tokens

chickenn00dle:
### All Submissions:

Closes `0/1206331646877116/1204088636207233/f`.

This fixes an issue where the homepage posts block was allowing invalid categories, authors, and tags to be saved, which resulted in an error in the block editor.

This fixes this by filtering out any invalid tokens for the autocomplete component.

1. Add the homepage posts block to any post/page
2. In block settings, input any invalid category name then press enter
3. Save the post/page
4. Reload the page
5. On `trunk` you should see the block render an error
Screenshot 2024-02-14 at 14 45 17
6. On this branch, you should not!

### Other information:

---
  - https://app.asana.com/0/0/1204088636207233

### DIFF
--- a/src/blocks/homepage-articles/utils.ts
+++ b/src/blocks/homepage-articles/utils.ts
@@ -1,7 +1,7 @@
 /**
  * External dependencies
  */
-import { times, isEqual, isUndefined, pick, pickBy } from 'lodash';
+import { times, isEqual, isNull, isUndefined, pick, pickBy } from 'lodash';
 
 /**
  * WordPress dependencies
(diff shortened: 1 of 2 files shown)


adekbadek (changes requested):
And invalid value should not be added to the block attributes in the first place. With this fix in place, the block still does not render any posts because it still tries to filter by an invalid category. 

image

chickenn00dle:
Thanks for having a look @adekbadek!

> And invalid value should not be added to the block attributes in the first place

Hmm. This is what this PR is doing. Attributes look to be saved via this `AutocompleteTokenField` component's [onChange prop](https://github.com/Automattic/newspack-blocks/blob/ebdb4ab1944db0c7c1c95a96025d7fab54fb538b/src/components/autocomplete-tokenfield.js#L156) which is passed along from the `edit.js` file: https://github.com/Automattic/newspack-blocks/blob/ebdb4ab1944db0c7c1c95a96025d7fab54fb538b/src/blocks/homepage-articles/edit.js#L363 

If you had previously saved an invalid attribute while in `trunk` then switch to this branch, you will need to re-save the attribute in order for those invalid values to be removed from attributes.

That said, for folks who have already saved null values in attributes, and don't wish to go through the process of re-saving after this change, I've added some validation when querying for posts here: https://github.com/Automattic/newspack-blocks/pull/1671/commits/19a30e43be5aedc3b799d3f6604c90cdb055d62f

adekbadek (approved):
Works fine now!

> If you had previously saved an invalid attribute while in trunk then switch to this branch

The testing instructions do not state that block should be re-added after switching the branch, so I assumed this would fix existing attributes.

chickenn00dle:
> The testing instructions do not state that block should be re-added after switching the branch, so I assumed this would fix existing attributes.

Ah. Sorry about that @adekbadek!

PR #1672, opened 2024-02-15
fix(author-blocks): handle HTML in author bio

adekbadek:
### All Submissions:

Adds HTML handling to the editing part of author blocks. 

See 1200550061930446-as-1205128714156103/f

1. Edit an author and insert some HTML in their bio
2. Insert Author Profile and Author List blocks on a page, and set them to display the edited author 
3. Observe the HTML is rendered correctly in the editor

### Other information:

### DIFF
--- a/src/blocks/author-profile/single-author.js
+++ b/src/blocks/author-profile/single-author.js
@@ -4,6 +4,7 @@
  * WordPress dependencies
  */
 import { __ } from '@wordpress/i18n';
+import { RawHTML } from '@wordpress/element';
 
 /**
  * External dependencies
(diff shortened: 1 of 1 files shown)


laurelfulford (approved):
This looks good, and fixes the issue for me! 🙌 

I spotted something else in testing with this fix that could probably be handled separately: on the front-end, any line breaks in the user descriptions are turned into paragraphs thanks to [wpautop()](https://developer.wordpress.org/reference/functions/wpautop/), but they are presented as one paragraph in the editor. 

It looks like there's [a similar option for the block editor](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-autop/) that could be used.

(It's worth nothing neither of these have been reported yet, but are just weird edge cases I've run into in testing! Let me know if I can spin up a separate issue for the paragraph thing!).

adekbadek:
Great catch, thanks Laurel! Fixed in 649bca419cf36cd30909f3891c1df65bf51fd35f

PR #1681, opened 2024-02-27
 fix: avoid use of @wordpress modules on front-end 

dkoo:
### All Submissions:

I'm not sure this is 100% true all the time, but at least on my local site `@wordpress/*` JS packages are not automatically present on the front-end unless explicitly enqueued, and in my experience tend to be pretty bloated and non-performant for front-end purposes anyway. This change avoids a potential unhandled JS error if `@wordpress/i18n` can't be loaded on the front-end, and moves the string translation functions to the back-end where we can be certain `_x()` already exists.

Confirm that the strings translated via `_x()` and localized to the `newspackBlocksModal.labels` object display correctly when loading a checkout modal.

### Other information:

### DIFF
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -416,7 +416,17 @@ final class Modal_Checkout {
 			[
 				'newspack_class_prefix' => self::get_class_prefix(),
 				'labels'                => [
-					'auth_modal_title' => self::get_modal_title(),
+					'auth_modal_title'     => self::get_modal_title(),
+					'signin_modal_title'   => _x(
+						'Sign in to complete transaction',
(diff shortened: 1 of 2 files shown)


chickenn00dle (approved):
Works as expected @dkoo!

Not a blocker here and maybe something to address in a follow-up PR, but what do you think about making the Login and Register modal titles filter-able like we do with the Checkout modal title?

dkoo:
> Not a blocker here and maybe something to address in a follow-up PR, but what do you think about making the Login and Register modal titles filter-able like we do with the Checkout modal title?

That's a great idea for consistency! Maybe all of filterable and translatable strings can be combined into a single array, too? Something for another PR.

PR #1687, opened 2024-03-05
fix: show credit option for homepage posts & carousel blocks

dkoo:
### All Submissions:

Adds new block attribute options to show credits for Homepage Posts and Post Carousel blocks (and to show captions for the latter, too).

Closes `1206709674365921/1206688511006431`.

1. Check out this branch.
2. Add a Homepage Posts block and a Post Carousel block to the page, and configure so they show articles with featured images that have captions and credits.
3. Enable captions + credits for both blocks and confirm the captions are shown in both editor and front-end context:

| Editor  | Front-end |
| ------------- | ------------- |
| Screenshot 2024-03-05 at 2 14 27 PM  | Screenshot 2024-03-05 at 2 19 19 PM  |

4. Play with different combinations of caption and/or credit and confirm that the right info appears as expected in both editor and front-end. Try:
  - Caption enabled, credit disabled
  - Caption disabled, credit enabled
  - Both disabled

### Other information:

---
  - https://app.asana.com/0/0/1206688511006431

### DIFF
--- a/src/blocks/homepage-articles/view.scss
+++ b/src/blocks/homepage-articles/view.scss
@@ -346,10 +346,6 @@
 			text-overflow: ellipsis;
 			z-index: 3;
 
-			&.has-credit {
-				text-align: left;
-			}
-
 			@include mixins.media( desktop ) {
 				padding: 0 1.5rem;
 			}
--- a/src/blocks/homepage-articles/view.scss
+++ b/src/blocks/homepage-articles/view.scss
@@ -345,6 +345,11 @@
 			text-align: right;
 			text-overflow: ellipsis;
 			z-index: 3;
+			
+			a,
+			a:visited {
+			    color: #fff;
+			}
 
 			@include mixins.media( desktop ) {
 				padding: 0 1.5rem;
--- a/src/blocks/homepage-articles/view.scss
+++ b/src/blocks/homepage-articles/view.scss
@@ -344,7 +344,11 @@
 			right: 0;
 			text-align: right;
 			text-overflow: ellipsis;
-			z-index: 2;
+			z-index: 3;
+
+			&.has-credit {
+				text-align: left;
+			}
 
 			@include mixins.media( desktop ) {
 				padding: 0 1.5rem;
--- a/includes/class-newspack-blocks-api.php
+++ b/includes/class-newspack-blocks-api.php
@@ -74,16 +74,6 @@ class Newspack_Blocks_API {
 	}
 
 	/**
-	 * Get thumbnail featured image captions for the rest field.
-	 *
-	 * @param array $object_info The object info.
-	 * @return string|null Image caption on success, null on failure.
-	 */
-	public static function newspack_blocks_get_image_caption( $object_info ) {
-		return (int) $object_info['featured_media'] > 0 ? trim( wp_get_attachment_caption( $object_info['featured_media'] ) ) : null;
-	}
-
-	/**
 	 * Get author info for the rest field.
 	 *
 	 * @param array $object_info The object info.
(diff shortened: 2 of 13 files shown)


dkoo:
Paging @kevinzweerink and @thomasguillot for design review! Just want to make sure the typography and colors look okay to you.

kevinzweerink:
I don't have time to test this locally before I head out for the evening, but the type and color look good to me in the screenshot. I can test locally first thing tomorrow morning, but if it's urgent to merge this I'm comfortable approving based on the screenshot and can double-check tomorrow. As long as we're using standard type sizes and theme colors it looks good to me :)

dkoo:
Thanks, @kevinzweerink! I'm comfortable with your 👍  based on the screenshots. We can always tweak the styles separately after the fix goes out, if we want to.

laurelfulford (changes requested):
This looks good @dkoo! I found a couple small things, but nothing major:

**Homepage Posts block + Featured image behind:** 
* There's a z-index issue with this placement and the caption -- I think bumping the z-index of the whole caption up to 3 or 4 [here](https://github.com/Automattic/newspack-blocks/blob/trunk/src/blocks/homepage-articles/view.scss#L347) should fix (right now it's visible, but you can't click on links in it).
* Links in the credit are also not picking up the correct colour (at least with the default theme/colours). Styling any links in the caption to be inherited or maybe white should fix!

image

* This is kind of an edge case, but when I first added captions to the Homepage Posts blocks, I styled them to only display as one line and then trail off with an ellipsis when the featured image was set to 'behind', which I'm sure made sense at the time 😅 Unfortunately, it ends up cutting off the credit if it's really long, or if it's also shown with a caption. I think it's something we could iterate on after this PR; I can't think of any sites that actually use this placement _and_ a caption, to me the two seem to be at odds with each other. In the near-term if it IS an issue, using only a credit should fix it in most cases, and we could tweak on specific sites with CSS.

image


**Post Carousel**
* When captions are enabled but a post doesn't have one, the div around it is still output on the front-end; it has a top margin, which adds a little extra space than normal on the front-end. This doesn't happen in the editor: 
image

Otherwise, things look good! I tested the different visible post counts with the Post Carousel, and different font sizes and image placements for the Homepage Posts block, and didn't spot anything else!

thomasguillot:
I pushed some small changes: https://github.com/Automattic/newspack-blocks/pull/1687/commits/4f856ce6f93e4521e22fa5f629b7e2d2f7187b34

What it does is it uses the css variables available in:

1. Block Theme, if not,
2. "Classic" Theme, if not,
3. default value

dkoo:
> This is kind of an edge case, but when I first added captions to the Homepage Posts blocks, I styled them to only display as one line and then trail off with an ellipsis when the featured image was set to 'behind', which I'm sure made sense at the time 😅 Unfortunately, it ends up cutting off the credit if it's really long, or if it's also shown with a caption. I think it's something we could iterate on after this PR; I can't think of any sites that actually use this placement and a caption, to me the two seem to be at odds with each other. In the near-term if it IS an issue, using only a credit should fix it in most cases, and we could tweak on specific sites with CSS.

@laurelfulford agreed, this looks like an edge case that we shouldn't spend a lot of time accommodating. cca9d3b does add a `has-credit` class name to the `figcaption` if we want to do more.

> When captions are enabled but a post doesn't have one, the div around it is still output on the front-end; it has a top margin, which adds a little extra space than normal on the front-end. This doesn't happen in the editor:

Whoops, good. catch! The new function was still outputting empty HTML elements even if there's no caption or credit to display. cca9d3b fixes this.

src/blocks/homepage-articles/view.scss
```diff
@@ -344,7 +344,7 @@
 			right: 0;
 			text-align: right;
 			text-overflow: ellipsis;
-			z-index: 2;
+			z-index: 3;
```
laurelfulford, line 347:
```suggestion
			z-index: 3;
			
			a,
			a:visited {
			    color: #fff;
			}
```

Just need to add a little tweak for links in the caption to make sure they're not picking up the theme colour -- I tested this and it should work for regular & visited links in credits!

(I only think this is an issue because the theme's link styles are pretty specific!)

laurelfulford (changes requested):
Thanks for the fixes, @dkoo!

There's just one tweak remaining - once that's in place I think this is good to go!

src/blocks/homepage-articles/view.scss
dkoo: Thanks—totally forgot to test the different link states!

laurelfulford (approved):
Thanks @dkoo -- looks good! 🚢

PR #1712, opened 2024-04-19
fix: fix wrapping issue happening in Firefox

laurelfulford:
### All Submissions:

This PR fixes a odd pixel rounding issue that's happening in FF 125.0.1. It's not an elegant fix by any means, but hopefully works as a stopgap until we find something better.

1. Make sure you're running Firefox 125.0.1.
2. Add a columns block to the editor, and add a series of grid-layout homepage post blocks inside - 2 across, 3-across, 4-across, etc.
3. Adjust the width of the column up and down by 0.1% increments.
4. Note that most of the blocks (all but the three across) will wrap one article to the next line at random points. You can also publish the post, and do the same on the front-end using the element inspector (unfortunately when it wraps in the editor, it doesn't always wrap on the front end, and vis versa.
6. Repeat step 4 in the editor & on the front-end and confirm the blocks no longer wrap. 

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/view.scss
+++ b/src/blocks/homepage-articles/view.scss
@@ -81,24 +81,24 @@
 		// width of column - ( column gap / number of columns * number of gaps )
 		&.columns-3 article,
 		&.columns-6 article {
-			flex-basis: calc( 33.33% - ( var( --wpnbha-col-gap ) / 3 * 2 ) );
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Couldn't replicate the issue as described in Firefox v125.01, but also not seeing any negative effects from the changes, so approving!

laurelfulford:
Thanks @dkoo! 🙌 Merging just in case, and I'll make myself a task to follow up on this so I can figure out what happened.

PR #1719, opened 2024-04-25
feat: additional component updates

laurelfulford:
### All Submissions:

Updates the CSS classes used in Newspack Blocks, related to these changes: https://github.com/Automattic/newspack-plugin/pull/3091

Specifically:

* Replaced the `newspack-ui__inline-info` class with `newspack-ui__helper-text`, to line up with the Figma naming convention.
* Replaced the `newspack-ui__helper-text` class on the address with `newspack-ui__font--xs` to get the font size without picking up the lighter grey colour.
* Replaced the `newspack-ui__input-list` classname with `newspack-ui__input-card`.
* Edited to add: updated the newsletter coda styles.

1. See https://github.com/Automattic/newspack-plugin/pull/3091 for testing steps and test with it.

### Other information:

### DIFF
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -853,23 +853,25 @@ final class Modal_Checkout {
 		}
 		?>
 			<div class="newspack-modal-newsletters">
-				<h4><?php echo esc_html( self::get_modal_checkout_labels( 'newsletter_title' ) ); ?></h4>
+				<h3><?php echo esc_html( self::get_modal_checkout_labels( 'newsletter_title' ) ); ?></h3>
 				<div class="newspack-modal-newsletters__info">
+					<p>
 					<?php
 					echo esc_html( self::get_modal_checkout_labels( 'newsletter_details' ) );
 					?>
-					<br>
-					<span>
-					<?php
-						echo esc_html(
-							sprintf(
-								// Translators: %s is the user's email address.
-								__( 'Sending to: %s', 'newspack-blocks' ),
-								$email_address
-							)
-						);
-					?>
-					</span>
+						<br>
+						<span class="newspack-ui__color-text-gray">
+						<?php
+							echo esc_html(
+								sprintf(
+									// Translators: %s is the user's email address.
+									__( 'Sending to: %s', 'newspack-blocks' ),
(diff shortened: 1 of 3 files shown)


laurelfulford:
Just adding a note that the corresponding [Newspack Plugin PR got a thumbs up from design](https://github.com/Automattic/newspack-plugin/pull/3091#issuecomment-2111970780), so this one is good to go on that front, too!

chickenn00dle (changes requested):
The changes look good, just noting that I think we have to account for the case where the blocks plugin is being used without the main newspack plugin. 

I've been creating some basic backup styles for this case and using the `newspack-blocks` prefix for this and have a helper method to decide which to use based on whether newspack ui is available or not: https://github.com/Automattic/newspack-blocks/blob/101c2c23c2a0acdb7d2005028fe32d96c9ce68f3/includes/class-modal-checkout.php#L1238-L1240

laurelfulford:
Oooo, thanks @chickenn00dle! I totally dropped the ball on updating those -- I'll do a sweep and make updates for when the Newspack Plugin isn't installed!

laurelfulford:
@chickenn00dle I did some comparisons with and without the Newspack Plugin -- things are very not 1:1, but I'm wondering if it'd make more sense to tackle separately. My thinking is:

* It'd be nice to figure out _how_ close this should be -- I know we talked about making sure it looked okay, but I'm not sure how far to bring that. Making it 1:1 will add some maintenance overhead we probably don't want. For example I noticed the regular Woo styles are bleeding in (like the checkout is two columns in the modal with Newspack Plugin isn't installed). Maybe we lean into that? 
* This probably isn't a blocker for phase one, since anyone using it on our end of things will have both plugins installed.
 
I spun up an issue for this in Asana here, and assigned it to 'Phase 2' for now: 1205234045751551-as-1206899964006422. Let me know what you think!

chickenn00dle:
Sounds good @laurelfulford! In that case, I'll give this a thumbs up 👍

chickenn00dle (approved):

laurelfulford:
Awesome -- thanks @chickenn00dle!

PR #1720, opened 2024-04-25
fix: add `woocommerce_after_order_notes` action for integrations that need it

dkoo:
### All Submissions:

Currently our custom checkout form doesn't include Woo's shipping form template, which is where the `woocommerce_after_order_notes` action lives. This PR adds it right before the `woocommerce_checkout_after_customer_details` action, which is roughly where it should appear in relation to the checkout fields. Only registers the action hook if it hasn't already been registered.

1. Check out this branch.
2. Create a custom action callback to execute on the `woocommerce_after_order_notes` hook, e.g.

```php
add_action(
	'woocommerce_after_order_notes',
	function( $checkout ) {
		error_log( 'woocommerce_after_order_notes action!!!' );
		error_log( print_r( $checkout, true ) );
	}
);
```

3. Start a purchase of a product via the Checkout Button block. On the second screen (with payment details) confirm that your action callback is executed and that the `$checkout` argument is an instance of `WC_Checkout`.
4. Go to your `/shop` page and start a purchase via the standard Woo checkout flow. Confirm that your action callback is executed here, too.

### Other information:

---
  - https://app.asana.com/0/0/1207083368446933

### DIFF
--- a/src/modal-checkout/templates/checkout-form.php
+++ b/src/modal-checkout/templates/checkout-form.php
@@ -161,6 +161,11 @@ if ( ! $checkout->is_registration_enabled() && $checkout->is_registration_requir
 				</div>
 			<?php endif; ?>
 
+			<?php
+			if ( ! has_action( 'woocommerce_after_order_notes' ) ) {
+				do_action( 'woocommerce_after_order_notes', $checkout );
+			}
+			?>
 			<div class="after-customer-details">
 				<?php do_action( 'woocommerce_checkout_after_customer_details' ); ?>
 			</div>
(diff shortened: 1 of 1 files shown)


src/modal-checkout/templates/checkout-form.php
```diff
@@ -161,6 +161,11 @@
 				</div>
 			<?php endif; ?>
 
+			<?php
+			if ( ! has_action( 'woocommerce_after_order_notes' ) ) {
```
adekbadek: In my testing, `has_action( 'woocommerce_after_order_notes' ) === true` (so the following lines don't get executed), _and_ the callback is never called.

adekbadek (changes requested):

src/modal-checkout/templates/checkout-form.php
dkoo: Whoops, `has_action` is the wrong method for checking whether an action hook has been registered. (`has_action` checks whether a _callback_ has been registered for the given action hook.) 9604ebb checks whether the action hook is registered on the global `$wp_actions` object instead.

adekbadek (approved):

PR #1726, opened 2024-05-07
fix(ras-acc): re-adds missing newsletters form styles

chickenn00dle:
### All Submissions:

This just re-adds some newsletters form styles that seem to have been removed from the checkout block.

Before:
Screenshot 2024-05-07 at 16 24 23

After:
Screenshot 2024-05-07 at 16 11 00

1. Enable the post checkout newsletter signup flag in `wp-config.php`: `define( 'NEWSPACK_ENABLE_POST_CHECKOUT_NEWSLETTER_SIGNUP', true );`
2. Logged in as a reader, go through the checkout flow
3. On the thank you page, on `trunk` you should see a newsletters signup form with no styles. On this branch you should see the styled form as pictured above.

### Other information:

### DIFF
--- a/src/modal-checkout/checkout.scss
+++ b/src/modal-checkout/checkout.scss
@@ -1,3 +1,5 @@
+@use '../shared/sass/colors';
+
 /* stylelint-disable-next-line */
 #newspack_modal_checkout_container { /* stylelint-disable-line */
 	padding: 24px;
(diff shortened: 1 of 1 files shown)


laurelfulford:
@chickenn00dle I think this is something I missed! We should have styles for this in the components (or at least getting refined [in this PR](https://github.com/Automattic/newspack-plugin/pull/3091)).

There's [a block PR that goes with that plugin update](https://github.com/Automattic/newspack-blocks/pull/1719); I can take a swing at applying these styles there, I think we should be able to avoid some duplicate CSS :) Let me know what you think (in case I'm missing something!)! 

----

Edited to add: I had to tweak the button styles a bit to be more specific, but otherwise, this is just with CSS classes: 
image

If this seems OK I can push it up to the blocks PR I linked.

Based on the Figma file I think the success state and newsletters may need to be split into two screens, but that could be a V2!

Auth Modal
Auth Modal - newsletter

chickenn00dle:
> There's https://github.com/Automattic/newspack-blocks/pull/1719; I can take a swing at applying these styles there, I think we should be able to avoid some duplicate CSS :) Let me know what you think (in case I'm missing something!)!

Oh nice! Great. I think I will leave this one to you then and close this PR 👍 

> Based on the Figma file I think the success state and newsletters may need to be split into two screens, but that could be a V2!

Good catch. I wasn't sure where we stood with the newsletter modal post-checkout. Based on conversations around this, the V1 version had already been built out so I was assuming what we had hidden behind the `NEWSPACK_ENABLE_POST_CHECKOUT_NEWSLETTER_SIGNUP` flag was what we were going with for V1. But if not, I am happy to add a new task to the project to get this split up.

laurelfulford:
Thanks @chickenn00dle! I've added the classes to [this block PR](https://github.com/Automattic/newspack-blocks/pull/1719) and included it in [the corresponding Newspack Plugin's PR](https://github.com/Automattic/newspack-plugin/pull/3091).

I had a couple questions around the variables styles for design, so I can ask about the newsletter coda stuff with that, too!

PR #1728, opened 2024-05-09
fix: respect subscriptions limiter, and handle cart errors in modal

dkoo:
### All Submissions:

Respects the "Limit subscription" option in Woo products. Also implements some styles so that the modal checkout window is still usable if the `add_to_cart` method fails and redirects to the cart page inside the modal.

Closes `1200550061930446/1207095202849399`.

1. Ensure you have two subscription products published: a simple subscription and a variable one (see the [Woo documentation](https://href.li/?https://woocommerce.com/document/subscriptions/store-manager-guide/#creating-a-subscription-product) for details on how to set up).
2. For each product, in Product data > Advanced, set the "Limit subscription" option to "Limit to one active subscription".

Screenshot 2024-05-09 at 4 53 19 PM

3. Publish a page or post with Checkout Button blocks for each product.
4. As a reader, purchase a subscription for each product using the Checkout Button blocks.
5. After purchasing, while still logged in as the reader, attempt to purchase each product again.
6. On `trunk`, observe that the modal checkout window displays the full Cart page with site header + footer, locked behind a loading animation that never resolves.

Screenshot 2024-05-09 at 5 02 15 PM

7. Check out this branch. Repeat step 6 and confirm that you now see a more graceful error message in the modal, along with a button that closes the modal when clicked.

Screenshot 2024-05-09 at 4 50 33 PM

8. As an admin, cancel the reader's subscriptions for both products.
9. Repeat step 7 and confirm that you can now repurchase those products.
10. As an admin, for each product, change the "Limit subscription" option to "Limit to one of any status". 
11. Repeat step 7 and confirm that you still see the same error message as before.

### Other information:

---
  - https://app.asana.com/0/0/1207095202849399

### DIFF
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -72,6 +72,15 @@ final class Modal_Checkout {
 		add_filter( 'googlesitekit_adsense_tag_blocked', [ __CLASS__, 'is_modal_checkout' ] );
 		add_filter( 'googlesitekit_tagmanager_tag_blocked', [ __CLASS__, 'is_modal_checkout' ] );
 		add_filter( 'jetpack_active_modules', [ __CLASS__, 'jetpack_active_modules' ] );
+
+		// Ensure that options to limit the number of subscriptions per product are respected.
+		if ( self::is_modal_checkout() && class_exists( 'WCS_Limiter' ) ) {
+			add_filter( 'woocommerce_subscription_is_purchasable', [ 'WCS_Limiter', 'is_purchasable_switch' ], 12, 2 );
+			add_filter( 'woocommerce_subscription_variation_is_purchasable', [ 'WCS_Limiter', 'is_purchasable_switch' ], 12, 2 );
+			add_filter( 'woocommerce_subscription_is_purchasable', [ 'WCS_Limiter', 'is_purchasable_renewal' ], 12, 2 );
+			add_filter( 'woocommerce_subscription_variation_is_purchasable', [ 'WCS_Limiter', 'is_purchasable_renewal' ], 12, 2 );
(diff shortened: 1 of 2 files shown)


includes/class-modal-checkout.php
```diff
@@ -72,6 +74,15 @@ public static function init() {
...
 		add_filter( 'googlesitekit_tagmanager_tag_blocked', [ __CLASS__, 'is_modal_checkout' ] );
 		add_filter( 'jetpack_active_modules', [ __CLASS__, 'jetpack_active_modules' ] );
+
+		// Ensure that options to limit the number of subscriptions per product are respected.
+		if ( self::is_modal_checkout() && class_exists( 'WCS_Limiter' ) ) {
+			add_filter( 'woocommerce_subscription_is_purchasable', [ 'WCS_Limiter', 'is_purchasable_switch' ], 12, 2 );
+			add_filter( 'woocommerce_subscription_variation_is_purchasable', [ 'WCS_Limiter', 'is_purchasable_switch' ], 12, 2 );
+			add_filter( 'woocommerce_subscription_is_purchasable', [ 'WCS_Limiter', 'is_purchasable_renewal' ], 12, 2 );
+			add_filter( 'woocommerce_subscription_variation_is_purchasable', [ 'WCS_Limiter', 'is_purchasable_renewal' ], 12, 2 );
+			add_filter( 'woocommerce_valid_order_statuses_for_order_again', [ 'WCS_Limiter', 'filter_order_again_statuses_for_limited_subscriptions' ] );
```
adekbadek, line 90: Nit: this looks like calling code which should be called by WC. So I chcecked and it is, but only in the regular checkout and API – that's why we need it here. Might be worth to drop a comment here about this.

adekbadek (approved):

includes/class-modal-checkout.php
dkoo: Good idea! Added in 2457dd6

PR #1730, opened 2024-05-10
feat(ras-acc): add name your price support to checkout button

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1206943664367857/1206041522383357/f

This PR adds support for name your price to the checkout button block:

Screenshot 2024-05-14 at 18 04 22

Note: Be sure the to checkout `epic/ras-acc` in the main newspack plugin before testing.

1. Ensure Name Your Price is installed and active
2. Create a simple product, a subscription product, and a variable product, all with name your price set, as well as any type of product with NYP NOT setup.
  - For at least one product set a maximum price 
3. Add a checkout button block for each of these products to any page
  - For at least one button set a custom price that is above the minimum price via the name your price block settings
4. Logged in as a reader, select the simple product button
5. Go through the edit billing details section (first step) of checkout and proceed to the payment section (second step)
6. Confirm you see a **Increase your support** section with a form to enter a new price for the product (pictured above)
7. Enter a price that is below the minimum price and submit. Confirm an error appears asking you to input an amount greater than the base price
8. If this is the product with a custom price set via block settings, enter a price that is below the set price listed at the top of the checkout modal, but above the minimum price and submit. Confirm an error appears asking you to input an amount greater than the base price
9. If this is the product with a maximum price set product settings, enter a price that is above the max price and submit. Confirm an error appears asking you to input a price that is less than the max price.
10. Enter a valid price and submit. Confirm the price updates at the top of the checkout modal.
11. Complete checkout and confirm the correct price was charged in WP-Admin WooCommerce Orders menu as well as in the customer receipt sent by Woo.
12. Repeat the above steps for the variable and subscription products and confirm the NYP form is present and functions as expected.
13. Repeat the above steps for the product that does not have NYP active. Confirm the NYP form does NOT appear during the checkout flow.
14. On any page, add a donation block
15. As a reader, make a donation via the donation block. Confirm the **Increase your support** field is not present during the checkout field

### Other information:

### DIFF
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -1391,6 +1391,8 @@ final class Modal_Checkout {
 					__( 'Thank you for supporting %s. Your transaction was successful.', 'newspack-blocks' ),
 					get_option( 'blogname' )
 				),
+				'checkout_nyp'               => __( "Your contribution directly funds our work. If you're moved to do so, you can opt to pay more than the standard rate.", 'newspack-blocks' ),
+				'checkout_nyp_thankyou'      => __( "Thank you for your generosity! We couldn't do this without you!", 'newspack-blocks' ),
 			];
 
 			/**
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -1286,7 +1286,7 @@ final class Modal_Checkout {
 						<p class="input-price" >
 							<label for="price">
 								<span class="currency"><?php echo esc_html( $currency_symbol ); ?></span>
-								<input type="number" step=".01" name="price" placeholder="<?php echo esc_attr( $price ); ?>" />
+								<input type="number" min="0" step=".01" name="price" placeholder="<?php echo esc_attr( $price ); ?>" />
 							</label>
 							<button type="submit" class="<?php echo esc_attr( "{$class_prefix}__button {$class_prefix}__button--outline" ); ?>">
 								<?php echo esc_html( self::get_modal_checkout_labels( 'checkout_nyp_apply' ) ); ?>
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -1264,17 +1264,17 @@ final class Modal_Checkout {
 					$_product = apply_filters( 'woocommerce_cart_item_product', $cart_item['data'], $cart_item, $cart_item_key );
 					if ( $_product && $_product->exists() && $cart_item['quantity'] > 0 && \WC_Name_Your_Price_Helpers::is_nyp( $_product ) && apply_filters( 'woocommerce_checkout_cart_item_visible', true, $cart_item, $cart_item_key ) ) :
 						?>
-						<h3><?php echo esc_html( self::get_modal_checkout_labels( 'checkout_nyp_title' ) ); ?></h3>
 						<form class="modal_checkout_nyp">
+							<h3><?php echo esc_html( self::get_modal_checkout_labels( 'checkout_nyp_title' ) ); ?></h3>
 							<input type="hidden" name="newspack_checkout_name_your_price" value="1" />
 							<input type="hidden" name="product_id" value="<?php echo esc_attr( $_product->get_id() ); ?>" />
-							<div class="modal_checkout_nyp_input">
+							<p class="input-price" >
 								<label for="price">
 									<span><?php echo esc_html( \get_woocommerce_currency_symbol() ); ?></span>
-									<input type="number" step="any" min="<?php echo esc_attr( \WC_Name_Your_Price_Helpers::get_minimum_price( $_product->get_id() ) ); ?>" name="price" placeholder="<?php echo esc_attr( \WC_Name_Your_Price_Helpers::get_suggested_price( $_product->get_id() ) ); ?>" />
+									<input type="number" step="any" min="<?php echo esc_attr( \WC_Name_Your_Price_Helpers::get_minimum_price( $_product->get_id() ) ); ?>" class="input-text" name="price" placeholder="<?php echo esc_attr( \WC_Name_Your_Price_Helpers::get_suggested_price( $_product->get_id() ) ); ?>" />
 								</label>
 								<button type="submit" class="<?php echo esc_attr( "{$class_prefix}__button {$class_prefix}__button--outline" ); ?>"><?php echo esc_html( self::get_modal_checkout_labels( 'checkout_nyp_apply' ) ); ?></button>
-							</div>
+							</p>
 							<p class="result <?php echo esc_attr( "{$class_prefix}__inline-info" ); ?>"><?php echo esc_attr( self::get_modal_checkout_labels( 'checkout_nyp' ) ); ?></p>
 						</form>
 						<?php
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -57,6 +57,7 @@ final class Modal_Checkout {
 		add_filter( 'woocommerce_order_button_text', [ __CLASS__, 'order_button_text' ], 5 );
 		add_filter( 'option_woocommerce_subscriptions_order_button_text', [ __CLASS__, 'order_button_text' ], 5 );
 		add_action( 'woocommerce_before_checkout_form', [ __CLASS__, 'render_before_checkout_form' ] );
+		add_action( 'woocommerce_before_checkout_form', [ __CLASS__, 'render_name_your_price_form' ], 11 );
 		add_action( 'woocommerce_checkout_before_customer_details', [ __CLASS__, 'render_before_customer_details' ] );
 		add_filter( 'woocommerce_enable_order_notes_field', [ __CLASS__, 'enable_order_notes_field' ] );
 		add_action( 'woocommerce_checkout_process', [ __CLASS__, 'wcsg_apply_gift_subscription' ] );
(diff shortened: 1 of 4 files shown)


chickenn00dle:
~~This is pretty much ready for review. Gonna wait to add the `Needs Review` label though so I can get a test site up with the changes for design review first. I'll try and get to this tomorrow.~~

I've set up a test site, which can be found in the Asana task

kevinzweerink:
Looking great, thanks Raz!

A couple notes:
1. The number input seems to pick up scroll events — can we disable that?
2. The left padding on the input is too narrow — I kind of see the case for this, since the currency sign isn't editable, but it feels a bit awkward. Can we make the left padding before the currency symbol match the left padding on other inputs?
3. The margins around the input are incorrect — should be 8px either side per Thomas's [Components](https://www.figma.com/design/NvJXxEicyiw7FWqv0JtjvT/Components?node-id=1099%3A333&t=Hueb087wlEmzg4to-1) work. Not sure if this is part of the scope of this PR or if this is an issue that should be resolved at the component level, though it does look like most of the other inputs in the flow have correct margins.

I also found some stuff that I think is outside the scope of this specific PR but worth noting for RAS-ACC generally:
1. The page behind the modal is scrollable while the modal is visible, this shouldn't be the case.
2. Transaction Breakdown table isn't appearing, which makes it hard to know whether you've applied your new price or not
3. Complete Transaction button should show total (Complete Transaction: $19.99).
4. The component that shows the product for the checkout flow doesn't appear on the sign in/create account steps.
5. The continuity of the flow breaks if you switch from "Sign in" to "Create an account" *via the link in the "no account" error message from the sign in screen* — you just make an account and then you get the success screen, no checkout.
6. The typeahead for the State selector seems to use default WordPress blue — can we change that to use the dark gray that our buttons use?

Let me know if you have any questions about any of these!

chickenn00dle:
Thanks for the review @kevinzweerink!

> The number input seems to pick up scroll events — can we disable that?

Hmm. I'm having a hard time reproducing this. I've tried on Safari, Chrome, and FF and on all three scrolling simply applies to the entire modal and not the number input. Can you tell me how to reproduce?

> The left padding on the input is too narrow — I kind of see the case for this, since the currency sign isn't editable, but it feels a bit awkward. Can we make the left padding before the currency symbol match the left padding on other inputs?

Fixed! I've updated the test site with the new padding as well (you may need to clear browser cache to see the change). Just in case here is a pic of the new padding:
Screenshot 2024-05-15 at 16 02 43

> The margins around the input are incorrect — should be 8px either side per Thomas's [Components](https://www.figma.com/design/NvJXxEicyiw7FWqv0JtjvT/Components?node-id=1099%3A333&t=Hueb087wlEmzg4to-1) work. Not sure if this is part of the scope of this PR or if this is an issue that should be resolved at the component level, though it does look like most of the other inputs in the flow have correct margins.

Ah. The component style updates are being addressed in another PR by @laurelfulford. So I think we can ignore this one in this PR if that's okay.

> I also found some stuff that I think is outside the scope of this specific PR but worth noting for RAS-ACC generally

Thanks! I've created a QA task for these points. Feel free to add more detail there if you'd like: https://app.asana.com/0/1205234045751551/1207328906416559/f

kevinzweerink:
Ok cool, thanks for checking on this stuff — 

> I've tried on Safari, Chrome, and FF and on all three scrolling simply applies to the entire modal and not the number input. Can you tell me how to reproduce?

Sure! For me, once I have a value in the input, if I put the cursor over the input element and scroll, it changes the value. I captured a video in Arc (which is Chrome), but I also confirmed it works in standard chrome.

https://github.com/Automattic/newspack-blocks/assets/3193536/26ddbf27-9e1c-429d-ab6b-9f218562ba48

Unfortunately, when testing in Chrome, I uncovered a surprise … button? Any idea why this is appearing? I'm sure it would be nice to support G Pay at some point, but if we can exclude it from the initial scope of this work I think that would help us get this out the door faster.

CleanShot 2024-05-15 at 16 48 29@2x

chickenn00dle:
> For me, once I have a value in the input, if I put the cursor over the input element and scroll, it changes the value. I captured a video in Arc (which is Chrome), but I also confirmed it works in standard chrome.

Strange. I still can't reproduce. I wonder if this has something to do with my mouse settings and it's fancy scroll features. In any case, I pushed up a change that should handle scroll events for the input. Let me know if that does the trick for you.

> Unfortunately, when testing in Chrome, I uncovered a surprise … button? Any idea why this is appearing? I'm sure it would be nice to support G Pay at some point, but if we can exclude it from the initial scope of this work I think that would help us get this out the door faster.

Oops. This is an express checkout option added by the Stripe plugin and is unrelated to RAS-ACC. I've disabled it on the test site so it shouldn't appear anymore.

Let me know if you still have any issues @kevinzweerink!

kevinzweerink:
The scroll thing seems resolved, but it looks like something else happened and maybe some regressions were introduced.

CleanShot 2024-05-16 at 15 43 39@2x

chickenn00dle:
> The scroll thing seems resolved, but it looks like something else happened and maybe some regressions were introduced.

Ah. The main plugin was updated with new styles for the form. I did add some changes to account for this though, so it should be fixed. You may need to clear your browsers cache again.

chickenn00dle:
Oops. Meant to edit my comment but accidentally deleted :( 

Here it is again.

> The scroll thing seems resolved, but it looks like something else happened and maybe some regressions were introduced.

Ah. The form styles were updated in the main plugin, but I added some changes to account for this in my last update. You might need to clear your browsers cache. This is what it looks like on the test site on my end:

Screenshot 2024-05-16 at 16 39 11

kevinzweerink:
Ah gotcha. It's looking super close on my end! The only weird thing I'm seeing at this point is a microscopic layout shift while the validation logic runs — any idea what's up with this?

https://github.com/Automattic/newspack-blocks/assets/3193536/eb24aa4b-ef18-44a7-b594-eab87d13bd7b

chickenn00dle:
> The only weird thing I'm seeing at this point is a microscopic layout shift while the validation logic runs — any idea what's up with this?

It looks like this is being caused by the jquery block method which adds an overlay on top of the form while its processing. I went ahead and removed it here and re-implemented the loading state with plain CSS/disabling the input: https://github.com/Automattic/newspack-blocks/pull/1730/commits/b0a8c9891c68257e4b77c0d7d1691273bb554af5

Test site is updated. Let me know if this works. Be sure to clear browser cache!

kevinzweerink:
Looks good to me! Thanks for digging into it! All set from the design side (with the caveat that there are some broader RAS-ACC things still at large).

chickenn00dle:
Great! Thanks @kevinzweerink 🙌 

Moving this to review from the @Automattic/newspack-product team.

laurelfulford (commented):
This functions really well in my testing, @chickenn00dle! I focused on just doing some functional testing so the code can get a second set of eyes.

I did run into a weird edge case when I mixed NYP with variable products:

image

If it's easy to change the markup a bit to wrap the 'Suggested price:' text in its own HTML element, I think it could be fixed by matching the 'Sale' styles (absolutely positioning the text before the price so it doesn't mess with the grid, and making the price the same size as the other prices), but @kevinzweerink might have some better ideas to make it look like it fits!

(The button spacing in that screenshot should be fixed when [this PR lands](https://github.com/Automattic/newspack-plugin/pull/3130)!)

I also have a non-blocking question: we have some of the variation styles in the Newspack Plugin, along with a bunch of WC overrides there -- would it make sense to combine them in one place (either here or there?). I'm not sure any will be needed for My Account so maybe it makes sense to have the styles for the block-based markup coming from the block, and just have fallbacks for the variables? 

(Absolutely something we can dig into once things are less in flux!)

chickenn00dle:
Thanks for having a look @laurelfulford!

> If it's easy to change the markup a bit to wrap the 'Suggested price:' text in its own HTML element, I think it could be fixed by matching the 'Sale' styles (absolutely positioning the text before the price so it doesn't mess with the grid, and making the price the same size as the other prices)

Good catch! I'll wait to hear from @kevinzweerink before making changes here. Maybe I can address this in a follow-up PR so it doesn't hold this one up?

> we have some of the variation styles in the Newspack Plugin, along with a bunch of WC overrides there -- would it make sense to combine them in one place (either here or there?). I'm not sure any will be needed for My Account so maybe it makes sense to have the styles for the block-based markup coming from the block, and just have fallbacks for the variables?

Good question! My thought is we should keep any blocks specific styles/overrides within the blocks plugin so it's easier to track these changes down in the future. But I'm happy to go with another approach if you feel it would be better to consolidate things within the Newspack Plugin instead.

laurelfulford:
> Maybe I can address this in a follow-up PR so it doesn't hold this one up?

I'd be okay with this! Digging around a bit, I ran across this section of the RAS-ACC Figma file, so there might be more to how this should work in specific cases that would be good to get confirmation on:

image

> Good question! My thought is we should keep any blocks specific styles/overrides within the blocks plugin so it's easier to track these changes down in the future

I think having these styles in the blocks makes the most sense, too 😄 The two pieces I'm thinking of are [a chunk of styles in the Newspack plugin that deal solely with creating the grid layout for the variations](https://github.com/Automattic/newspack-plugin/blob/epic/ras-acc/assets/newspack-ui/scss/_modals.scss#L127-L215), and [a whole file](https://github.com/Automattic/newspack-plugin/blob/epic/ras-acc/assets/newspack-ui/scss/elements/woocommerce/_overrides.scss) for WooCommerce overrides. For the latter we might need to verify that none of it is needed for screens like My Account, but the former is probably safe to move outright. I can make myself a task for this.

In the meantime, I can make a task for that and assign to me -- would you like me to spin one up for NYP + variations? I can mark this PR as approved for incorporating NYP in the checkout bit in the meantime 🙂

laurelfulford (approved):

chickenn00dle:
Thanks @laurelfulford!

> would you like me to spin one up for NYP + variations? I can mark this PR as approved for incorporating NYP in the checkout bit in the meantime 🙂

~~I'll get a task added for that now 👍~~

Created a task for this here: 1205234045751551-as-1207428738079159/f

kevinzweerink:
Just wanted to follow up on the screenshot you shared @laurelfulford!

That style shouldn't be implemented — I was experimenting with it, but ended up deciding that it makes more sense to consolidate all of the NYP functionality at the checkout step to make things simpler and more stylistically consistent, at least for now. Sorry for the confusion, I've removed that exploration from the Figma file.

The style for NYP options in a variable checkout button should match what's shown in [this asana task](https://app.asana.com/0/1205234045751551/1207428738079159/f).

PR #1734, opened 2024-05-15
feat(donate): add new "Modern" style variation

thomasguillot:
### All Submissions:

This PR adds a new style variation to the Donate block. Its design is based on the new UI components we are currently building (thanks @laurelfulford!).

It's using newspack ui variables (with fallback) and forces the font (family, sizes, weight) to system.

Going forward, this is the style publishers should really be using and it will be extremely valuable to have this in our patterns.

There's also an "hidden" class for the Tiers layout. If a publisher adds `list` to the css classes (`Advanced > ADDITIONAL CSS CLASS(ES)`) and they are using a Tiers layout then the options will switch to a list. This is quite useful if they want to display this block in a narrow column/sidebar for example.

example

1. Switch to this branch
2. Add multiple donate blocks to a page
3. Switch them to "Modern" style
4. Play with the various settings (Tiered, Untiered, Frequency/Tiers layout, etc...)
5. Add a custom css class of `list` to a block using the Tiers layout

### Other information:


dkoo (approved):
This is looking nice and I don't see any conflicts with existing blocks/styles, so approving. Just a couple of non-blocking questions:

> Going forward, this is the style publishers should really be using

Should we consider making this the first and default style for new instances of the block, then? Or does it make more sense to wait until RAS-ACC gets merged first?

> There's also an "hidden" class for the Tiers layout.

Why make it "hidden"? Would it be useful/desirable to either make it a separate style in its own right, or to borrow a UI pattern from the Homepage Posts block and use this inline toolbar to let publishers switch the style?

Screenshot 2024-05-16 at 4 42 30 PM

thomasguillot:
> Why make it "hidden"? Would it be useful/desirable to either make it a separate style in its own right, or to borrow a UI pattern from the Homepage Posts block and use this inline toolbar to let publishers switch the style?

Oh, I like the idea of using the toolbar. I can merge this PR for now and then look into adding the buttons to the toolbar in the near future.

thomasguillot:
> Should we consider making this the first and default style for new instances of the block, then? Or does it make more sense to wait until RAS-ACC gets merged first?

Good idea! Maybe let's wait until RAS-ACC gets merged 1st.

PR #1735, opened 2024-05-16
fix(ras-acc): make sure checkout modal fires when logged in

laurelfulford:
### All Submissions:

This PR fixes a small issue I noticed when working on the components, but didn't really investigate: when you're logged in as a WordPress User (and not a Reader), the checkout modal didn't open. The variable modal did open, but then when you picked a variation, it would close the modal and not move to the checkout.

My fix is just to check for the `logged-in` CSS class, but there's probably a more elegant way that ties into the existing code!

1. On the epic/ras-acc branch, set up a donate block and a checkout button block that loads a product with variations.
2. In a window when you're logged into WordPress, try to run the checkout from each block. Note that it doesn't work (either doesn't open a modal, or closes the variation modal when you pick a variation).
3. In an incognito window, repeat step 2 and note the modals do work (it will open the registration modal first before the checkout modal). 
5. Repeat steps 2 and 3 and confirm the checkout modal can be reached in both cases.

### Other information:

### DIFF
--- a/src/modal-checkout/modal.js
+++ b/src/modal-checkout/modal.js
@@ -1,5 +1,4 @@
-/* globals newspack_ras_config, newspackBlocksModal */
-window.newspack_ras_config = window.newspack_ras_config || {};
+/* globals newspackBlocksModal, newspack_ras_config */
 
 /**
  * Style dependencies
--- a/src/modal-checkout/modal.js
+++ b/src/modal-checkout/modal.js
@@ -1,4 +1,5 @@
-/* globals newspackBlocksModal */
+/* globals newspack_ras_config, newspackBlocksModal */
+window.newspack_ras_config = window.newspack_ras_config || {};
 
 /**
  * Style dependencies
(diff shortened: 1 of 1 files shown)


src/modal-checkout/modal.js
```diff
@@ -207,7 +208,8 @@ domReady( () => {
 
 					if (
 						window?.newspackReaderActivation?.openAuthModal &&
-						! window?.newspackReaderActivation?.getReader?.()?.authenticated
+						! window?.newspackReaderActivation?.getReader?.()?.authenticated &&
+						! loggedIn
```
adekbadek: There's the `newspack_ras_config` global – which is actually preventing the modal from opening.

https://github.com/Automattic/newspack-plugin/blob/ff256ee26865256d6817d7885fc7456e7f22d347/assets/reader-activation/index.js#L159-L160

Note that it should be included in the `globals` declaration at the top of the file.

adekbadek (changes requested):

src/modal-checkout/modal.js
laurelfulford: AH! Thanks @adekbadek! I combed around for something in the Blocks like that, I should've checked the plugin. 

I've updated the PR -- just let me know if there's anything that needs adjusting!

adekbadek (approved):
👌

PR #1740, opened 2024-05-20
feat: separate out decimal removal regex for wider use

laurelfulford:
### All Submissions:

This PR makes a couple changes to an existing function that filters the subscription price:

* It moves the regex that removes empty `.00` decimal values from longer prices to a more generic spot, so it's applied to products outside of subscriptions in the variable checkout.
* It tweaks the `$replace_pattern` value -- something about the `/\\` paired with a `.` as the decimal value caused it not to work, but other characters in that spot (like `,`) would be swapped out fine. It seemed to cause some kind of weird escaping issue, but I don't understand it -- echoing `$replace_pattern` would look like `/\\.00$/` when I used a period, but like `/\,00$/` with characters like a comma. I'm admittedly not super proficient in regex, so there may be a better fix than what I did -- any suggestions welcome! 

1. Set up a variable product and a variable subscription, with a variety of prices: make sure you have prices less than $100 with and without varied decimal values, and prices greater than $100 with varied decimal values.
2. Make sure your decimal separator is set to `.` under WooCommerce > Settings.
3. Set up a checkout block for each product.
4. Click the button for the variable subscription product and note all the prices are displayed in full.

image

5. Try changing the decimal separator to a different character, like a `,`.
6. Repeat step 4 and note that the decimal values are trimmed for prices greater than $100 that just have a `.00` (any other decimal value should be left alone) -- in this example, it's just the price in the top right corner:

image

8. Apply the PR.
9. Check both the variable subscription and variable product and confirm that trailing `00` values are being trimmed.
10. Change the decimal separator back to `.` and confirm it still works for both popups.

### Other information:


adekbadek (commented):
Trimming of the trailing zeroes works as described with both `.` and `,` used as the decimal separator. But why is it applied only to prices >100?

laurelfulford:
Thanks @adekbadek! 

> Trimming of the trailing zeroes works as described with both . and , used as the decimal separator. But why is it applied only to prices >100?

It's based on some notes in the RAS-ACC mockups for the Variable Subscription - I think it's to hit some kind of visual balance between the shorter prices and longer prices, and to keep the longer prices from taking up too much space. From the Figma file:

> If possible, it would be nice to cut “.00” off of prices that are 3-digits or larger. (Obviously we can’t do that if the publisher has specified a price with cents like $149.99)

adekbadek (approved):

PR #1755, opened 2024-05-31
feat(donate): add setting to change the style of tiered donations

thomasguillot:
### All Submissions:

_The Problem_

When using the Tiered layout, we end up with a 3-column layout. It's great when it's within the main content area but if a publisher wants to insert this block in a narrow column, it looks pretty bad...

Screenshot 2024-05-31 at 14 06 49@2x

_The Solution_

The idea is to be able to toggle between a Grid, and a List style.
This is something I originally introduced with the [Modern style variation](https://github.com/Automattic/newspack-blocks/pull/1734) as a custom CSS class and thought it would be better to just add this at the block level (props @dkoo!).

Screenshot 2024-05-31 at 14 08 00@2x

Instead of being a simple CSS class, this PR adds the buttons to the Toolbar, when the Tiers layout is selected. By default it's the Grid layout but now publishers are able to switch to a List if they want/need to.

Screenshot 2024-05-31 at 14 07 17@2x

1. On a page, add a bunch of Donate blocks, including some in narrow columns (e.g. 33/33/33)
2. Have them Tiered based
3. Check the front-end
4. Switch to this branch
5. Refresh editor and check if you see the Tier style buttons
6. Play with the buttons and see if it toggles the correct styles in the editor and front-end

### Other information:

### DIFF
--- a/src/blocks/donate/block.json
+++ b/src/blocks/donate/block.json
@@ -52,6 +52,10 @@
 			"type": "string",
 			"default": "frequency"
 		},
+		"tierStyle": {
+			"type": "string",
+			"default": "grid"
+		},
 		"useModalCheckout": {
 			"type": "boolean",
 			"default": true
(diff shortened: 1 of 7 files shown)


dkoo (approved):
Looks great!

PR #1760, opened 2024-06-03
feat(ras-acc): keyboard navigation improvements

laurelfulford:
### All Submissions:

This PR, with https://github.com/Automattic/newspack-plugin/pull/3164, adds some accessibility improvements to the reader auth and modal checkout. Both PRs should be tested together, and the testing steps can be found [here](https://github.com/Automattic/newspack-plugin/pull/3164).

### Other information:


src/modal-checkout/modal.js
```diff
@@ -27,6 +28,7 @@ domReady( () => {
 	const modalContent = modalCheckout.querySelector( `.${ MODAL_CLASS_PREFIX }__content` );
 	const modalCheckoutHiddenInput = createHiddenInput( 'modal_checkout', '1' );
 	const spinner = modalContent.querySelector( `.${ CLASS_PREFIX }__spinner` );
+	let modalTrigger = document.querySelector( '.newspack-reader__account-link' )[0];
```
chickenn00dle:
```suggestion
	let modalTrigger = document.querySelector( '.newspack-reader__account-link' )?.[0];
```

chickenn00dle (changes requested):
Left my main code review in the plugin PR, but also had a bit of feedback here.

For the `trapFocus` function, rather than just duplicating the code across the repos, I wonder if we can do something like assign it to browser's global object in plugin, and call it here that way?

laurelfulford:
> For the `trapFocus` function, rather than just duplicating the code across the repos, I wonder if we can do something like assign it to browser's global object in plugin, and call it here that way?

This is a good point! My concern was that the version here in the Newspack Blocks gets super specific in the `is ( iframe )` check, to work with elements that get hidden inside of the modal checkout, and iframes. The version in Newspack Plugin is more of a non-specific, reusable version of the code. 

The `iframe`-specific version is only used in one spot; the other two in Newspack Blocks would use the more generic version. Would it make sense to rework the `trapFocus` in the Newspack Block to cater to just the modal checkout, and to call the generic version from the Newspack Plugin? 

(Just let me know if I'm not making sense!)

src/modal-checkout/modal.js
laurelfulford: Thanks @chickenn00dle! This has been fixed in [fc3cee3](https://github.com/Automattic/newspack-blocks/pull/1760/commits/fc3cee36fcd5a9ce315c5c56016c95bb083d2dd4).

chickenn00dle (approved):
Thanks @laurelfulford!

> My concern was that the version here in the Newspack Blocks gets super specific in the is ( iframe ) check, to work with elements that get hidden inside of the modal checkout, and iframes. The version in Newspack Plugin is more of a non-specific, reusable version of the code.

Ah. I see what you mean. Makes sense.

> The iframe-specific version is only used in one spot; the other two in Newspack Blocks would use the more generic version. Would it make sense to rework the trapFocus in the Newspack Block to cater to just the modal checkout, and to call the generic version from the Newspack Plugin?

I'm also just remembering its possible for modal checkout to be on a site without the main newspack plugin present (although from what I understand this isn't really a thing?). Given this is a possibility, I think I take back my previous comment and am okay with the duplicate code here. Sorry for the confusion.

chickenn00dle (approved):

laurelfulford:
Thanks @chickenn00dle!

PR #1761, opened 2024-06-04
fix(ras-acc): add extra check for iframe for spinner visibility

laurelfulford:
### All Submissions:

This PR fixes an issue where the iframe load event is getting fired twice in Chrome and Safari -- one too early -- and it ended up hiding the spinner too soon so it was never visible in the modal checkout. 

I'm not sure of the issue the original `else` statement was fixing or how to recreate it -- [it was added](https://github.com/Automattic/newspack-blocks/commit/bbc9946df37f47972e1d653e9c7db70ae1298a6e) as part of [this PR](https://github.com/Automattic/newspack-blocks/pull/1602), so the commit doesn't go into too much depth. This seemed like a relatively safe approach -- if the iframe hasn't loaded anything, it's probably safe to keep showing the spinner -- but I'm definitely good with going another way! 

See 1205234045751551-1207426203801787

1. On the `epic/ras-acc` branch, set up a checkout button, and run a checkout in Chrome or Safari. Note when the modal is trying to load the checkout, it doesn't show a loader. Compare with Firefox, which will show a loader on the epic branch.

image

2. Apply this PR and run `npm run build`.
3. Confirm that the loader now shows in Chrome, and is removed when the checkout finishes loading:

image

4. Confirm that the loader still works in Firefox.

### Other information:

### DIFF
--- a/src/modal-checkout/modal.js
+++ b/src/modal-checkout/modal.js
@@ -261,7 +261,9 @@ domReady( () => {
 					spinner.style.display = 'none';
 				} );
 			}
-		} else {
+		} else if ( 'about:blank' !== location.href ) {
+			// Make sure the iframe has actually loaded something, even if not the expected container.
+			// This check prevents an issue in Chrome where the 'load' event fired twice and the spinner was hidden too soon.
 			spinner.style.display = 'none';
 		}
 	} );


chickenn00dle (approved):
Nice! Good catch. Looks good 👍

laurelfulford:
Thanks @chickenn00dle!

PR #1762, opened 2024-06-04
fix(ras-acc): tweak billing, shipping columns styles

laurelfulford:
### All Submissions:

This PR makes some minor tweaks to the billing/shipping field styles in the modal checkout, to make sure they sit in even columns. 

Rather than doing a bigger overhaul of the CSS, I tried to leverage the existing classes and a bit of hacky CSS to make fields full width when they'd otherwise be filling half the space.

See 1205234045751551-as-1207477278860637

1. Start on `epic/ras-acc` and set up at least two modal checkout situations: the donate block, and a checkout button for a regular product. 
2. Run through a checkout for each; with the default settings, you'll likely have a little orphan field, like this:

image

4. Confirm that the checkout with regular settings now doesn't have a lonely field:

image

5. Navigate to WP Admin > Newspack > Reader Revenue, and toggle off some of the billing fields in the Donate tab, then run through the Donate checkout to make sure there are no gaps.
6. Repeat step 5 a couple times to make sure there aren't weird combinations that will leave a gap. Basically anytime a field that's normally floated left or right is alone on their row, they should become full-width.

### Other information:

### DIFF
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -50,7 +50,6 @@ final class Modal_Checkout {
 		add_filter( 'woocommerce_get_checkout_order_received_url', [ __CLASS__, 'woocommerce_get_return_url' ], 10, 2 );
 		add_filter( 'wc_get_template', [ __CLASS__, 'wc_get_template' ], 10, 2 );
 		add_filter( 'woocommerce_checkout_fields', [ __CLASS__, 'woocommerce_checkout_fields' ] );
-		add_filter( 'woocommerce_checkout_fields', [ __CLASS__, 'change_phone_css_class' ] );
 		add_filter( 'woocommerce_update_order_review_fragments', [ __CLASS__, 'order_review_fragments' ] );
 		add_filter( 'newspack_recaptcha_verify_captcha', [ __CLASS__, 'recaptcha_verify_captcha' ], 10, 2 );
 		add_filter( 'woocommerce_enqueue_styles', [ __CLASS__, 'dequeue_woocommerce_styles' ] );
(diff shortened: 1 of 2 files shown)


includes/class-modal-checkout.php
```diff
@@ -759,6 +760,18 @@ public static function woocommerce_checkout_fields( $fields ) {
...
 	}
 
+	/**
+	 * Add the form-row-last CSS class to billing phone field.
+	 *
+	 * @param array $fields Checkout fields.
+	 *
+	 * @return array
+	 */
+	public static function change_phone_css_class( $fields ) {
```
chickenn00dle: Is there a reason we are using a separate method for this vs including this in the `woocommerce_checkout_fields` method above?

If not, I would suggest moving it there. Otherwise, we should probably add an `is_modal_checkout` check to make sure we are only applying this for modal checkout.

chickenn00dle (changes requested):
Nice! Works as expected. I left a comment for one small issue, then this should be good to go 👍

includes/class-modal-checkout.php
laurelfulford: Ha, nope! I was having a hard time getting it to work the way I expected at first, and I didn't think to go back and try to consolidate these once it was okay. 

This has been updated in [65c3711](https://github.com/Automattic/newspack-blocks/pull/1762/commits/65c3711cf2e81cf225274bef4c58163b719987ee)!

laurelfulford:
Thanks @chickenn00dle! Good call on combining the functions -- this is ready for your eyes again!

chickenn00dle (approved):
Thanks Laurel!

PR #1764, opened 2024-06-10
fix(ras-acc): adjust nyp suggested price variation modal markup

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1205234045751551/1207428738079159/f

Adjusts nyp price summary styles in variation modal to match other price summary formats:

Screenshot 2024-06-11 at 13 17 54

- Add a checkout button for a variable product to any page. Ensure the `Allow the reader to select the variation before checkout.` option is checked in block settings.
- Add a second checkout button for a variable subscription to any page. Again, ensure the `Allow the reader to select the variation before checkout.` option is checked in block settings.
- Enable NYP, and set at least one of the variations (but not all) to use nyp. Do this for both the variable product AND the variable subscription
- As a reader, select the variable product checkout button and confirm the variable modal appears. Confirm the alignment of the prices match up, and the suggested price text appears above the price as pictured above
- Do the same for the variable subscription checkout button. Confirm the alignment of the suggested price text, the price, and the "per month" does not appear strange.
- Play with the various subscription variation options (i.e. trial period, sign up fee, etc.) and repeat the previous step

### Other information:

### DIFF
--- a/src/modal-checkout/modal.js
+++ b/src/modal-checkout/modal.js
@@ -15,6 +15,7 @@ const IFRAME_NAME = 'newspack_modal_checkout_iframe';
 const IFRAME_CONTAINER_ID = 'newspack_modal_checkout_container';
 const MODAL_CHECKOUT_ID = 'newspack_modal_checkout';
 const MODAL_CLASS_PREFIX = `${ CLASS_PREFIX }__modal`;
+const VARIATON_MODAL_CLASS_PREFIX = 'newspack-blocks__modal-variation';
 
 domReady( () => {
 	const modalCheckout = document.querySelector( `#${ MODAL_CHECKOUT_ID }` );
@@ -175,7 +176,7 @@ domReady( () => {
 						} );
 					}
 					const formData = new FormData( form );
-					const variationModals = document.querySelectorAll( `.${ MODAL_CLASS_PREFIX }-variation` );
+					const variationModals = document.querySelectorAll( `.${ VARIATON_MODAL_CLASS_PREFIX }` );
 					// Clear any open variation modal.
 					variationModals.forEach( variationModal => {
 						closeModal( variationModal );
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -466,6 +466,10 @@ final class Modal_Checkout {
 										'variation_id' => $variation_id,
 									];
 
+									// Remove colon for nyp variations.
+									if ( class_exists( '\WC_Name_Your_Price_Helpers' ) && \WC_Name_Your_Price_Helpers::is_nyp( $variation->get_id() ) ) {
+										$price = str_replace( ':', '', $price );
+									}
 									?>
 									<li class="newspack-blocks__options__item"">
 										<div class="summary">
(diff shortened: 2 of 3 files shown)


chickenn00dle:
@laurelfulford I'm tagging you for review as this one involves some shady style changes that I'd love your second opinion on!

laurelfulford (changes requested):
Thanks for tackling this one @chickenn00dle! My feedback is 100% super-nitpicky:

The negative margin gets the spacing super close, but it's not quite 1:1 with how the absolutely positioning gets things for the Sale price -- things are off a little bit when you compare the two:

nyp-3

(Told you it was going to be nitpicky!). 

We're probably going to struggle to get it 1:1 with the different approaches, but if we wanted to use absolute positioning here as well, it looks like the NYP plugin has a filter for that price, `wc_nyp_suggested_price_html`. 

I tested it out with some _super_ clunky HTML, but something like this would let us wrap the 'Suggested price' text in its own `span` with the CSS class `.suggested-prefix`, which could then be used to absolutely position the text the same way as the [sale price](https://github.com/Automattic/newspack-blocks/blob/epic/ras-acc/src/modal-checkout/modal.scss#L191-L197):

```
add_filter( 'wc_nyp_suggested_price_html', [ __CLASS__, 'update_suggested_price_html' ], 10, 2 );

... 

public static function update_suggested_price_html( $html, $product ) {
	$html = str_replace( '<span class="suggested-text">', '<span class="suggested-text"><span class="suggested-prefix">', $html );
	$html = str_replace( '<span class="woocommerce-Price-amount amount">', '</span><span class="woocommerce-Price-amount amount">', $html );
	return $html;
}
```

[Edited to add: I belatedly realized this may also work where you're removing the colon in the Suggested Price string!]

I realize I'm well into the weeds, so let me know if this sounds like mega-overkill 🙂 Marking as 'request changes' but I'm happy to discuss how far we want to go with these edge cases!

src/modal-checkout/modal.js
```diff
@@ -175,7 +176,7 @@ domReady( () => {
 						} );
 					}
 					const formData = new FormData( form );
-					const variationModals = document.querySelectorAll( `.${ MODAL_CLASS_PREFIX }-variation` );
+					const variationModals = document.querySelectorAll( `.${ VARIATON_MODAL_CLASS_PREFIX }` );
```
chickenn00dle, line 179: Added this quick fix to address a problem loading the variation modal introduced in https://github.com/Automattic/newspack-blocks/pull/1751

chickenn00dle:
Nice! Thanks for the suggestion @laurelfulford.

I went with your approach, but placed the string replacement directly in the `render_variation_selection` method since adding it to the nyp hook will result in replacing the html everywhere as opposed to just in the checkout modal: https://github.com/Automattic/newspack-blocks/pull/1764/commits/1823fb9cb6ff84446edf7833360a5e890194e0f7

I also noticed I accidentally broke the variation modal in another merged PR, so added the fix to this PR since it was just one line and this PR is related to the variation modal (see https://github.com/Automattic/newspack-blocks/pull/1764/files#r1644863617)

laurelfulford (approved):
> I went with your approach, but placed the string replacement directly in the render_variation_selection method since adding it to the nyp hook will result in replacing the html everywhere as opposed to just in the checkout modal

Ah, good catch, @chickenn00dle! 

This looks good to me -- thanks for making the changes! :shipit:

PR #1769, opened 2024-06-14
fix(block-styles): make sure columns style targets direct children

thomasguillot:
### All Submissions:

If you use a Columns Block inside a Columns Block that as one one the "Move first col..." styles it will also be affected. It's because at the moment the CSS targets ALL the Columns inside a Columns Block that has these specific styles. Instead, it should only target the direct Column Blocks.

__Before:__

1-error

__After:__

2-fixed

1. Create a new page and add a Columns Block, change the style to "Move first col..." (test both styles)
2. Inside this Columns Block, add another Columns Block but keep the style to Default (or Border)
3. Notice the first column moves too.
4. Publish the page.
5. Switch to this branch and refresh the page (and editor)
6. The child Columns Block should render correctly now, with the Column un-moved.

### Other information:

### DIFF
--- a/src/block-styles/core/columns/editor.scss
+++ b/src/block-styles/core/columns/editor.scss
@@ -11,9 +11,9 @@
 
 .wp-block-columns {
 	@include mixins.media( tablet ) {
-		&.is-style-first-col-to-second [data-type='core/column']:nth-child( 2 ),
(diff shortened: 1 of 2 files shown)


laurelfulford (approved):
Ah, nice find and fix, @thomasguillot! 🚢

PR #1770, opened 2024-06-14
fix(block-styles): columns shouldn't enforce gap for block themes

thomasguillot:
### All Submissions:

If we're using Newspack Blocks with a block theme, like Newspack Block Theme, we shouldn't enforce the gap between columns to be 32px since this can be modified in the editor.

This PR excludes the newspack-block-theme so a publisher can enjoy full control of the Block Spacing setting whilst using our Block Theme.

Setting

1. Activate NP Block Theme
2. In a page, add a columns block.
3. Try to update the block spacing -- nothing happens
4. Switch to this branch
5. Try again

### Other information:

### DIFF
--- a/src/block-styles/core/columns/editor.scss
+++ b/src/block-styles/core/columns/editor.scss
@@ -1,7 +1,7 @@
 @use '../../../shared/sass/variables';
 @use '../../../shared/sass/mixins';
 
-.editor-styles-wrapper div.wp-block-columns {
+body:not( [class*="block-theme"] ) .editor-styles-wrapper div.wp-block-columns {
 	gap: 32px;
 
 	&.is-style-borders {
@@ -34,11 +34,11 @@
 
 				&::after {
 					border-right: 1px solid variables.$color__border;
-					bottom: 12px;
+					bottom: 0;
 					content: '';
 					position: absolute;
 					right: -32px;
-					top: 12px;
+					top: 0;
 				}
 
 				&:nth-child( odd )::after {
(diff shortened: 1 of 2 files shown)


laurelfulford (changes requested):
Thanks for looking at this, @thomasguillot! 

I think we may need to do something about the Borders style to keep the gap -- without it, it ends up looking kind of broken:

image

By default, the gap is doubled in the columns block when the borders are applied (`64px`) and then the borders (`:after` pseudo elements) are absolutely positioned at the halfway mark (`32px`). 

I'm not really sure the best way to fix this - the style is just so flipping hacky. Ideally there would be a way to set the `:after` elements to automatically be half the gap width,  whatever it is, but I don't think there's a nice way to do that.

Maybe for the short-term the gap style can stay for the border style, regardless of theme? Or we could remove the border style from the columns block based on theme? There is an option to add borders to individual blocks when you use a block theme (including columns) so there might be a way to recreate it when needed, it's just going to be a lot more manual 😕 

Let me know what you think or if you have any questions!

laurelfulford (approved):
Following up on our Slack conversation, this was definitely an error in my testing! I had taken the latest release of the block theme to put on a test site I was already running the classic theme on, to compare how this change affected them, and ended up with some out-of-date styles 😓 

Switching the theme to the `trunk` version of the block theme, the border styles look good. The gap option doesn't work with that style, which I think is the best compromise for now (at least unless we can figure out a better way to build the border styles to work with block themes). Switching my review to 'approved'!

thomasguillot:
> The gap option doesn't work with that style, which I think is the best compromise for now (at least unless we can figure out a better way to build the border styles to work with block themes)

Yes, it's not ideal, but I'm not sure how we could do it differently 🤔 unless we can somehow add an empty div or something between each column blocks.

PR #1772, opened 2024-06-14
Block caching

claudiulodro:
### All Submissions:

This PR introduces caching for the Posts and Carousel blocks. This is off by default for now, so I can run this on a dozen or so sites for a while and make sure it's solid, and we'll switch it to on by default at some point before elections.

You'll probably want to use a WP Cloud site for testing so that you have a good object cache running, but it may work OK on your local site. It definitely shouldn't cause issues or break anything if an object cache isn't present.

1. To your wp-config add: `define( 'NEWSPACK_LOG_LEVEL', 2 );` and `define( 'NEWSPACK_BLOCKS_CACHE_BLOCKS', true );`.
2. Visit a Page with posts blocks and/or carousel blocks. On the first visit you should see something like this for each block on the page in the logs:
```
[14-Jun-2024 19:43:33 UTC] [74084][NEWSPACK][Newspack_Blocks_Caching::debug_log]: Checking cache for: np_cached_block_f87a58c1699b8f382352a28962df71a8
[14-Jun-2024 19:43:33 UTC] [74084][NEWSPACK][Newspack_Blocks_Caching::debug_log]: Cached data not found for: np_cached_block_f87a58c1699b8f382352a28962df71a8
[14-Jun-2024 19:43:33 UTC] [74084][NEWSPACK][Newspack_Blocks_Caching::debug_log]: Caching block: np_cached_block_f87a58c1699b8f382352a28962df71a8
```
3. Refresh the page. It should load much faster but should otherwise look exactly the same. In the logs you should see something like:
```
[14-Jun-2024 19:43:33 UTC] [74084][NEWSPACK][Newspack_Blocks_Caching::debug_log]: Checking cache for: np_cached_block_ae84d66a8217d97cc4e70226086bb965
[14-Jun-2024 19:43:33 UTC] [74084][NEWSPACK][Newspack_Blocks_Caching::debug_log]: Serving cached block: np_cached_block_ae84d66a8217d97cc4e70226086bb965
```
4. Wait 5 minutes. This should invalidate the cache and force a re-cache of everything.
5. Add a Posts block to a widget area (e.g. sidebar). Visit a post, observe the block gets cached. Visit a different post, observe the cached version of the block gets served. This does not happen on Pages but should happen on Posts. See the explanation of the heuristic in the code.
6. `wp cache flush` to flush the cache if you want to start from the top.

### Other information:


includes/class-newspack-blocks-caching.php
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	/**
+	 * Debug logging for observing cache behavior.
+	 *
+	 * @param string $message Message to log.
+	 */
+	protected static function debug_log( $message ) {
+		if ( defined( 'NEWSPACK_LOG_LEVEL' ) && (int) NEWSPACK_LOG_LEVEL >= 2 ) {
+			Newspack\Logger::log( $message );
```
adekbadek, line 90: There should be a `method_exists` check here, it's not guaranteed that this code will always run alongside `newspack-plugin`.

includes/class-newspack-blocks-caching.php
```diff
@@ -0,0 +1,151 @@
...
+
+/**
+ * Cache dynamic blocks for improved performance.
+ */
+class Newspack_Blocks_Caching {
+	/**
+	 * Add hooks and filters.
+	 */
+	public static function init() {
+		if ( defined( 'NEWSPACK_CACHE_BLOCKS' ) && NEWSPACK_CACHE_BLOCKS ) {
```
adekbadek: Nit: the prefix should match the repository name:
```suggestion
		if ( defined( 'NEWSPACK_BLOCKS_CACHE_BLOCKS' ) && NEWSPACK_BLOCKS_CACHE_BLOCKS ) {
```

includes/class-newspack-blocks-caching.php
```diff
@@ -0,0 +1,151 @@
...
+				'order'          => 'DESC',
+				'posts_per_page' => 1,
+			]
+		);
+		if ( empty( $latest_update ) ) {
+			return $block_html;
+		}
+
+		$last_update_timestamp = strtotime( $latest_update[0]->post_modified_gmt );
+		if ( $last_update_timestamp > $cached_data['timestamp_generated'] ) {
```
adekbadek: This looks like the only method of invalidating the cache is publishing or updating a post. Changes to e.g. categories would not affect the cache. On my test site, I had a block set to display posts of "Sports" category. I've then removed this category - the editor correctly displayed a hole, while the frontend was displaying the cached content:

| Editor | Frontend | 
|--------------|-----------|
| image | image |
  
This is definitely an edge case, but I think there should at least be a way to reset the cache (without having to fake-update a post). Alternatively, the cache could also be invalidated on any taxonomy update (e.g. removing a tag).

adekbadek (changes requested):

includes/class-newspack-blocks-caching.php
claudiulodro: Good call. Fixed in https://github.com/Automattic/newspack-blocks/pull/1772/commits/4b570413299b5c22aabf9973d59290fed8b4fcc0

includes/class-newspack-blocks-caching.php
claudiulodro: https://github.com/Automattic/newspack-blocks/pull/1772/commits/4b570413299b5c22aabf9973d59290fed8b4fcc0

includes/class-newspack-blocks-caching.php
claudiulodro: The latest version sort-of solves this problem by having the cache expire after 5 minutes. The overall performance is the same according to the logs, and we don't have to worry about nearly as many edge cases if they fix themselves after a few minutes.

claudiulodro:
@adekbadek sorry about the delay on the revisions! 

I've adjusted the strategy after fine-tuning the logic on some sites. Now it simply caches for 5 minutes (or however long). The check to determine when the last post was published actually created DB load sometimes, and simply expiring the cache every 5 minutes is simpler and seems to achieve the same performance.

includes/class-newspack-blocks-caching.php
```diff
@@ -0,0 +1,146 @@
...
+		// Posts and other publicly_queryable post types should all share a cache, because 99% of the time article blocks
+		// are in the sidebar, below-content, or (if within content) fetching specific posts. We want an article block in
+		// the e.g. sidebar to be served from cache across all posts.
+		// The tradeoff is that occasionally the current post may show up in an article block on a post.
+		// Archives and non-singular should all use the global cache, because there is nothing that would need de-duplication.
+		if ( is_singular() ) {
+			$post_type        = get_post_type();
+			$post_type_object = get_post_type_object( $post_type );
+			if ( ! $post_type_object->publicly_queryable ) {
+				$block_attributes['parent_post'] = get_the_ID();
```
leogermani: I have a suggestion.

Instead of adding the `parent_post` id to the mix to build the cache key, we should use it as the cache group. We could have a `global` cache group for everything, and then a `post-$ID` group for pages.

That way we can a way (a button, a CLI) to clear the cache of all blocks in a page if needed at any point. (5 minutes can feel like an eternity for a news site homepage).

And also, having the global cache group allows us to clear all blocks cache without having to clear all the cache

claudiulodro: Great suggestion! So it would be e.g. `newspack_blocks` cache group for posts/global and `newspack_blocks-post-123` for pages (with the appropriate post ID)?

adekbadek (approved):
Works fine in my testing. Accepting but noting that there an unresolved discussion above 👆

claudiulodro:
Note to self: On cached blocks, it would be best to still populate that global array of used posts, so that other uncached blocks on the page continue to deduplicate nicely

includes/class-newspack-blocks-caching.php
claudiulodro: https://github.com/Automattic/newspack-blocks/pull/1772/commits/d4a8be67c4866e513162483ef48b3fd05fec0876

claudiulodro:
I spent a couple hours trying to do the global posts array population part, so the deduplication logic works if e.g. half the posts blocks are cached and half are not, but it was pretty complicated. I'll add that part in a follow-up PR after this is merged. Should be OK though since it's a pretty rare situation and fixes itself after a couple minutes. If it looks good with the cache groups, I think this can be shipped.

PR #1774, opened 2024-06-14
fix: update dependencies to avoid conflict in WP 6.6

dkoo:
### All Submissions:

Refactoring required for changes in https://github.com/Automattic/newspack-scripts/pull/209. Note that this PR will not pass CI tests and should NOT be merged until https://github.com/Automattic/newspack-scripts/pull/209 is released as an NPM package, so this repo can install it as a dependency.

Follow instructions in https://github.com/Automattic/newspack-scripts/pull/209.

### Other information:


package.json
```diff
@@ -7,52 +7,48 @@
...
 		"lint:php:staged": "./vendor/bin/phpcs --filter=GitStaged",
-		"format:js": "prettier 'src/**/*.{js,jsx,ts,tsx}' --write",
-		"lint:scss": "stylelint '**/*.scss' --customSyntax postcss-scss --config=./node_modules/newspack-scripts/config/stylelint.config.js",
-		"format:scss": "prettier --write 'src/**/*.scss'",
-		"lint:scss:staged": "stylelint --customSyntax postcss-scss --config=./node_modules/newspack-scripts/config/stylelint.config.js",
+		"lint:scss": "newspack-scripts wp-scripts lint-style '**/{src,includes}/**/*.scss' --customSyntax postcss-scss",
+		"lint:scss:staged": "newspack-scripts wp-scripts lint-style --customSyntax postcss-scss",
+		"format:scss": "newspack-scripts wp-scripts lint-style '**/{src,includes}/**/*.scss' --customSyntax postcss-scss --fix",
 		"i18n": "NODE_ENV=production npm run build; bash ./bin/update-translations.sh",
 		"typescript:check": "newspack-scripts typescript-check",
```
chickenn00dle, line 48: Getting an error when I run this:

```
> @automattic/newspack-blocks@3.3.0 typescript:check
> newspack-scripts typescript-check

[newspack-scripts@1.0.0] Starting TypeScript check…
src/blocks/donate/edit/index.tsx:11:22 - error TS2307: Cannot find module '@wordpress/api-fetch' or its corresponding type declarations.

11 import apiFetch from '@wordpress/api-fetch';
                        ~~~~~~~~~~~~~~~~~~~~~~


Found 1 error in src/blocks/donate/edit/index.tsx:11

npm verbose exit 2
npm verbose code 2
```

package.json
```diff
@@ -7,52 +7,48 @@
...
+		"build": "newspack-scripts wp-scripts build",
+		"start": "npm ci && npm run watch",
+		"watch": "newspack-scripts wp-scripts start",
 		"test": "newspack-scripts test",
 		"lint": "npm run lint:scss && npm run lint:js",
-		"lint:js": "eslint 'src/**/*.{js,jsx,ts,tsx}'",
-		"lint:js:staged": "eslint --ext .js,.jsx,.ts,.tsx",
+		"lint:js": "newspack-scripts wp-scripts lint-js '**/{src,includes}/**/*.{js,jsx,ts,tsx}'",
+		"lint:js:staged": "newspack-scripts wp-scripts lint-js --ext .js,.jsx,.ts,.tsx",
+		"format:js": "newspack-scripts wp-scripts format '**/{src,includes}/**/*.{js,jsx,ts,tsx}'",
```
chickenn00dle, line 40: Running this results in many changes. Is this expected? I would assume if linting passes, this would not trigger any updates.

chickenn00dle (dismissed):
I left a few comments about npm script issues/questions, but also ran into an issue when attempting to smoke test on current WP. Looks like none of the blocks are available in the editor:

Screenshot 2024-06-20 at 14 31 38

This does not happen on 6.6:

Screenshot 2024-06-20 at 14 34 51

It's also worth noting that when I add blocks to a page on 6.6, then downgrade, the blocks continue to render on the front end, but the back end shows not supported errors:

Screenshot 2024-06-20 at 14 40 27

package.json
dkoo: That might be due to the [dependency extraction](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-dependency-extraction-webpack-plugin/) for `@wordpress/*` modules. Maybe we can just ignore errors like that for now.

package.json
dkoo: Hmm, it looks like `format:js` is applying Prettier formatting to the code (so no whitespace in `()` or `[]`). I'm not sure if we want that. The majority of what it's changing seems to be whitespacing stuff, so maybe we can remove this script for now if it's not useful...

package.json
dkoo: Weird! I'm not seeing this error when I run this locally.

Try running `rm -rf node_modules && start` from the Blocks repo directory and see if it resolves things? Sometimes Node has a habit of messing up the install if there are already `node_modules` installed from a prior run...

package.json
dkoo: Ok, I have a better answer on this. First off, d89cd4b restores Prettier config files in this repo, but extending a base config file from `newspack-scripts`. But also, `lint-js` and `format:js` are using ESLint and Prettier, respectively, and they have different opinions on formatting. By default, the `@wordpress/scripts` ESLint config file will import Prettier rules and report them as errors, but this seems like a heavy-handed change for us and would require that we change our established coding style, which is more ESLint-based. So adding `'prettier/prettier': 'off'` [here](https://github.com/Automattic/newspack-scripts/pull/209/files#diff-1bd145b27e5bc507fcffdc88b445a1b341284a8f581d53375452838316832a41R75) makes it so that ESLint is only concerned about code quality rules and not Prettier's formatting rules. However, `format:js` still uses Prettier, so running that will result in all JS/JSX/TS/TSX files being reformatted according to Prettier's formatting rules. In the future we may choose to update our team's coding style, so we can keep this script available, but for now we can just choose to not use it until we're ready.

dkoo: Unfortunately there's no equivalent solution for the `lint:scss` implementation, so that's why we're reformatting all SCSS files as part of this update. But that seems less risky than reformatting all JS at this time, IMO.

package.json
dkoo: This should be fixed now with the `v5.6.0-alpha.4` version (and up) of `newspack-scripts`.

chickenn00dle: > This should be fixed now with the v5.6.0-alpha.4 version (and up) of newspack-scripts.

Yup! Seems to be all good now 👍

package.json
chickenn00dle: Thanks for digging into this! 

> In the future we may choose to update our team's coding style, so we can keep this script available, but for now we can just choose to not use it until we're ready.

I'm fine with this, although I would suggest removing this altogether for the time being to prevent any confusion and re-add if/when we become ready. In any case, not a blocker.

chickenn00dle:
I did some smoke tests and am running into an issue when WP 6.5.4 is installed. Looks like when creating a new newsletter, layouts fail and I am taken directly to the editor with no content:

Screenshot 2024-06-21 at 10 10 22

Also, I'm not sure if this is because this branch is not up-to-date with trunk yet, but I'm noticing some issues with the social icons block on 6.6:

Screenshot 2024-06-21 at 10 17 49

I spot checked a handful of blocks and didn't see other issues, but didn't go through them all.

package.json
dkoo: I'm okay with that!

miguelpeixe (approved):
After rebasing, it should be good to go!

PR #1777, opened 2024-06-21
feat(checkout-button): add wp-block-button to the wrapperclasses

thomasguillot:
### All Submissions:

The themes (Classic && Block) are missing some `:hover` styles because the wrapper doesn't have the `wp-block-button` class. This PR adds the class (on save).

Note: The issue is even more visible with the Block Theme because the outline style variation simply doesn't work.

error
fix

1. Add a checkout button block
2. Change the style to outline
3. Publish page and check front-end
4. No hover
5. Switch to this branch
6. Refresh editor (note: for some reason it breaks the block -- probably because it adds the class on save -- I could use some help here!)
7. Save the block again
8. Refresh front-end

### Other information:

### DIFF
--- a/src/blocks/checkout-button/save.js
+++ b/src/blocks/checkout-button/save.js
@@ -43,7 +43,7 @@ export default function save( { attributes, className } ) {
 		...spacingProps.style,
 	};
 
-	const wrapperClasses = classnames( className, {
+	const wrapperClasses = classnames( className, 'wp-block-button', {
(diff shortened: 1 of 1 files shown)


adekbadek (approved):
> for some reason it breaks the block -- probably because it adds the class on save 

That's correct – the stored HTML changes, so the block is marked with "This block contains unexpected or invalid content." warning. But the block recovery button works. I'm not sure if we can do anything about it, since the rendering is done JS-side. I think a proper hover state is worth the trade-off of getting this warning.

PR #1553, opened 2023-10-11
Hotfix: WP 6.4 HPB rendering issue

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

https://github.com/Automattic/newspack-blocks/pull/1551 as a hotfix

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1794, opened 2024-07-08
fix: account for patterns when rendering hpp in editor

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1207720049982626/1207721228317933/f

This PR fixes an issue introduced when updating WP to 6.6 where the HPP block rendered within a synced pattern causes the hpp block to render defaults in the editor after a post is edited.

Before:

Screenshot 2024-07-08 at 17 11 48

After:

Screenshot 2024-07-08 at 17 11 23

1. Start on a site running WP 6.6 RC
2. Add the HPP block to any page or post, convert to a saved patter, and save
3. Reload the post editor. The HPP block should be rendering real post data
4. Now edit anything in the post aside from the HPP synced pattern. On `release` this will cause the real post data to revert to placeholder data. On this branch, it will not.

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/utils.ts
+++ b/src/blocks/homepage-articles/utils.ts
@@ -222,12 +222,25 @@ export const postsBlockSelector = (
 		attributes,
 	}: { clientId: Block[ 'clientId' ]; attributes: HomepageArticlesAttributes }
 ): HomepageArticlesPropsFromDataSelector => {
-	const { getEditedPostAttribute } = select( 'core/editor' );
-	const editorBlocks = getEditedPostAttribute( 'blocks' ) || [];
 	const { getBlocks } = select( 'core/block-editor' );
-	const editorBlocksIds = getEditorBlocksIds( editorBlocks );
+	const { getEditedPostAttribute } = select( 'core/editor' );
+
+	const editorBlocks = getEditedPostAttribute( 'blocks' ) || [];
+	const allEditorBlocks = [];
+
+	for ( const block of editorBlocks ) {
+		// Get pattern blocks as well.
+		if ( block.name === 'core/block' ) {
+			allEditorBlocks.push( ...getBlocks( block.clientId ) );
+		} else {
+			allEditorBlocks.push( block );
+		}
+	}
+
+	const editorBlocksIds = getEditorBlocksIds( allEditorBlocks );
 	const blocks = getBlocks();
 	const isWidgetEditor = blocks.some( block => block.name === 'core/widget-area' );
+
 	// The block might be rendered in the block styles preview, not in the editor.
 	const isEditorBlock =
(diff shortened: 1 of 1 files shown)


src/blocks/homepage-articles/utils.ts
```diff
@@ -222,12 +222,25 @@ export const postsBlockSelector = (
...
+		} else {
+			allEditorBlocks.push( block );
+		}
+	}
+
+	const editorBlocksIds = getEditorBlocksIds( allEditorBlocks );
 	const blocks = getBlocks();
 	const isWidgetEditor = blocks.some( block => block.name === 'core/widget-area' );
+
 	// The block might be rendered in the block styles preview, not in the editor.
```
chickenn00dle, line 244: Just noting I'm not sure what `block styles preview` refers to here, so haven't been able to test whether these changes break anything in this context. Would love any guidance here to confirm this.

chickenn00dle:
cc @laurelfulford

src/blocks/homepage-articles/utils.ts
laurelfulford: I _think_ this might be referring this preview (which also doesn't work in WP 6.5)

image

... and based on the next bit, it sounds like it also may have to do with the preview you get adding this block to a widget area (both in the Customizer, and possibly under Appearance > Widgets).

I can give this a spin to make sure it fixes the issue, but I'd love to also get @adekbadek's eyes on the code since [he fought with the placeholder stuff pretty recently](https://github.com/Automattic/newspack-blocks/pull/1553) -- I'm not sure I'm totally getting the nuance of what's happening there 🙂

laurelfulford (commented):
Thanks for tackling this one, @chickenn00dle! 

It works well in my testing -- I tested the following with the WP 6.6 RC and WP 6.5.5:

* Adding a Homepage Post block to the editor and making it into a synced pattern
* Reordering the synced pattern along with other regular homepage post blocks
* Starting a fresh page and inserting the synced pattern
* Adding more blocks before and after the synced pattern
* I also tested the block in widget spaces in the Customizer and the Appearance > Widget screen.

I also did the above with the Post Carousel block.

In all cases, the stories remained stories and not placeholders so your changes seem to have done the trick! I'm leaving this as a comment and not a review, just to get confirmation no cases were missed there based on the code, but this gets a 👍 from me for addressing the issue and not introducing any new issues I could see.

adekbadek (approved):
Works well!

src/blocks/homepage-articles/utils.ts
adekbadek: Yeah, the preview is not working (not sure why), but you can see the preview version in the patterns explorer:

image

PR #1795, opened 2024-07-09
fix(ras-acc): update woocommerce checkout payment error styles

chickenn00dle:
### All Submissions:

**Important: This PR relies on updates to newspack ui which will be added in https://github.com/Automattic/newspack-plugin/pull/3235. DO NOT MERGE until that PR has been merged.**

Closes https://app.asana.com/0/1207720049982626/1207546771559780/f

This PR updates the styles for the generic woocommerce payment errors that appear for invalid payment attempts:

Screenshot 2024-07-10 at 12 57 30

1. Make sure to have `epic/ras-acc` checked out in `newspack-plugin`
2. Initiate modal checkout via any donate or checkout button
3. Enter all checkout info and proceed to the payment portion of modal checkout
4. Enter any invalid payment details (for stripe based payment gateways, you can use the following test card with any expiration date and any cvv which will trigger a failure: 4000 0000 0000 0002)
5. On `epic/ras-acc` the error will have default woo styles and clash a bit with our newspack ui styles. On this branch the styles will match that of our newspack notices as pictured above

Bonus: Test some other payment gateways and try to generate other errors in the payment form (some payment gateways will trigger this error when attempting to submit empty fields while others will not)

### Other information:

### DIFF
--- a/src/modal-checkout/checkout.scss
+++ b/src/modal-checkout/checkout.scss
@@ -464,22 +464,11 @@
 		margin-top: 0;
 	}
 
-	form.checkout .woocommerce-NoticeGroup {
-		align-items: center;
-		background: color.adjust( wp-colors.$alert-red, $lightness: 51% );
-		display: flex;
-		justify-content: flex-start;
-		margin: 32px 0;
-		padding: 6px 8px;
-		border-radius: 2px;
-		box-shadow: 0 0 8px 4px rgba( black, 0.08 );
-		border-left: 3px solid wp-colors.$alert-red;
-		box-sizing: border-box;
-
-		ul.woocommerce-error {
-			background: none;
-			color: wp-colors.$gray-900;
-			margin: 0 !important;
-		}
+	form.checkout .woocommerce-NoticeGroup ul.woocommerce-error {
+		background: none;
+		color: inherit;
+		list-style-type: none;
+		padding-left: 0;
+		padding-right: 0;
 	}
 }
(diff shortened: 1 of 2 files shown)


dkoo (approved):

laurelfulford (changes requested):
Just a small one on this one: 

It looks like it's picking up styles from the classic theme for `woocommerce-error`, making the font size `12.8px` instead of `14px`.

The `woocommerce-error` class is inside of the Notice markup, so I don't think setting a font size on the Newspack UI side will fix this.

I think setting a font size of `inherit` for `.woocommerce-message`, `.woocommerce-error`, and `.woocommerce-info` in the modal.scss would be the way to go, but there's probably other ways to fix it! That's just where our modal-specific overrides live. 

Just let me know if you have any questions about this!

chickenn00dle:
Thanks for pointing this out @laurelfulford!

I tried adding the font size rule to `modal.scss` but that didn't do the trick. I think `checkout.scss` is where the iframed checkout overrides are actually applied. In any case I've added them in https://github.com/Automattic/newspack-blocks/pull/1795/commits/ea3b56929120d42cc8a96659ae7521e9da62e123 and it seems to do the trick.

Let me know if you had another place in mind for this instead!

dkoo:
Thanks for catching that, @laurelfulford!

laurelfulford (approved):
> I think checkout.scss is where the iframed checkout overrides are actually applied. In any case I've added them in https://github.com/Automattic/newspack-blocks/commit/ea3b56929120d42cc8a96659ae7521e9da62e123 and it seems to do the trick.

Ah thanks for catching that, @chickenn00dle! 🤦‍♀️ I was going from memory for where we put styles, and had it wrong!

The font size looks good! I think this one's ready to go, pending the plugin changes -- thanks!

PR #1810, opened 2024-07-26
fix(ras-acc): remove use of jquery blockUI

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1207817176293825/1207836897783913/f

Jquery's block UI can cause some strange behavior sometimes, like form jumpiness, or in this case not overlaying the complete form. This PR removes jquery block UI calls from modal checkout in favor of some custom blocking logic and styling:

Before:
Screenshot 2024-07-29 at 15 29 04

After:
Screenshot 2024-07-29 at 15 29 53

1. As a reader trigger modal checkout via any donation button
2. Go through the checkout flow, paying attention to processing states (any time the form is validating)
3. Confirm the form appears to be in a "loading state" whenever it is processing
4. Repeat the above steps for the checkout button. Make sure to test the coupon and nyp fields as well (coupons and the nyp extension will need to be active and set up for the products for these)

### Other information:

### DIFF
--- a/src/modal-checkout/checkout.scss
+++ b/src/modal-checkout/checkout.scss
@@ -472,4 +472,8 @@
 			font-size: var( --newspack-ui-font-size-xs, 14px );
 		}
 	}
+
+	.processing {
+		opacity: 0.5;
+	}
 }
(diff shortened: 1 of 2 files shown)


laurelfulford (approved):
This looks good! I tested the regular checkout, increasing the payment in the NYP field, and adding a valid and invalid coupon.

The only issue I noticed is that the coupon error message is no longer red, but it looks like that's happening on epic/ras-acc, too -- it's not unique to this PR! I'll file a separate issue for that.

PR #1811, opened 2024-07-27
fix: reCAPTCHA v2 + v3 for RAS-ACC flows

dkoo:
### All Submissions:

Fixes for reCAPTCHA implementation in RAS-ACC modal checkout.

See https://github.com/Automattic/newspack-plugin/pull/3283

### Other information:

### DIFF
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -547,7 +547,7 @@ final class Modal_Checkout {
 
 		$dependencies = [ 'jquery' ];
 		// Add support reCAPTCHA dependencies, if connected.
-		if ( class_exists( 'Newspack\Recaptcha' ) ) {
+		if ( class_exists( 'Newspack\Recaptcha' ) && \Newspack\Recaptcha::can_use_captcha() ) {
 			$dependencies[] = \Newspack\Recaptcha::SCRIPT_HANDLE;
 		}
 
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -545,10 +545,16 @@ final class Modal_Checkout {
 			return;
 		}
 
+		$dependencies = [ 'jquery' ];
+		// Add support reCAPTCHA dependencies, if connected.
+		if ( class_exists( 'Newspack\Recaptcha' ) ) {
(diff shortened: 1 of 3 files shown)


src/modal-checkout/index.js
```diff
@@ -391,6 +398,15 @@ domReady(
 						} );
 						$form.on( 'submit', handleFormSubmit );
 					} else {
+						if ( $form.find( '[name="is_validation_only"]' ) ) {
+							$form.find( '[name="is_validation_only"]' ).remove();
+						}
```
chickenn00dle: Nonblocking nitpick: Maybe we can just create a new variable for the first call to `find` here so we don't need to run this twice.

```suggestion
                        const validationOnlyField = $form.find( '[name="is_validation_only"]' );
                        if ( validationOnlyField.length ) {
							$validationOnlyField.remove();
						}

```

chickenn00dle (approved):
Approving pending changes in the main plugin PR for this.

src/modal-checkout/index.js
dkoo: Great idea!

includes/class-modal-checkout.php
```diff
@@ -528,18 +528,24 @@ class="<?php echo esc_attr( "$class_prefix {$class_prefix}__modal-container news
...
 			return;
 		}
 
 		if ( ! self::is_modal_checkout() ) {
 			return;
 		}
 
+		$dependencies = [ 'jquery' ];
+		// Add support reCAPTCHA dependencies, if connected.
+		if ( class_exists( 'Newspack\Recaptcha' ) ) {
```
chickenn00dle: We need to add a check for `\Newspack\Recaptcha::can_use_captcha()` here so we aren't loading the dependency when recaptcha is disabled.

chickenn00dle (dismissed):

includes/class-modal-checkout.php
chickenn00dle:
```suggestion
		if ( class_exists( 'Newspack\Recaptcha' ) && \Newspack\Recaptcha::can_use_captcha() ) {
```

dkoo: Aha!

chickenn00dle (approved):

PR #1813, opened 2024-07-29
fix(ras-acc): dont skip first step of modal checkout when gifting enabled

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1205234045751551/1207872481635256/f

This PR forces modal checkout to remain on step 1 whenever subscriptions gifting is present. This is because our modal checkout can skip ahead to step 2 when payment details are already saved, however step 2 of the modal does not have the option for gifting.

1. Set up a checkout button block for a subscription product with gifting enabled as well as a second button with a non-subscription product
2. As a reader with a saved payment method, go through checkout for the non-subscription product. Confirm the modal opens and immediately validates into the payment method step of modal checkout
3. Complete checkout being sure to test going back and re-editing details in the form.
4. As a reader with a saved payment method, go through checkout for the subscription product with gifting enabled. Confirm the modal opens into the billing details step of modal checkout
5. Complete checkout being sure to test going back and re-editing details in the form.

### Other information:

### DIFF
--- a/src/modal-checkout/index.js
+++ b/src/modal-checkout/index.js
@@ -136,19 +136,12 @@ domReady(
 				 */
 				if ( $checkout_continue.length ) {
 					setEditingDetails( true );
-					// Perform initial validation so it can skip 1st step if possible.
-					validateForm( true, () => {
-						// Attach handler to "Back" button.
-						$form.on( 'click', '#checkout_back', function ( ev ) {
-							ev.preventDefault();
-							setEditingDetails( true );
-						} );
-						// Remain on 1st step if subscriptions gift options are present.
-						if ( $gift_options.length ) {
-							setEditingDetails( true );
-						}
+					if ( ! $gift_options.length ) {
+						// Perform initial validation so it can skip 1st step if possible.
+						validateForm( true, setReady );
+					} else {
 						setReady();
-					} );
+					}
 				} else {
 					setReady();
 				}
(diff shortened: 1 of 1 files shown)


laurelfulford (changes requested):
The "going to the first screen" behaviour is working as described, but I'm running into an issue on this branch that I can't replicate on epic/ras-acc:

1. As a new reader (not already logged in), purchase a subscription product -- I've been testing with a recurring donation. 
2. On the first screen, check that it's a gift and enter an email address.
3. On the second screen, enter payment information and click "Donate now".

The screen seems to load for a second, and then it returns to a non-loading state and doesn't send it -- it basically gets "stuck" on the last screen. There aren't any JS or form validation errors. 

This doesn't seem to happen if you're already logged in and haven't saved payment information, just when the sign up is part of the steps. 

This seems like a strange one to me, so please let me know if you have issues reproducing it!

chickenn00dle:
> The screen seems to load for a second, and then it returns to a non-loading state and doesn't send it -- it basically gets "stuck" on the last screen. There aren't any JS or form validation errors.

Good catch! I'm not entirely sure, but I think this has something to do with calling `setEditingDetails` multiple times. We initialize with `setEditingDetails( true )`, then the validation method calls `setEditingDetails( false )`, and finally the code I added in this PR initially calls `setEditingDetails( true )` once again when gifting is available. I suspect too many consecutive triggers of `update_checkout` might be causing issues.

In any case, I pushed up a change to prevent validation altogether when gifting is available which seems to do the trick: https://github.com/Automattic/newspack-blocks/pull/1813/commits/f7c91d5406186452a972b844a36f5aee3b730589

Let me know if this works on your end.

laurelfulford (approved):
Thanks @chickenn00dle! 

This looks good! The PR still does what's described, but I can no longer trip it up as a new sign up 🚢

PR #1819, opened 2024-07-31
feat(ras-acc): add GA4 events to modal checkout (redo)

laurelfulford:
### All Submissions:

Along with https://github.com/Automattic/newspack-plugin/pull/3297, reworks the GA4 events to move more of them to the front-end.

Each sequence will include the following actions:

| `action`                   | When it's used      |
| ------------------- | ------------- |
| `opened`                 | When the modal checkout is opened, either from the Donate Block, Checkout Button Block, or Variation picker screen |
| `opened_variations` | When a variation picker is opened from a Checkout button block|
| `loaded`               | When the modal completes loading |
| `continue`    | When the 'Continue' button is clicked |
|`back`            | When the 'Back' button is clicked |
|`dismissed`   | When the modal is closed before completion |
|`form_submission`   | When a submission attempt is made |
|`form_submission_success`   | When a submission attempt is completed (back-end event)

The action types used are:

| `action_type`     | When it's used      |
| ------------- | ------------- |
| `checkout_button` | When the modal trigger is the Checkout Button block |
| `donation`       | When the modal trigger is the Donation block |

We're capturing the following information on most steps:

| What's captured     | When it's captured      |
| -------------------- | --------------------- |
| `amount` | For each step, except when opening a variation picker since the price isn't known |
| `currency` | For each step |
| `is_variable` | Only the initial `open_variations` step when opening a variation picker |
| `product_id` | For each step |
| `product_type`*** | For each step |
| `recurrence` | For each step, except when opening a variation picker for a variable subscription product |
| `referer` | For each step |
| `variation_id` | For steps after picking a product variation |

*** The product types have yet to be finalized, but currently are: `product`, `subscription`, `membership`, and `donation`.

### Open questions

* ~~The "opened" action fires twice for Checkout Button blocks set to use Variable Products that the reader can select. The initial time opens the variation picker, and the second time opens the modal checkout for the product that's picked. Should there be a different `action` param for these?~~ resolved with the `opened_variations` action, added to the table above.
* When applying a coupon, the `amount` doesn't change until the form is successfully submitted (`form_submission_success`). Should this change sooner?
* When changing the price via the NYP field, the `amount` doesn't change until the form is successfully submitted (`form_submission_success`). Should this change sooner?
* ~~What `product_id` should be captured for the Donation products. Technically there's the main Donation product, plus there's a separate product for each frequency. Right now, it (incorrectly) uses both at random points.~~ -- resolved, we're grabbing the specific donation child product being purchased.

~~I've left the `console.log` in place on the GA4 `payload` so it should be relatively easy to see what's been captured and sent at each step.~~ I've removed the `console.log( payload );`, but it can be re-added under the line `window.gtag( 'event', eventName, payload );` in modal-checkout/analytics/ga4/utils/index.js.

1. Apply both this PR and https://github.com/Automattic/newspack-plugin/pull/3297, and run `npm run build` for each.
2. You may need to create products to set this up, but for testing you'll ultimately need:
* A checkout button block with a simple product
* A checkout button block with a variable product, where the reader can pick the variation
* A checkout button block with a simple subscription
* A checkout button block with a variable subscription, where the reader can pick the variation
* A checkout button block with a NYP product
* A checkout buttton block with a variable product, where the variation is pre-selected
* A checkout button block with a membership product
* A donate block.
3. Run through a checkout for each block, and track what `action`s are being fired with what information. Confirm that each step listed in the table above is represented, and is consistently populated.
4. On form submission, the `form_submission_success` is the only event fired from the back-end, so it can be tracked either by watching realtime events in Google Analytics, or by `error_log`ing the value of `$event` by adding `error_log( print_r( $event, true ) );` after [here](https://github.com/Automattic/newspack-plugin/pull/3297/files#diff-3dbb459c03feab9c03a7a641babce543932c7899efccb35734cc1cbf275633e4L154).

As a part of this PR, I also needed to add a fix similar to [this](https://github.com/Automattic/newspack-blocks/pull/1842), but for the buttons that appear in the variation picker. On epic/ras-acc, if you open the variation picker more than once on a page view, duplicate inputs were being added to the form that each button has.

When testing this PR, can you please also check:
1. When you open the variation picker from the checkout button block that a single set of hidden inputs are appended to the form, like the ones in the element inspector here: 
CleanShot 2024-08-30 at 12 28 03
2. If you close and reopen the same variation picker, that no additional hidden inputs are added to each form.

### Other information:


src/blocks/donate/frontend/class-newspack-blocks-donate-renderer-base.php
```diff
@@ -140,12 +140,13 @@ function ( $item ) {
 			'wpbnbd-frequencies--' . count( $configuration['frequencies'] ),
 			$is_tiers_based ? 'wpbnbd--tier-style-' . $tier_style : null,
 		];
-
 		if ( ! Newspack_Blocks::can_use_name_your_price() ) {
 			$class_names[] = 'wpbnbd--nyp-disabled';
 		}
 		$configuration['container_classnames'] = implode( ' ', $class_names );
 
+		$configuration['product_data'] = wp_json_encode( $product_data );
```
leogermani: I'm getting `PHP Warning:  Undefined variable $product_data` here

includes/tracking/class-data-events.php
```diff
@@ -0,0 +1,111 @@
...
+	public static function register_listeners() {
+		if ( ! method_exists( 'Newspack\Data_Events', 'register_handler' ) ) {
+			return;
+		}
+
+		\Newspack\Data_Events::register_listener(
+			'woocommerce_order_status_completed',
+			'modal_checkout_interaction',
+			function ( $order_id, $order ) {
+				return \Newspack\Data_Events\Utils::get_order_data( $order_id, true );
```
leogermani: If I understand this correctly, I think we need to add some checks here to only fire this event if we can identify that this is a modal checkout interaction. Otherwise it will fire for all woo status order completed events

leogermani: also, this `get_order_data` has some issues apparently.. we have to look into it.

includes/tracking/class-data-events.php
```diff
@@ -0,0 +1,111 @@
...
+		if ( method_exists( 'Newspack\Donations', 'is_donation_product' ) ) {
+			if ( \Newspack\Donations::is_donation_product( $product_id ) ) {
+				$action_type = 'donation';
+				$product_type = 'donation';
+			}
+		}
+
+		// Check if it's associated with a membership and update product_type.
+		if ( self::is_membership_product( $product_id ) ) {
+			$product_type = 'membership';
```
leogermani: I'm gonna propose that we create a canonical method in the main plugin to get the product_type, so we have one source of truth on how we categorize products. More on this coming soon

includes/tracking/class-data-events.php
leogermani: WHen purchasing a normal product, I don't see the event being triggered. Probably because that method returns nothing.

When it's a donation, I see there's no action or action_type params. And there was an invalid param as well

`Parameter platform_data has an invalid value. It was removed from the np_modal_checkout_interaction event.`

src/modal-checkout/analytics/ga4/checkout-attempt.js
```diff
@@ -0,0 +1,25 @@
...
+ * @param {string} action Action name for the event: 'continue' or 'back'.
+ */
+
+export const manageCheckoutAttempt = () => {
+	if ( 'function' !== typeof window.gtag ) {
+		return;
+	}
+
+	const { action_type, amount, currency, product_id, product_type, recurrence, referer } = getProductDetails( 'modal-checkout-product-details' );
+	const payload = getEventPayload( 'checkout_attempt', {
```
leogermani: For consistency this action should be `form_submission`

includes/tracking/class-data-events.php
laurelfulford: Thanks for catching this, @leogermani!

Digging into it, it seems to be the `woocommerce_order_status_completed` that's not firing for the checkout button block, but I'm not sure why (I tested it by trying to `error_log` a string when it was fired). It seems to work fine for Donate; in testing `woocommerce_payment_complete` and `woocommerce_checkout_order_processed` both seemed to fire for both the checkout button block and donate block, but I'm not sure why (or if they're even in comparable spots -- I'm not sure either of them being triggered really means the same thing).

I'll keep digging into that and the `platform_data` bit, but if anything about the hook rings a bell, let me know!

src/modal-checkout/analytics/ga4/checkout-attempt.js
laurelfulford: Good call! This has been updated in [0fcf0a1](https://github.com/Automattic/newspack-blocks/pull/1819/commits/0fcf0a14ce09b79adef8407618665eafe400cccc).

src/blocks/donate/frontend/class-newspack-blocks-donate-renderer-base.php
laurelfulford: Eep, good catch! This looks like something I left behind when trying to pass information for the Donate block -- I've removed it in [3b1b4e1](https://github.com/Automattic/newspack-blocks/pull/1819/commits/3b1b4e1b2cacd1921cb15ba441b3fbfddc914a21).

includes/tracking/class-data-events.php
laurelfulford: In [a32e023](https://github.com/Automattic/newspack-blocks/pull/1819/commits/a32e023e6a316d498449697661b9bb397c156288) I updated this bit to use `woocommerce_checkout_order_processed` (so it fires for both the donate block and checkout button block) and added a check (possibly not a great approach) to check whether it's in a modal checkout. It seems to work but it could probably use some refining! 

Edited to add: I'll also dig into `woocommerce_order_status_completed` a bit more; there's another spot I was drawing my "this doesn't work for the checkout button block" from in general, but I'd like to dig into it more in case it's just me!

laurelfulford: Bah, this took a ridic amount of time for me to figure out, but I think the issue with `platform_data` is that I was supposed to be doing some more massaging of the data in the `handle_modal_checkout_interaction()` function for this event [here](https://github.com/Automattic/newspack-plugin/pull/3297/files#diff-3dbb459c03feab9c03a7a641babce543932c7899efccb35734cc1cbf275633e4R459-R467).

I'm working on that now, following some of the other examples there -- I originally followed the `handle_prompt_interaction()` function as an example, but in retrospect it makes sense that it didn't work as is, especially looking at the other `handle_*` functions in the file.

src/blocks/checkout-button/view.php
```diff
@@ -125,17 +132,39 @@ function render_callback( $attributes ) {
...
+			'recurrence'   => $recurrence,
+			'referer'      => substr( \get_permalink(), strlen( home_url() ) ), // TODO: Is this OK?
 		];
 
 		if ( ! $is_variable || $variation_id ) {
 			$product_data['price']                 = $price;
 			$product_data['product_price_summary'] = $product_price_summary;
 		}
 
+		if ( $varition_id ) {
```
dkoo:
```suggestion
		if ( $variation_id ) {
```

includes/tracking/class-data-events.php
```diff
@@ -0,0 +1,191 @@
...
+
+	/**
+	 * Returns an array of product information.
+	 *
+	 * @param string $product_id Product's ID.
+	 * @param array  $cart_item Cart item.
+	 * @param string $product_parent_id Product's ID when it has variations.
+	 */
+	public static function build_js_data_events( $product_id, $cart_item, $product_parent_id = '' ) {
+		// Reassign the IDs to make sure the product is the product and the variation is the variation.
```
dkoo, line 140: Assigning `$variation_id` with an empty string value helps avoid a `PHP Warning:  Undefined variable $variation_id in /newspack-repos/newspack-blocks/includes/tracking/class-data-events.php on line 154` when interacting with a non-variable product. We can also simplify the check of `$product_parent_id` slightly with just one `! empty()` check.

```suggestion
		// Reassign the IDs to make sure the product is the product and the variation is the variation.
		$variation_id = '';
		if ( ! empty( $product_parent_id ) ) {
			$variation_id = $product_id;
			$product_id   = $product_parent_id;
		}
```

includes/tracking/class-data-events.php
```diff
@@ -0,0 +1,191 @@
...
+		if ( '' !== $product_parent_id && 0 !== $product_parent_id ) {
+			$variation_id = $product_id;
+			$product_id = $product_parent_id;
+		}
+
+		$data_order_details = [
+			'amount'       => $cart_item['data']->get_price(),
+			'action_type'  => self::get_action_type( $product_id ),
+			'currency'     => \get_woocommerce_currency(),
```
dkoo: Probably want to be a bit defensive here just in case, since this global function is defined outside of this plugin:

```suggestion
			'currency'     => function_exists( 'get_woocommerce_currency' ) ? \get_woocommerce_currency() : 'USD',
```

src/blocks/checkout-button/view.php
```diff
@@ -112,6 +113,12 @@ function render_callback( $attributes ) {
 			$frequency = \WC_Subscriptions_Product::get_period( $product );
 		}
 
+		$product_type = 'product';
+		// Update the product type.
+		if ( method_exists( 'Newspack_Blocks\Tracking\Data_Events', 'get_product_type' ) ) {
```
dkoo: Unlike the defensive checks for `\Newspack\` methods or global functions defined by WooCommerce, there's no real reason we need to check for the existence of these methods since they're defined by this plugin (so if this code is running, we can be certain that the other code in the plugin exists too).

```suggestion
		$product_type = \Newspack_Blocks\Tracking\Data_Events::get_product_type( $product_id );
```

src/blocks/donate/frontend/class-newspack-blocks-donate-renderer-base.php
```diff
@@ -188,9 +187,17 @@ protected static function render_hidden_form_inputs( $attributes ) {
...
+
+		$donation_product_id = '';
+		if ( defined( '\Newspack\Donations::DONATION_PRODUCT_ID_OPTION' ) ) {
+			$donation_product_id = get_option( \Newspack\Donations::DONATION_PRODUCT_ID_OPTION, 0 );
+		}
+
 		wp_referer_field();
 		?>
 			<input type='hidden' name='newspack_donate' value='1' />
+			<input type='hidden' name='donation_currency' value='<?php echo esc_attr( \get_woocommerce_currency() ); ?>' />
```
dkoo: Probably want to check for the existence of this woo function before calling it, and store the value in a variable above where the HTML starts. Something like

```
$currency = function_exists( 'get_woocommerce_currency' ) ? \get_woocommerce_currency() : 'USD';

File: src/modal-checkout/analytics/ga4/checkout-attempt.js [hunk 10]
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @param {string} action Action name for the event: 'continue' or 'back'.
+ */
+
+export const manageCheckoutAttempt = () => {
+	if ( 'function' !== typeof window.gtag ) {
+		return;
+	}
+
+	const { action_type, amount, currency, product_id, product_type, recurrence, referer, variation_id } = getProductDetails( 'modal-checkout-product-details' );
```
dkoo: In general, when defining these variables from the output of `getProductDetails`, some will probably not be included in the output. So to guard against `undefined` values being sent, we can set defaults like this (using `variation_id` as an example):

```suggestion
	const { action_type, amount, currency, product_id, product_type, recurrence, referer, variation_id = '' } = getProductDetails( 'modal-checkout-product-details' );
```

includes/class-modal-checkout.php
```diff
@@ -273,6 +270,19 @@ function ( $item ) {
 				\wc_get_page_permalink( 'checkout' )
 			);
 
+			// Get data to send for this purchase.
+			$checkout_button_metadata = [
+				'currency'   => \get_woocommerce_currency(),
+				'amount'     => $price,
+				'product_id' => $product_id,
+				'referer'    => $referer,
```
dkoo: Nit: in general, unless we're literally calling a PHP or WP function that intentionally spells it `referer`, we should use the spelling `referrer` for consistency. This includes when using defining variable names as well as the param name in the payload.

```suggestion
				'referrer'    => $referer,
```

dkoo (changes requested):
@laurelfulford you've done great work on a big feature set here! I have some minor suggestions and nits on first pass here, but I think @leogermani's suggestion for standardizing product type and other product info is a good one.

To answer some of the open questions:

> The "opened" action fires twice for Checkout Button blocks set to use Variable Products that the reader can select. The initial time opens the variation picker, and the second time opens the modal checkout for the product that's picked. Should there be a different `action` param for these?

I think a separate action param is appropriate for this, since it's an additional step in the conversion flow. Maybe `opened_variations`?

> When applying a coupon, the `amount` doesn't change until the form is successfully submitted (`form_submission_success`). Should this change sooner?

I think this is okay for now, unless we want to try to fire an additional event when a coupon is successfully applied and extract the total then? That could get complicated, so maybe a separate PR if needed at all. Let's confirm with @kmwilkerson that it's needed

> When changing the price via the NYP field, the `amount` doesn't change until the form is successfully submitted (`form_submission_success`). Should this change sooner?

Same answer here—we could maybe do this with an additional event when the amount is updated, but not necessary for this PR until we confirm with @kmwilkerson whether it's needed.

> What `product_id` should be captured for the Donation products. Technically there's the main Donation product, plus there's a separate product for each frequency. Right now, it (incorrectly) uses both at random points.

I would suggest using the child product ID for each frequency, since that gives us more detailed info.

includes/class-modal-checkout.php
laurelfulford: D'oh! I remembered [this fix](https://github.com/Automattic/newspack-plugin/pull/2917) backwards 🤦‍♀️ Thank you!

laurelfulford:
Thank you for all this great feedback, @dkoo! I especially appreciate the "whys" -- working through the changes now! 🎉

laurelfulford:
I noticed while working on this that `loaded` fires when you try to submit the checkout with an error, too -- it should only fire when the modal is opened.

src/blocks/checkout-button/view.php
laurelfulford: 🤦‍♀️ This has been fixed!

includes/tracking/class-data-events.php
laurelfulford: Ah, makes sense! Added 🙂

includes/tracking/class-data-events.php
laurelfulford: Done! I added it here and the other spots I used `get_woocommerce_currency()` -- thanks!

src/blocks/checkout-button/view.php
laurelfulford: Makes sense! This has been removed, plus any others for methods coming from this plugin -- thanks!

src/blocks/donate/frontend/class-newspack-blocks-donate-renderer-base.php
laurelfulford: 👍

src/modal-checkout/analytics/ga4/checkout-attempt.js
laurelfulford: Ahhh good call! I've added this to this file, and the others. Thanks!

laurelfulford:
Okay, I think I got everything @dkoo!

> I think a separate action param is appropriate for this, since it's an additional step in the conversion flow. Maybe `opened_variations`?

Added, and added to the table above -- `opened_variations` fires when you open the variation picker step, and then `opened` fires when the modal checkout itself opens. 

When the variation is pre-selected and the reader isn't presented with the picker, just `opened` fires.

> I think this is okay for now, unless we want to try to fire an additional event when a coupon is successfully applied and extract the total then? That could get complicated, so maybe a separate PR if needed at all. Let's confirm with @kmwilkerson that it's needed

Left for now!

> Same answer here—we could maybe do this with an additional event when the amount is updated, but not necessary for this PR until we confirm with @kmwilkerson whether it's needed.

Also left for now!

> I would suggest using the child product ID for each frequency, since that gives us more detailed info.

Makes sense -- this has been updated!

> I noticed while working on this that loaded fires when you try to submit the checkout with an error, too -- it should only fire when the modal is opened.

This issue I commented about has also been fixed.

laurelfulford:
Thanks again for your feedback on this, @leogermani and @dkoo! 

I've marked this and the corresponding PR ready for review officially. I know we're still discussing `product_type` internally, and that can be updated in this code once it's finalized, but it would be really nice to have this in place for the beta testing to make sure the other information is tracking correctly, and can be used to build reports as expected. 

I updated a couple points in the description, one being that I removed the `console.log( payload );` from modal-checkout/analytics/ga4/utils/index.js. If needed for testing, it can be added before/after the `window.gtag( 'event', eventName, payload );` line.

dkoo:
@laurelfulford working nicely! I verified that every action, action_type, and parameter documented in the description is getting logged in GA:

Screenshot 2024-09-10 at 3 12 50 PM

I think this is ready to merge, with the caveat that we may need a follow-up PR to address this feedback which we're still discussing internally:

> I'm gonna propose that we create a canonical method in the main plugin to get the product_type, so we have one source of truth on how we categorize products. More on this coming soon

dkoo:
Also need to resolve conflicts with the base, but after that this is good to go

dkoo (approved):

dkoo (approved):

laurelfulford:
Thanks @dkoo! Happy to see the back of this one... 

> with the caveat that we may need a follow-up PR to address this feedback which we're still discussing internally.

... mostly 😆

PR #1823, opened 2024-08-02
fix(ras-acc): only trigger update_checkout event on validation

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1207817176293825/1207944985845149/f

This PR resolves an issue where the `update_checkout` woo event causes the order review table to end up in a blocked state. This looks to be happening because we are hiding the original order review table, cloning it, and rendering it in a non-standard part of the checkout. This seems to be happening at the same time as Woo attempts to block the table, perform some Woo logic, and then finally unblock the table and as a result the cloned table ends up in a blocked state.

We fix this by removing the call to trigger `update_checkout` event from the `setEditingDetails` method as well as from the cover fees script in https://github.com/Automattic/newspack-plugin/pull/3330 to allow updating checkout to be managed entirely within the `validateForm` function.

Before:
Screenshot 2024-08-02 at 14 17 23

After:
Screenshot 2024-08-02 at 14 14 42

You will need to also checkout the following Plugin PR before testing: https://github.com/Automattic/newspack-plugin/pull/3330

1. As a new reader with no saved billing details/payment information, trigger modal checkout via any donation block (recurring donation to guarantee the transaction details table appears)
2. Proceed through the checkout flow until you get to the payment information step. Scroll down to see the Transactions details table.
3. On `epic/ras-acc` this will be "blocked" (grayed out). On this branch it will not.
4. Smoke test the checkout modal in various other contexts (checkout buttons with tax/shipping/coupons/nyp/etc. setup, one-time donations, etc.)

### Other information:

### DIFF
--- a/src/modal-checkout/index.js
+++ b/src/modal-checkout/index.js
@@ -340,7 +340,7 @@ domReady(
 								$nyp.find( 'input[name="price"]' ).focus();
 								$nyp.find( 'h3, input[name="price"]' ).addClass( 'newspack-ui__field-error' );
 							}
-							$( document.body ).trigger( 'update_checkout' );
+							$( document.body ).trigger( 'update_checkout', { update_shipping_method: false } );
 						},
 						complete: () => {
 							unblockForm( $nyp );
@@ -369,10 +369,6 @@ domReady(
 				 * @param {boolean} isEditingDetails
 				 */
 				function setEditingDetails( isEditingDetails ) {
-					// Scroll to top.
-					window.scroll( { top: 0, left: 0, behavior: 'smooth' } );
-					// Update checkout.
-					$( document.body ).trigger( 'update_checkout' );
 					clearNotices();
 					// Clear checkout details.
 					$( '#checkout_details' ).remove();
@@ -409,6 +405,8 @@ domReady(
 						} );
 					}
 					$form.triggerHandler( 'editing_details', [ isEditingDetails ] );
+					// Scroll to top.
+					window.scroll( { top: 0, left: 0, behavior: 'smooth' } );
 				}
 
 				/**


laurelfulford (changes requested):
Thanks for tackling this one, @chickenn00dle! 

This works well for me when the modal is initially opened, but if you have two or more payment options, toggling between them whites it out again and it doesn't come even when you switch to the original payment method.

I'm marking as 'Request changes', but given that it's an improvement on the current behaviour, I'd also be good with splitting the toggling thing into a separate issue if it's gnarly. Just let me know!

chickenn00dle:
> This works well for me when the modal is initially opened, but if you have two or more payment options, toggling between them whites it out again and it doesn't come even when you switch to the original payment method.

Good catch @laurelfulford! Strangely, this seems to only be a problem for donation checkout. This doesn't happen for standard subscriptions on my end. My hunch is it has something to do with the cover transaction fee option we add for donations.

In any case, would you mind verifying this only happens for donations and only when the cover transaction fees option is set in Newspack > Reader Revenue > Stripe Settings on your end as well?

chickenn00dle:
I found the issue. We are adding a custom script whenever that cover transaction fee option is set which also triggers a call to `update_checkout` whenever payment method is updated: https://github.com/Automattic/newspack-plugin/blob/epic/ras-acc/src/other-scripts/wc-cover-fees/index.js

I'm still not sure why that event causes our order review table to end up in a blocked state though. Looking into it.

chickenn00dle:
Okay @laurelfulford, I think I got to the bottom of this!

I had to make an additional change in the Plugins repo to fix things: https://github.com/Automattic/newspack-plugin/pull/3330

I've updated this PR descriptions details and testing instructions, but this should be working for donations as well now.

Let me know if you are still running into problems.

laurelfulford (approved):
Awesome sleuthing, @chickenn00dle! 

The latest version fixed the issue I saw; I also tested the coupon and NYP submissions, and added some GST/PST tax options (which makes the transaction summary table show for products taxes are applied to), and didn't spot any issues 🙌

PR #1832, opened 2024-08-06
fix (ras-acc): improve error styles

laurelfulford:
### All Submissions:

This PR fixes some error styling in the modal checkout. It's a little strange because I haven't been able to consistently recreate the error -- it seems to do with multiple back-to-back transactions, but I'm not 100% sure. In the meantime, I've included testing steps that involve copy-pasting the HTML for the error into the checkout form to test the styles.

If it'd make more sense to test with the actual error, I can go back to hammering it to see if I can get better steps to recreate!

See: 1207817176293825-as-1207922679399315

1. Purchase a product or donation product using the modal checkout.
2. On the first screen (Billing Details), use the element inspector to edit the HTML, and add this HTML after the opening form tag:

```
<div class="woocommerce-NoticeGroup woocommerce-NoticeGroup-checkout newspack-ui__notice newspack-ui__notice--error">
<ul class="woocommerce-error" role="alert">
<li>
<div class="woocommerce-error">There was an error processing your order. Please check for any charges in your payment method and review your <a href="#">order history</a> before placing the order again.</div></li></ul></div>
```

4. Note the appearance of the error -- it looks like two errors nested in one, and there's an odd line break after the link (this is due to `display: flex` being used in these messages outside of the modal checkout to help position a button):

image

6. Repeat step 2 and confirm that the styles look okay:

image

### Other information:

### DIFF
--- a/src/modal-checkout/checkout.scss
+++ b/src/modal-checkout/checkout.scss
@@ -483,18 +483,19 @@
 	}
 
 	form.checkout .woocommerce-NoticeGroup {
-		ul.woocommerce-error {
-			background: none;
-			color: inherit;
-			list-style-type: none;
-			padding-left: 0;
-			padding-right: 0;
+		li div {
+			display: block; // Overrides display: flex from theme with listed errors.
 		}
 
 		.woocommerce-message,
 		.woocommerce-error,
 		.woocommerce-info {
+			background: none;
+			color: inherit;
 			font-size: var( --newspack-ui-font-size-xs, 14px );
+			list-style-type: none;
+			margin: 0;
+			padding: 0;
 		}
 	}
 


chickenn00dle (commented):
Style changes look good.

My only concern is errors like this are not recoverable. So the reader has to completely close the form and try again. However, modal checkout will still remain open and the reader has no way of knowing this. 

It would be great if we could create a general error state in the modal that hides the checkout entirely and asks the reader to close the form, refresh the page, and try again (or something along those lines)

Regardless, I'm gonna go ahead and approve this since this addresses the style issue.

chickenn00dle (approved):

laurelfulford:
Thanks @chickenn00dle! 

> It would be great if we could create a general error state in the modal that hides the checkout entirely and asks the reader to close the form, refresh the page, and try again (or something along those lines)

Agreed! I'm leaving the original issue open in Asana in hopes that I can figure out what's actually triggering it, then we can take it from there. I've run into it randomly while running through the checkout repeatedly, so I suspect it's something to do with too many orders in too short of time.

PR #1834, opened 2024-08-06
fix(ras-acc): update max-height for modals

laurelfulford:
### All Submissions:

This PR needs to be tested with https://github.com/Automattic/newspack-plugin/pull/3316.

Some mobile versions of Chrome and Safari (in my testing, on iPhone 13+) aren't interpreting our modal's max-height of `90vh` correctly -- in Chrome, the modal is getting cut off a bit, and in Safari, it's acting like it's `100vh`.

This PR switches the modal's max-height from `90vh` to `90%`, which seems to be more consistently respected.

1. Ideally test in an iPhone 13 or higher (my own iPhone 11 doesn't show the original issue)****
2. Try running the modal checkout in Safari and Chrome. Note the height of the modal:

3. Apply this PR and https://github.com/Automattic/newspack-plugin/pull/3316, and run `npm run build` for each.
4. Retest on your mobile device and confirm that the modal is now ~90% of the window height at the most. 

5. Test in desktop and try adjusting the browser window size, making it shorter and narrower. Confirm that the modal respects the 90% max-height.

**** I'm testing in Browserstack; if you don't have access to the right devices but are testing, please let me know!

### Other information:

### DIFF
--- a/src/modal-checkout/modal.scss
+++ b/src/modal-checkout/modal.scss
@@ -58,7 +58,7 @@
 				border-radius: 6px;
 				background: colors.$color__background-body;
 				max-width: 600px;
-				max-height: 90vh;
+				max-height: 90%;
 				min-height: 300px;
 				display: flex;
 				flex-direction: column;


src/modal-checkout/modal.scss
```diff
@@ -58,7 +58,7 @@
 				border-radius: 6px;
 				background: colors.$color__background-body;
 				max-width: 600px;
-				max-height: 90vh;
+				max-height: 90%;
```
chickenn00dle, line 61: Can we add a breakpoint for mobile devices so this change doesn't impact desktop?

chickenn00dle (changes requested):

chickenn00dle:
Works as expected, however the height of the variable modal is impacted on desktop as well. Left a comment about making the changes specific to mobile in this and the plugin PR. 

I'm not sure if we want to maintain the desktop height, so marking this as `Request changes`. Let me know if this isn't a problem though and I'll go ahead and change that to approve.

chickenn00dle (approved):

src/modal-checkout/modal.scss
laurelfulford: Followed up in the other PR for some reason, even though this one has all the info in it 🤦‍♀️ Not enough coffee! 

I haven't noticed a difference before and after this change myself (the 90vh was applied on desktop, too), but there may be a certain case where this is causing things to look different than in the epic/ras-acc branch? Just let me know!

laurelfulford:
Thanks for double-checking this, @chickenn00dle!

laurelfulford:
OMG I just realized this still had the Needs Design Review label 🤦‍♀️ I'll merge the other one and get these up on test sites so I can get @kevinzweerink to take a look there, and can make adjustments on top of that.

PR #1837, opened 2024-08-08
fix(ras-acc): merge trunk into epic/ras-acc

laurelfulford:
### All Submissions:

This PR is to merge trunk into the epic/ras-acc branch after the dependency updates.

1. Run `npm run build` and confirm there are no errors.
2. Spot check the updates and confirm that nothing looks out of place.

### Other information:


includes/class-modal-checkout.php
```diff
@@ -1478,7 +1478,7 @@ public static function get_summary_card_price_string( $name, $price = '', $frequ
 	 *
 	 * @return void.
 	 */
-	public static function maybe_add_checkout_registration_order_meta( $order ) {
+	public static function maybe_add_checkout_registration_order_meta( $order ) { // phpcs:ignore WordPressVIPMinimum.Hooks.AlwaysReturnInFilter.VoidReturn, WordPressVIPMinimum.Hooks.AlwaysReturnInFilter.MissingReturnStatement
```
laurelfulford: I wanted to flag this since I'm not sure ignoring these sniffs is the preferred way to fix these errors.

laurelfulford:
@dkoo Roping you in on this one, too 🙂 Just let me know if I should be looping in anyone else instead!

includes/class-modal-checkout.php
dkoo: It looks like this hook is actually an [action, not a filter](https://woocommerce.github.io/code-reference/files/woocommerce-includes-class-wc-checkout.html#source-view.449). Updated the `add_action` call in 4399227 and that means we can remove the `// phpcs:ignore` statement as well.

dkoo (approved):
@laurelfulford the changes look and work well for me, so if 4399227 looks good to you, this can be merged.

laurelfulford:
Ah, thanks for catching that, @dkoo! I wrongly assumed that the sniff was being to sensitive since this wasn't an issue before.

This change looks good to me, so I think this is good to go!

PR #1855, opened 2024-08-26
fix(ras-acc): don't close modal when redirecting

laurelfulford:
### All Submissions:

This PR, along with https://github.com/Automattic/newspack-plugin/pull/3369, prevent the last step of the modal checkout from closing if the next step is redirecting to a new page/previous page. This will hopefully reduce confusion/make the transition a little smoother.

Note: It's entirely possible there's a more elegant way to handle all or parts of this; I'd love some feedback on that front if you have any! 

See 1207817176293825-as-1207817176293828

1. Start with a test site running epic/ras-acc, and with "Present newsletter signup after checkout and registration" disabled under Engagement > Reader Activation > Advanced Settings.
2. Set up at least the following blocks:
     * A checkout button block with no after checkout behaviour set.
     * A checkout button block with a redirect to a different page after checkout.
     * A checkout button block that goes back to the previous page after checkout.
     * A donate block with no after checkout behaviour set.
     * A donate block with a redirect to a different page after checkout.
     * A donate block that goes back to the previous page after checkout.
3. Test running checkouts with each, for the blocks with the redirect, note that the modal closes before the redirect happens.
4. Apply this PR and https://github.com/Automattic/newspack-plugin/pull/3369, and run `npm run build` for each.
5. Repeat step 2; confirm that the modal stays open while the redirect happens, but closes as expected when not.
6. Navigate to Engagement > Reader Activation > Advanced Settings and enable "Present newsletter signup after checkout and registration".
7. Repeat step 2, and confirm the modal stays open while the redirect happens, but closes as expected when not.
8. Repeat step 2 on at least one block, but sign up for all newsletters on one of the block.
9. Repeat step 2 again and confirm that the newsletter modal isn't shown, and the thank you modal stays open while the redirect happens. 
10. Smoke test other screens, like triggering the newsletter sign up when you Sign Up to the site, and running checkouts as a new reader, and confirm there aren't any issues.

### Other information:

### DIFF
--- a/src/modal-checkout/modal.js
+++ b/src/modal-checkout/modal.js
@@ -57,18 +57,9 @@ domReady( () => {
 		const hasNewsletterPopup = document?.querySelector( '.newspack-newsletters-signup-modal' );
 
 		// We want to block closing the modal if redirecting elsewhere:
-		let shouldCloseModal = true;
+		const shouldCloseModal = ! afterSuccessUrlInput || ! afterSuccessBehaviorInput || ! container?.checkoutComplete;
 
-		if (
-			afterSuccessUrlInput &&
-			afterSuccessBehaviorInput &&
-			container?.checkoutComplete &&
-			! hasNewsletterPopup
-		) {
-			shouldCloseModal = false;
-		}
-
-		if ( shouldCloseModal ) {
+		if ( shouldCloseModal || hasNewsletterPopup ) {
 			spinner.style.display = 'flex';
 			if ( iframe && modalContent.contains( iframe ) ) {
 				// Reset iframe and modal content heights.
@@ -112,7 +103,7 @@ domReady( () => {
 			if ( window?.newspackReaderActivation?.openNewslettersSignupModal ) {
 				window.newspackReaderActivation.openNewslettersSignupModal( {
 					callback: handleCheckoutComplete,
-					skipClose: ( afterSuccessUrlInput && afterSuccessBehaviorInput ) ? true : false,
+					closeOnSuccess: shouldCloseModal,
 				} );
 			} else {
 				handleCheckoutComplete();
--- a/src/modal-checkout/modal.js
+++ b/src/modal-checkout/modal.js
@@ -54,14 +54,16 @@ domReady( () => {
 		const afterSuccessBehaviorInput = container.querySelector(
 			'input[name="after_success_behavior"]'
 		);
+		const hasNewsletterPopup = document?.querySelector( '.newspack-newsletters-signup-modal' );
 
 		// We want to block closing the modal if redirecting elsewhere:
 		let shouldCloseModal = true;
+
 		if (
 			afterSuccessUrlInput &&
 			afterSuccessBehaviorInput &&
 			container?.checkoutComplete &&
-			! window?.newspackReaderActivation?._openNewslettersSignupModal
+			! hasNewsletterPopup
 		) {
 			shouldCloseModal = false;
 		}
--- a/src/modal-checkout/modal.js
+++ b/src/modal-checkout/modal.js
@@ -49,40 +49,51 @@ domReady( () => {
 
 	const initialHeight = modalContent.clientHeight + spinner.clientHeight + 'px';
 	const closeCheckout = () => {
-		spinner.style.display = 'flex';
-
 		const container = iframe?.contentDocument?.querySelector( `#${ IFRAME_CONTAINER_ID }` );
+		const afterSuccessUrlInput = container.querySelector( 'input[name="after_success_url"]' );
+		const afterSuccessBehaviorInput = container.querySelector(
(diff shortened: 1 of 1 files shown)


src/modal-checkout/modal.js
```diff
@@ -49,40 +49,53 @@ domReady( () => {
...
-			modalContent.removeChild( iframe );
-			modalContent.style.height = initialHeight;
+		const afterSuccessUrlInput = container.querySelector( 'input[name="after_success_url"]' );
+		const afterSuccessBehaviorInput = container.querySelector(
+			'input[name="after_success_behavior"]'
+		);
+		const hasNewsletterPopup = document?.querySelector( '.newspack-newsletters-signup-modal' );
+
+		// We want to block closing the modal if redirecting elsewhere:
+		let shouldCloseModal = true;
```
chickenn00dle: Seems like we can just define this directly using the various checks for after success params:

```suggestion
		const shouldCloseModal = ! afterSuccessUrlInput && ! afterSuccessBehaviorInput;
```

src/modal-checkout/modal.js
```diff
@@ -49,40 +49,53 @@ domReady( () => {
...
+
+		if (
+			afterSuccessUrlInput &&
+			afterSuccessBehaviorInput &&
+			container?.checkoutComplete &&
+			! hasNewsletterPopup
```
chickenn00dle: IMO, we should only use the after success params to determine the value of `shouldCloseModal`. The other two (`checkoutComplete` and `hasNewsletterPopup`) seem to be for specific cases and so we can do things like move some of the logic within the `if ( container?.checkoutComplete ) ...` block below, or add the check specifically where it's needed. 

This makes it very clear why the `shouldCloseModal` exists and makes it more reusable throughout the code.

src/modal-checkout/modal.js
```diff
@@ -49,40 +49,53 @@ domReady( () => {
...
+		) {
+			shouldCloseModal = false;
 		}
 
-		if ( iframeResizeObserver ) {
-			iframeResizeObserver.disconnect();
-		}
+		if ( shouldCloseModal ) {
```
chickenn00dle: If you decide to go with the strategy above for `shouldCloseModal`:

```suggestion
		if ( shouldCloseModal || hasNewslettersPopup ) {
```

src/modal-checkout/modal.js
```diff
@@ -95,16 +108,21 @@ domReady( () => {
 				}
 				window?.newspackReaderActivation?.resetCheckoutData?.();
 			};
+
 			if ( window?.newspackReaderActivation?.openNewslettersSignupModal ) {
 				window.newspackReaderActivation.openNewslettersSignupModal( {
 					callback: handleCheckoutComplete,
+					skipClose: ( afterSuccessUrlInput && afterSuccessBehaviorInput ) ? true : false,
```
chickenn00dle: If you decide to go with the strategy above for `shouldCloseModal`:

```suggestion
					skipClose: ! shouldCloseModal,
```

chickenn00dle (approved):
Thanks for tackling this @laurelfulford!

This is working as expected and the redirect flow looks much better!

I do have some nitpicks about some of the code, however since it's pretty subjective and this is working well, I'll give this a thumbs up and let you decide whether to make the changes or not.

src/modal-checkout/modal.js
laurelfulford: Thanks @chickenn00dle! The only issue I found with this specific change is that, without the `container?.checkoutComplete` check, it could block the modal from being closed by the 'Close' button, too. 

I made a modified version of this feedback in [38b6d13](https://github.com/Automattic/newspack-blocks/pull/1855/commits/38b6d13fa73348a78637420dd280455d66215320) -- could I please get your eyes on this one more time to make sure I'm not missing something?

src/modal-checkout/modal.js
laurelfulford: Good call on the `hasNewsletterPopup` bit -- this has been moved!

src/modal-checkout/modal.js
laurelfulford: Ah this is much nicer!

laurelfulford:
@chickenn00dle I'm going to hold off merging this; I made most of the nit changes since they're definitely improvements (thanks for that!) but one is slightly modified to prevent an issue with closing the modal before the checkout was done. When you have a moment can I get your eyes on this one more time to make sure it makes sense? Thanks!

chickenn00dle (approved):
> one is slightly modified to prevent an issue with closing the modal before the checkout was done. When you have a moment can I get your eyes on this one more time to make sure it makes sense?

Good catch! Your changes make sense. Giving this another thumbs up 👍

laurelfulford:
Awesome! Thank you for the extra-extra double check, @chickenn00dle! 🎉

PR #1866, opened 2024-08-31
fix(ras-acc): add check for if RAS is enabled to allow checkout

laurelfulford:
### All Submissions:

This PR makes it possible to use the modal checkout when RAS isn't enabled, to match the current behaviour on trunk.

The main difference between the RAS-based checkout is that the initial checkout screen will require you to fill our the email address with the other billing details, since it won't be pre-populated from the Sign In/Sign Up screen.

See: 1207817176293825-as-1208164200882316

1. Start on a test site without RAS enabled, or disable it by running `wp option delete newspack_reader_activation_enabled`.
2. Set up three blocks: a regular Checkout Button block, one with a variable product, and a Donate block.
3. On epic/ras-acc in an incognito window, try to run the checkout in any of the blocks; only the initial variation picker will open.
5. In the incognito window, run through each checkout and confirm it works.
6. Enable RAS (run `wp option update newspack_reader_activation_enabled 1` or use the wizard), or re-test this PR on a second test site with RAS enabled.
7. Smoke test the checkout flows as a new reader and as a logged in reader and confirm that you get the Sign In prompts at the appropriate points, and that the checkout works as usual.

### Other information:


laurelfulford:
Switching this back to draft mode -- looks like it's not working as I'd hoped!

Edited to add: It was a variable name change at the last moment 🤦‍♀️ Working now!

chickenn00dle (changes requested):
Hmm. It feels strange that the openAuthModal method would be available when ras is not enabled. I think a cleaner solution might be to move this logic into the main plugin and make sure ras is enabled before making the `openAuthModal` function available. This would also resolve things in blocks since when that function is not present, modal checkout will just open normally.

WDYT?

laurelfulford:
Oh yeah, that would be nicer 😄 I'll dig into that tomorrow!

laurelfulford:
Replaced with https://github.com/Automattic/newspack-plugin/pull/3396, and closing -- thanks @chickenn00dle!

PR #1869, opened 2024-09-04
fix: support new is_my_account() method from main plugin

dkoo:
### All Submissions:

Currently, transactions completed from the "My Account" dashboard are not using the modal checkout flow. However, due to a fix in #1573 for express checkout flows, when completing a transaction from My Account that's also related to a subscription that originated with a modal checkout flow, these transactions are flagged as My Account transactions and show the modal checkout thank you template, but in a non-modal context, trapping the user in a dead-end.

This fix relies on a new method in https://github.com/Automattic/newspack-plugin/pull/3389 that lets us determine if a checkout flow originated from My Account, and show the correct non-modal thank you template if so.

The better solution would be to use the modal checkout flow for My Account transactions too, but this is a bigger change than is appropriate for a hotfix.

See testing instructions in https://github.com/Automattic/newspack-plugin/pull/3389.

### Other information:

### DIFF
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -124,34 +124,37 @@ final class Modal_Checkout {
 	 * Change the post-transaction return URL.
 	 * This is specifically for non-redirect-based flows, such as Apple Pay.
 	 *
-	 * @param array $result The return payload for a successfull transaction.
+	 * @param array $result The return payload for a successful transaction.
 	 */
 	public static function woocommerce_payment_successful_result( $result ) {
 		$order_id           = $result['order_id'];
 		$order              = \wc_get_order( $order_id );
 		$modal_checkout_url = $order->get_meta( '_newspack_modal_checkout_url' );
-		if ( empty( $modal_checkout_url ) ) {
+		if (
+			empty( $modal_checkout_url ) ||
+			(
+				// Until we use the modal checkout flow from My Account, we don't want to show the modal checkout thank you template for checkouts originating from My Account.
+				method_exists( 'Newspack\WooCommerce_My_Account', 'is_from_my_account' ) && \Newspack\WooCommerce_My_Account::is_from_my_account()
+			)
+		) {
 			return $result;
 		}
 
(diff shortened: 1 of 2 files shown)


includes/class-modal-checkout.php
```diff
@@ -124,34 +124,37 @@ public static function woocommerce_checkout_create_order_line_item( $item, $cart
...
 	public static function woocommerce_payment_successful_result( $result ) {
 		$order_id           = $result['order_id'];
 		$order              = \wc_get_order( $order_id );
 		$modal_checkout_url = $order->get_meta( '_newspack_modal_checkout_url' );
-		if ( empty( $modal_checkout_url ) ) {
+		if (
+			empty( $modal_checkout_url ) ||
+			(
+				// Until we use the modal checkout flow from My Account, we don't want to show the modal checkout thank you template for checkouts originating from My Account.
+				method_exists( 'Newspack\WooCommerce_My_Account', 'is_from_my_account' ) && \Newspack\WooCommerce_My_Account::is_from_my_account()
```
adekbadek: When following the testing instructions, the successful result is never referred from my-account, but from /checkout. That might be why I'm not seeing what's expected here: https://github.com/Automattic/newspack-plugin/pull/3389#pullrequestreview-2279254728

dkoo: The hidden field [added here](https://github.com/Automattic/newspack-plugin/pull/3389/files#diff-74c4e95b1dc2575ce46450e3fa231d01a291e298a72ed7656a6271757eb585afR500-R505) should account for that. However, it won't work for express checkouts, and I'm not sure how we might pass that info along to express checkout requests...

dkoo: Fixed by [146917f](https://github.com/Automattic/newspack-plugin/commit/146917fb25b1b36368fe92f50adc9c2574330f6f)

dkoo:
Marking approved based on @adekbadek's [approval in the other PR](https://github.com/Automattic/newspack-plugin/pull/3389#pullrequestreview-2302561432).

ronchambers:
New bug posted here: https://github.com/Automattic/newspack-plugin/issues/3426

PR #1870, opened 2024-09-04
fix(ras-acc): undo bad merge updates

chickenn00dle:
### All Submissions:

* [ ] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-theme/blob/master/.github/CONTRIBUTING.md)?
* [ ] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [ ] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

This PR undoes a small change that occured during a bad merge on my part.

Specifically, a few lines that were removed in https://github.com/Automattic/newspack-blocks/pull/1823 were left in the merge.

### How to test the changes in this Pull Request:

Visually verify the code. You can also see https://github.com/Automattic/newspack-blocks/pull/1823 for testing instructions if you'd like to test more thoroughly.

### Other information:

* [ ] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #1872, opened 2024-09-04
fix (ras-acc): make processing class unique to avoid clashes

laurelfulford:
### All Submissions:

This PR attempts to fix an issue where a subscription product with a free trial won't let you advance to the second screen of the modal checkout, because the 'Continue' button is greyed out. 

My best guess is that something else (WooCommerce?) is removing the `processing` class before [this check](https://github.com/Automattic/newspack-blocks/blob/epic/ras-acc/src/modal-checkout/index.js#L657), causing it to fail and not finish the "unblock" process -- it's only happening with a very specific product setting, though, so there may be more to it than that. 

See 1207817176293825-as-1208214895993707

1. Set up a few different checkout button blocks, plus a donate block for comparison. Make sure one of the checkout button blocks links to a subscription product with a free trial. 
2. On epic/ras-acc, run through a few trial checkouts with your buttons; for the subscription with a trial, note that the 'Continue' button is greyed out:

CleanShot 2024-09-04 at 14 20 31

4. Confirm that the subscription product with a free trial can be purchased, and that you can make it past the first screen.
5. Smoke test the other checkouts (including using the NYP and coupon forms) and confirm the "processing" behaviour still works as expected.

### Other information:

### DIFF
--- a/src/modal-checkout/checkout.scss
+++ b/src/modal-checkout/checkout.scss
@@ -253,7 +253,7 @@
 	// Processing styles
 	.modal_checkout_nyp,
 	.modal_checkout_coupon {
-		&.processing {
+		&.modal-processing {
 			opacity: 0.5;
 		}
 	}
@@ -558,7 +558,7 @@
 		margin-top: 0;
 	}
 
-	.processing {
+	.modal-processing {
 		opacity: 0.5;
 	}
 }
--- a/src/modal-checkout/index.js
+++ b/src/modal-checkout/index.js
@@ -635,14 +635,14 @@ domReady(
 				 * @return {boolean} Whether the form was blocked or not.
 				 */
 				function blockForm( form ) {
-					if ( form.is( '.processing' ) ) {
+					if ( form.is( '.modal-processing' ) ) {
 						return false;
 					}
 					const buttons = form.find( 'button[type=submit]' );
 					buttons.each( ( i, el ) => {
 						$( el ).attr( 'disabled', true );
 					} );
-					form.addClass( 'processing' );
+					form.addClass( 'modal-processing' );
 					return true;
 				}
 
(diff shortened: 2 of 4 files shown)


chickenn00dle (commented):
Hey @laurelfulford, 

Strangely, I can't reproduce on `epic/ras-acc`:

Screenshot 2024-09-05 at 09 03 50


I did just merge in https://github.com/Automattic/newspack-blocks/pull/1870/files, so I'm wondering if that extra `update_checkout` trigger might have something to do with this issue?

Are you still seeing the issue with the latest from `epic/ras-acc`?

laurelfulford:
Ah weird, @chickenn00dle! I can still reproduce it -- I wonder if it's something specific I'm doing when I set up subscriptions? I set up the checkout button that's still showing the issue on the ras-acc test site too 🤔 

It's an absolute shot in the dark, but are you doing anything wildly different with your subscription product setup? I have NO idea what would cause the issue but ¯\_(ツ)_/¯ 

CleanShot 2024-09-04 at 17 35 16

The other thing I noticed is that the Google Pay express checkout button won't show when the Continue Button is greyed out, too. I'll dig into that and see if it's somehow related!

laurelfulford:
Ahhhh, I think it's having [WooPayments](https://wordpress.org/plugins/woocommerce-payments/) installed and activated (even if it's not enabled as the payment source). 

@chickenn00dle When you have a moment, could you try enabling that plugin and see if you get the Continue weirdness? Thanks!

chickenn00dle (approved):
> @chickenn00dle When you have a moment, could you try enabling that plugin and see if you get the Continue weirdness? Thanks!

Strange, I am still not able to reproduce with WooPayments active, both when it is enabled as a payment gateway AND when it is not. 

That said, I am noticing the issue on `ras-acc.newspackstaging.com` and there is no harm in using a more specific classname generally speaking, so I'm gonna go ahead and approve this one anyway.

laurelfulford:
Thanks @chickenn00dle! Thank you for retesting it -- I was so sure I had the cause figured out, I wonder what's going on 🤔 

I'm going to merge, but let me know if you think it'd be "tidier" to remove the `processing` class check from the `unblockForm()` function. I'm not sure if that would introduce other unintended side effects, but part of me is (irrationally) bothered by adding another processing class 😆

PR #1881, opened 2024-09-13
fix (ras-acc): prevent NYP, coupon submission from clearing credit card fields

laurelfulford:
### All Submissions:

This PR moves setting and updating the 'Complete Transaction' button price from PHP to JS. This is to prevent an issue where submitting a coupon or NYP after filling in the credit card fields would cause the credit card fields to be cleared.

See: 1207817176293825-as-1208255803261156.

1. Start on epic/ras-acc.
2. Enable coupons, and set up at least one regular coupon.
3. Set up Checkout button blocks for a simple product and a NYP product, plus a donate block.
4. Set up Stripe Gateway and WooPayments as your payment options.
5. As a reader, run through a transaction using the simple product.
6. On the second screen, fill out your payment information first. 
7. Fill in your coupon, and Apply.
8. Note that the credit card fields reset.
9. Repeat steps 5-8 with the NYP product, and with both payment gateways.
10. Apply this PR and run `npm run build`.
11. Repeat steps 5-8 with both your simple and NYP products, and both payment processors. When you do this, confirm that the credit card values remain in place, and also confirm that the Confirm Transaction button at the bottom updates with the correct price. 
12. Try applying your coupon, then removing it, and checking the price in the Confirm Transaction button each time. 
13. Run a transaction with the Donate Button block, and confirm the button on the second screen just says "Donate Now". 
14. Smoke test a few other product options, like a subscription product or variable product, and make sure the Confirm Transaction button displays the correct initial payment amount.

### Other information:

### DIFF
--- a/src/modal-checkout/index.js
+++ b/src/modal-checkout/index.js
@@ -334,7 +334,6 @@ domReady(
 						},
 					} );
 				}
-
 				if ( $coupon.length ) {
 					$coupon.on( 'submit', handleCouponSubmit );
 					$( document.body ).on( 'removed_coupon_in_checkout', () => {
@@ -388,7 +387,6 @@ domReady(
 								$nyp.find( 'h3, input[name="price"]' ).addClass( 'newspack-ui__field-error' );
 							}
 							$( document.body ).trigger( 'update_checkout', { update_shipping_method: false } );
-
 						},
 						complete: () => {
 							unblockForm( $nyp );
--- a/src/modal-checkout/index.js
+++ b/src/modal-checkout/index.js
@@ -129,6 +129,48 @@ domReady(
 				} );
 
 				/**
+				 * Get updated cart total to update the "Place Order" button.
+				 */
+				function returnUpdatedCartTotal() {
+					let cartTotal;
+					$.ajax({
+						url: newspackBlocksModalCheckout.ajax_url,
+						method: 'POST',
+						async: false,
+						data: {
+							action: 'get_cart_total',
+						},
+						success: (response)=>{
+							cartTotal = response;
+						}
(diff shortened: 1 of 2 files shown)


src/modal-checkout/index.js
```diff
@@ -128,6 +128,48 @@ domReady(
 					$( '.payment_methods' ).after( $el );
 				} );
 
+				/**
+				 * Get updated cart total to update the "Place Order" button.
+				 */
+				function returnUpdatedCartTotal() {
```
chickenn00dle: non-blocking nit: might be clearer to name this something like `getUpdatedCartTotal`

chickenn00dle (approved):
Nice work @laurelfulford! Looks good for the most part. Left a small non-blocking nit.

I also ran into a very edge case issue where when a product has both coupons AND nyp enabled, submitting a new NYP value will override the coupon entirely. However, this seems to be happening on `epic/ras-acc` as well so will approve and [create a new task for this](https://app.asana.com/0/1207817176293825/1208344897866336/f).

laurelfulford:
> I also ran into a very edge case issue where when a product has both coupons AND nyp enabled, submitting a new NYP value will override the coupon entirely.

Eeeeep good catch! 😱 I ran _so_ many checkouts testing this and still missed it -- so glad you caught it!

src/modal-checkout/index.js
laurelfulford: Good call! This has been updated in [013161e](https://github.com/Automattic/newspack-blocks/pull/1881/commits/013161edda5b51a1aa8d32e4f0b5daafd5b50142).

PR #1896, opened 2024-10-08
fix(modal-checkout): optimize iframe load and refactor anonymous cart generation

miguelpeixe:
### All Submissions:

This PR addresses a few different issues:

Introduces a new strategy to detect iframe load so it can be ready sooner. The `load` event dispatch waits for render-blocking assets to load before firing. The new strategy watches for the iframe content document to trigger when the DOM is ready.

The dequeuing strategy is now reversed so that only whitelisted assets can be enqueued. This prevents unnecessary assets from loading.

#### Cart generation via ajax

To handle an edge case in which carts are getting wiped between an account registration and the modal checkout render, this PR changes the strategy for anonymous readers.

Carts should now be generated in parallel to the auth/account creation flow. This allows for the cart session to be created before the authentication, which, once authenticated, Woo should handle the transition without issues.

Because of this new strategy, readers signing in via link should also be able to restore the checkout flow with the existing cart. The pending checkout logic has been simplified here and in https://github.com/Automattic/newspack-plugin/pull/3471.

1. Checkout this branch and https://github.com/Automattic/newspack-plugin/pull/3471
2. In a fresh session, start the checkout button flow with a subscription product
3. Register a new reader and confirm you are able to purchase without issues
4. In another fresh session, start the checkout button flow with the same product
5. This time, enter an existing email and click to sign in via link
6. Paste the link in your session and confirm you are authenticated and able to proceed with the checkout
7. Repeat steps 2-6 but through the Donate block

### Other information:

### DIFF
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -160,9 +160,8 @@ final class Modal_Checkout {
 		}
 
 		$is_newspack_checkout = filter_input( INPUT_GET, 'newspack_checkout', FILTER_SANITIZE_NUMBER_INT );
-		$is_newspack_donate   = filter_input( INPUT_GET, 'newspack_donate', FILTER_SANITIZE_NUMBER_INT );
 
-		if ( ! $is_newspack_checkout && ! $is_newspack_donate ) {
+		if ( ! $is_newspack_checkout ) {
 			return;
 		}
 
@@ -292,9 +291,14 @@ final class Modal_Checkout {
 			 */
 			\do_action( 'newspack_blocks_checkout_button_modal', $checkout_button_metadata );
 
-			if ( ! defined( 'DOING_AJAX' ) ) {
+			$checkout_url = apply_filters( 'newspack_blocks_checkout_url', $checkout_url );
+
+			if ( defined( 'DOING_AJAX' ) ) {
+				echo wp_json_encode( [ 'url' => $checkout_url ] );
+				exit;
+			} else {
 				// Redirect to checkout.
-			\wp_safe_redirect( apply_filters( 'newspack_blocks_checkout_url', $checkout_url ) );
+				\wp_safe_redirect( $checkout_url );
 				exit;
 			}
 		}
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -52,7 +52,11 @@ final class Modal_Checkout {
 	 * Initialize hooks.
 	 */
 	public static function init() {
-		add_action( 'wp_loaded', [ __CLASS__, 'process_checkout_request' ], 5 );
+		// Process checkout request.
+		add_action( 'wp_ajax_modal_checkout_request', [ __CLASS__, 'process_checkout_request' ] );
+		add_action( 'wp_ajax_nopriv_modal_checkout_request', [ __CLASS__, 'process_checkout_request' ] );
+		add_action( 'wp', [ __CLASS__, 'process_checkout_request' ] );
+
 		add_filter( 'wp_redirect', [ __CLASS__, 'pass_url_param_on_redirect' ] );
 		add_filter( 'woocommerce_cart_product_cannot_be_purchased_message', [ __CLASS__, 'woocommerce_cart_product_cannot_be_purchased_message' ], 10, 2 );
 		add_filter( 'woocommerce_add_error', [ __CLASS__, 'hide_expiry_message_shop_link' ] );
@@ -151,22 +155,23 @@ final class Modal_Checkout {
 	 * Process checkout request for modal.
 	 */
 	public static function process_checkout_request() {
-		if ( is_admin() ) {
+		if ( is_admin() && ! defined( 'DOING_AJAX' ) ) {
 			return;
 		}
 
-		$is_newspack_checkout       = filter_input( INPUT_GET, 'newspack_checkout', FILTER_SANITIZE_NUMBER_INT );
-		$is_newspack_donate         = filter_input( INPUT_GET, 'newspack_donate', FILTER_SANITIZE_NUMBER_INT );
+		$is_newspack_checkout = filter_input( INPUT_GET, 'newspack_checkout', FILTER_SANITIZE_NUMBER_INT );
+		$is_newspack_donate   = filter_input( INPUT_GET, 'newspack_donate', FILTER_SANITIZE_NUMBER_INT );
+
+		if ( ! $is_newspack_checkout && ! $is_newspack_donate ) {
+			return;
+		}
+
 		$product_id                 = filter_input( INPUT_GET, 'product_id', FILTER_SANITIZE_NUMBER_INT );
 		$variation_id               = filter_input( INPUT_GET, 'variation_id', FILTER_SANITIZE_NUMBER_INT );
 		$after_success_behavior     = filter_input( INPUT_GET, 'after_success_behavior', FILTER_SANITIZE_SPECIAL_CHARS );
 		$after_success_url          = filter_input( INPUT_GET, 'after_success_url', FILTER_SANITIZE_URL );
 		$after_success_button_label = filter_input( INPUT_GET, 'after_success_button_label', FILTER_SANITIZE_SPECIAL_CHARS );
 		$is_checkout_registration   = filter_input( INPUT_GET, self::CHECKOUT_REGISTRATION_FLAG, FILTER_SANITIZE_NUMBER_INT );
-
-		if ( ! $is_newspack_checkout && ! $is_newspack_donate ) {
-			return;
-		}
 
 		// Flag the checkout as a registration.
 		if ( $is_checkout_registration ) {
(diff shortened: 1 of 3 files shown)


includes/class-modal-checkout.php
```diff
@@ -622,9 +622,53 @@ public static function dequeue_scripts() {
...
+			'newspack-wc',
+			'newspack-ui',
+			'newspack-style',
+			'newspack-recaptcha',
+			// Woo.
+			'woocommerce',
+			'WCPAY',
+			'Woo',
+			'wc-',
+			'wc_',
```
chickenn00dle, line 657: Its not uncommon for woo extensions to utilize these prefixes for scripts, so we probably wouldn't be fully covered. That said this is definitely a more thorough approach and better than what we have now.

src/modal-checkout/modal.js
```diff
@@ -39,6 +39,109 @@ domReady( () => {
...
+				clearTimeout( iframe._readyTimer );
+				cb.call( this );
+			}
+		}
+		function readyState() {
+			if ( this.readyState === "complete" ) {
+				ready.call( this );
+			}
+		}
+		function addEvent( elem, event, fn ) {
```
chickenn00dle: Question: Is the reason we need to generalize adding the event listener like this related to scope?

chickenn00dle (commented):
Nice! Very cool way to decouple the iframe logic from the load event. I like this approach, although it doesn't seem to take into account the transition between completing checkout and loading the thank you view. 

Screenshot 2024-10-08 at 12 25 21

I think we will need to unflag the ready state during this transition so the handler doesn't return early. Or maybe implement a new flag that triggers the complete state separately?

miguelpeixe:
> doesn't seem to take into account the transition between completing checkout and loading the thank you view.

 Oops! Thank you for catching that. That was actually because the newsletter subscription script wasn't whitelisted to be enqueued. Fix in f0972b7345e95b11b11365500e162b0225062de2

Also, the iframe resize observer failed to resize the modal after the checkout because of the new `iframe._ready` checks. I tweaked it a bit in f55c4bb4a739af8d4a6baaf1812fac0635892745 and should work now.

src/modal-checkout/modal.js
miguelpeixe: It's meant to support <= IE8. The `addEvent` calls both `addEventListener` and the legacy `attachEvent`, but we're not doing that anywhere else so I removed in f55c4bb4a739af8d4a6baaf1812fac0635892745

miguelpeixe:
25f6118f9e95b507e410f721ab4a98a4d5269c3a removes the `init_checkout` listener and runs the `init()` directly. The Woo trigger is called on the main thread, whereas our listener is attached asynchronously (both async script tag and after `domReady`). This causes a race condition and the modal to eventually never leave the loading state.

It's safe to run the logic outside the trigger. To keep the diff reasonable, I'm maintaining the code inside an `init()` function, but we might want to remove that.

You can reproduce the race condition by throttling your network to 3G and disabling the cache before starting the modal while on the epic branch.

chickenn00dle (commented):
Thanks for the changes @miguelpeixe!

I am running into a new issue where stripe throws the following exception when removing the init logic from the `init_checkout` event, which causes modal checkout to be stuck in a blocked state on the payment gateway step:

```
{
  "message": "We could not retrieve data from the specified Element.\n              Please make sure the Element you are attempting to use is still mounted."
}
```
This only happens when I've throttled my network speed to 3G and when going through the auth + checkout flow as a new reader, so maybe another race condition between Woo payment gateway logic and our init method? Not sure.

miguelpeixe:
I found another race condition issue with the domReady callback. It's an immediately invoked function that will not run as a domReady callback. Updated in 7dcfc124c9824c791b85ddc64fef1aa979d0cee1

includes/class-modal-checkout.php
```diff
@@ -151,23 +155,23 @@ public static function dequeue_woocommerce_styles( $enqueue_styles ) {
...
+		}
+
+		$is_newspack_checkout = filter_input( INPUT_GET, 'newspack_checkout', FILTER_SANITIZE_NUMBER_INT );
+
+		if ( ! $is_newspack_checkout ) {
 			return;
 		}
 
-		$is_newspack_checkout       = filter_input( INPUT_GET, 'newspack_checkout', FILTER_SANITIZE_NUMBER_INT );
-		$is_newspack_donate         = filter_input( INPUT_GET, 'newspack_donate', FILTER_SANITIZE_NUMBER_INT );
```
miguelpeixe, line 159: I removed this because donations are not handled here. Please let me know if I'm missing something.

chickenn00dle: IIRC, this method will set the registration checkout flag for donations, which is why this was added. This was to set some order meta so we know whether to send a regular thank you email, or a welcome + thank you email. Maybe there is a better way to handle this now with the async cart.

includes/class-modal-checkout.php
chickenn00dle: I think we talked about making this filterable so that publishers can have other payment gateways if they'd like.

includes/class-modal-checkout.php
miguelpeixe: Is it working? Because donation checkout requests are processed in a [different method](https://github.com/Automattic/newspack-plugin/blob/bfab5b77db7ce142c8649508cadf9c975c3c3442/includes/class-donations.php#L652). We can replicate the flag logic there.

includes/class-modal-checkout.php
miguelpeixe: Right! Thank you for reminding me, updated in 893a9c41576a432465ded2fc9ac84186690b0ee0

includes/class-modal-checkout.php
miguelpeixe: Ah I see it works because it's before the `is_newspack_product` check. WDYT we replicate it to the donate request method?

miguelpeixe: Hmm... Given the entire logic for the checkout registration flag lives in the blocks plugin, I see why it's better to have it there. I'll restore it.

miguelpeixe: Updated in 35b6d59c5146f3588fe0b09316582274715809b6. I've also changed the `is_newspack_checkout` to be a guard clause for consistency.

miguelpeixe: I just now realized that because of the new strategy, the registration flag has to be handled differently. The anonymous cart generation doesn't know if it's a registration or not. Moved in 50f0e686f1719411ec5d82946fe21077010c6e29 so [this](https://github.com/Automattic/newspack-blocks/blob/af02c11a8930c67821253d47afabadebd64d1910/src/modal-checkout/modal.js#L376-L383) can set the WC session var.

src/modal-checkout/modal.js
```diff
@@ -39,6 +39,402 @@ domReady( () => {
...
+		// This check prevents an issue in Chrome where the 'load' event fired twice and the spinner was hidden too soon.
+		} else if ( 'about:blank' !== location.href ) {
+			setModalReady();
+		}
+	}
+
+	iframe.addEventListener( 'load', handleIframeReady );
+
+	/**
+	 * Generate cart.
```
chickenn00dle: Non-blocking suggestion: Just so it's documented somewhere in the code, maybe its worth adding some further details about the need to trigger this asynchronously to avoid potential session buginess?

chickenn00dle (changes requested):
Nice work @miguelpeixe!

This is working very nicely for the most part. It looks like I am no longer able to receive the welcome + thankyou emails for new registration checkout though. These should look like this:

Screenshot 2024-10-11 at 15 51 49

We use that checkout registration flag to set session data, and when the order is created set some order meta that our email logic can use to determine which thank you email to send: https://github.com/Automattic/newspack-blocks/blob/6a68ca67f9f85edb2d4fab374fa75673a353b698/includes/class-modal-checkout.php#L1703

src/modal-checkout/modal.js
miguelpeixe: Good call!

miguelpeixe:
Thank you for reviewing, @chickenn00dle!

>  I am no longer able to receive the welcome + thankyou emails for new registration checkout though. These should look like this

Yes! I noticed that too, I've recently pushed some changes that should address that. I commented further [here](https://github.com/Automattic/newspack-blocks/pull/1896#discussion_r1797377637)

src/modal-checkout/modal.js
miguelpeixe: Updated in bffe9297a6a96ead84dbffb094cde3a73e0f12ff, let me know if it's clear enough

src/modal-checkout/modal.js
```diff
@@ -565,124 +653,18 @@ domReady( () => {
...
-		}
-		if ( form ) {
-			triggerCheckout( form );
-		}
-	}
-
 	/**
 	 * Handle modal checkout url param triggers.
 	 */
 	const handleModalCheckoutUrlParams = () => {
```
chickenn00dle, line 752: Non-blocking nit: Even though we are no longer using url params for the magic link, I think it might be worth keeping this checkoutURLParam logic so publishers can do things like craft urls in newsletters so readers can be taken directly to checkout for xyz product.

chickenn00dle (approved):
Thanks for all the changes @miguelpeixe! This is looking good now! I left one non-blocking comment but feel free to ignore that.

src/modal-checkout/modal.js
miguelpeixe: Good call! Restored in 21f542cc6cea76f7bdee1f881100bd7f23424564

chickenn00dle (approved):
Great. Thanks for the quick turnaround @miguelpeixe! 🙌

miguelpeixe:
Thank you for the reviews!

PR #1899, opened 2024-10-09
fix(block-caching): only serve from cache if all blocks are cached

adekbadek:
### All Submissions:

If only some blocks are cached, rendered blocks may be duplicated. This change ensures that the cached blocks are served only if all applicable blocks on a page can be cached. 

1. On `release`,
2. Throw the wrench in the works by returning `rand( 0, 1 )` from `should_cache_block` method
3. Load a page with Homepage Posts blocks, observe that some are duplicated (might have to reload, depending on your luck)
4. Switch to this branch, reload, observe the blocks are never duplicated

### Other information:

---
  - https://app.asana.com/0/0/1207800459952421

### DIFF
--- a/includes/class-newspack-blocks-caching.php
+++ b/includes/class-newspack-blocks-caching.php
@@ -20,6 +20,18 @@ class Newspack_Blocks_Caching {
 	private static $can_serve_all_blocks_from_cache = true;
 
 	/**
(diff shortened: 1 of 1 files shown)


claudiulodro (approved):
Latest revision works great in my testing and code looks solid. Thanks for digging into this!

PR #1905, opened 2024-10-14
fix(modal-checkout): move checkout_error trigger

miguelpeixe:
### All Submissions:

The `checkout_error` may not always run because the script might end up running after the `DOMContentLoaded` event dispatch. This PR tweaks the approach so the trigger runs in the modal script, which runs inside `domReady()` and is ensured to always run.

1. While on the `epic/ras-acc` branch, use the checkout button to purchase a subscription you already own (make sure the subscription product is configured to restrict such purchase)
2. Click the checkout button a few times to confirm it sometimes hangs in the loading state and sometimes goes through with the correct error message (try also throttling the internet connection in a few attempts)
3. Checkout this branch, click the checkout button several times and confirm the error message renders consistently

### Other information:


chickenn00dle (approved):
Great catch!

PR #1907, opened 2024-10-17
fix(ras-acc): prevent iframeobserving from observing empty container

chickenn00dle:
### All Submissions:

While working on a bug on TheFifthEstate, I ran into the following error:

Screenshot 2024-10-17 at 10 47 35

Seems for tier based donation forms, the `process_donation_request` method returns early since there is a missing `donation_value_FREQUENCY` input: https://github.com/Automattic/newspack-plugin/blob/c9e91743c9ac7bf0053b506ed343da2a2e77609f/includes/class-donations.php#L679

In this case, the checkout url returned during the checkout process ends up being something like: `.../undefined&_newspack_checkout_registration=1`

There is also a case where `setModalReady` can be called before container is available: https://github.com/Automattic/newspack-blocks/blob/7ed3c4ee0b84a273fe1bda013fef54ef3e2db264/src/modal-checkout/modal.js#L126. Which will trigger in this case and lead to the console error above.

This PR fixes these issues by adding the input to the tiered donation form on button click, and removing the additional `setModalReady` case since this is no longer solely tied to the iframe `load` event.

1. Set up a tier based donation form on any page or post
2. As a new reader, go through the auth + checkout flow using the donation form
3. On `epic/ras-acc` the form will hang with an error. On this branch it should render checkout as expected.
4. Smoke test the checkout flow as a new reader, an existing reader, etc.

### Other information:

### DIFF
--- a/src/modal-checkout/modal.js
+++ b/src/modal-checkout/modal.js
@@ -384,6 +384,7 @@ domReady( () => {
 					} )
 					.catch( error => {
 						console.warn( 'Unable to generate cart:', error ); // eslint-disable-line no-console
+						closeCheckout();
 					} );
 				},
 				skipSuccess: true,
--- a/src/modal-checkout/modal.js
+++ b/src/modal-checkout/modal.js
@@ -56,7 +56,8 @@ domReady( () => {
 			if ( this.readyState === "complete" ) {
 				ready.call( this );
 			}
-		}		function checkLoaded() {
+		}
+		function checkLoaded() {
 			if ( iframe._ready ) {
 				clearTimeout( iframe._readyTimer );
 				return;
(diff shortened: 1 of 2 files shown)


src/modal-checkout/modal.js
```diff
@@ -387,6 +384,7 @@ domReady( () => {
 					} )
 					.catch( error => {
 						console.warn( 'Unable to generate cart:', error ); // eslint-disable-line no-console
+						closeCheckout();
```
chickenn00dle, line 387: For cases where this errors, let's ensure the checkout is closed. I think this is a better experience than waiting for the modal to load when it might never render.

miguelpeixe (approved):
Good catch! It fixes the issue and from my testing didn't introduce any regression to other checkout flows.

Also, thank you for the lint in `iframeReady` 😄

PR #1908, opened 2024-10-17
fix(modal-checkout): handle auth modal analytics

miguelpeixe:
### All Submissions:

1200530782742699-as-1208558693529961/f

When a reader starts the modal checkout they may get prompted to authenticate or register, which will not trigger the GA4 event to register the intent.

This PR, along with https://github.com/Automattic/newspack-plugin/pull/3481, implements that support.

Along with that:
 - Some variable names have also been updated for better readability.
 - Fixed the gtag integration within the checkout, which was being dequeued and unable to trigger the `form_submission` event

1. Checkout this branch and add the following to line 24 of `src/modal-checkout/analytics/ga4/utils/index.js` for easier testing:

```js
console.log({ eventName, payload });
```

2. In a fresh session, start a checkout button block flow
3. Confirm the `opened` event is logged in the console
4. Close the auth modal and confirm the `dismissed` event is logged in the console with the correct payload
5. Start the flow again and confirm it also triggers a new `opened` event (`checkoutOpen` changed to `false`)
6. Go through registration and confirm that opening the checkout does not trigger a new `opened` event
7. Purchase and confirm the `form_submission` event is triggered with the correct payload
8. Confirm that finishing the purchase flow will not trigger the `dismissed` event
9. Start the checkout flow for a variable subscription
10. Confirm that opening the variation modal triggers the `opened_variations` event
11. Select a variation and confirm the `opened` event is triggered
12. Close the modal and confirm `dismissed` is triggered
13. Click again and confirm `opened_variations` triggers again

### Other information:

### DIFF
--- a/src/modal-checkout/modal.js
+++ b/src/modal-checkout/modal.js
@@ -388,7 +388,7 @@ domReady( () => {
 			// Initialize auth flow if reader is not authenticated.
 			window.newspackReaderActivation.openAuthModal( {
 				title: newspackBlocksModal.labels.auth_modal_title,
-				callback: ( message, authData ) => {
+				onSuccess: ( message, authData ) => {
 					cartReq.then( url => {
 						// If registered, append the registration flag query param to the url.
 						if ( authData?.registered ) {
@@ -402,7 +402,10 @@ domReady( () => {
 						closeCheckout();
 					} );
 				},
-				onClose: () => {
+				onError: () => {
+					closeCheckout();
+				},
+				onDismiss: () => {
 					// Analytics: Track a dismissal event (modal has been manually closed without completing the checkout).
 					manageDismissed( analyticsData );
 					inCheckoutIntent = false;
@@ -524,7 +527,8 @@ domReady( () => {
 
 			if ( window?.newspackReaderActivation?.openNewslettersSignupModal ) {
 				window.newspackReaderActivation.openNewslettersSignupModal( {
-					callback: handleCheckoutComplete,
+					onSuccess: handleCheckoutComplete,
+					onError: handleCheckoutComplete,
 					closeOnSuccess: shouldCloseModal,
 				} );
 			} else {
--- a/src/modal-checkout/modal.js
+++ b/src/modal-checkout/modal.js
@@ -21,7 +21,8 @@ const VARIATON_MODAL_CLASS_PREFIX = 'newspack-blocks__modal-variation';
 
 // Track the checkout state for analytics.
 let analyticsData = {};
-let checkoutOpen = false;
+// Track the checkout intent to avoid multiple analytics events.
+let inCheckoutIntent = false;
 
 domReady( () => {
 	const modalCheckout = document.querySelector( `#${ MODAL_CHECKOUT_ID }` );
@@ -233,10 +234,10 @@ domReady( () => {
 				const formAnalyticsData = form.getAttribute( 'data-product' );
 				analyticsData = formAnalyticsData ? JSON.parse( formAnalyticsData ) : {};
 
-				// For the variation modal we will not set `checkoutOpen = true` and
+				// For the variation modal we will not set `inCheckoutIntent = true` and
 				// let the `opened` event get triggered once the user selects a
 				// variation so we track the selection.
-				if ( ! checkoutOpen ) {
+				if ( ! inCheckoutIntent ) {
 					manageOpened( analyticsData );
 				}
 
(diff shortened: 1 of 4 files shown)


src/modal-checkout/modal.js
```diff
@@ -19,7 +19,9 @@ const MODAL_CHECKOUT_ID = 'newspack_modal_checkout';
 const MODAL_CLASS_PREFIX = `${ CLASS_PREFIX }__modal`;
 const VARIATON_MODAL_CLASS_PREFIX = 'newspack-blocks__modal-variation';
 
-let getProductDataModal = {};
+// Track the checkout state for analytics.
+let analyticsData = {};
+let checkoutOpen = false;
```
chickenn00dle: Is this flag necessary? Can we just use the `data-state` of the modal checkout container which will be `open` when the modal is open and `closed` when not?

src/modal-checkout/modal.js
```diff
@@ -389,6 +403,12 @@ domReady( () => {
 						console.warn( 'Unable to generate cart:', error ); // eslint-disable-line no-console
 					} );
 				},
+				onClose: () => {
```
chickenn00dle: Non-blocking nit/question. Feels weird to provide an `onClose` callback option that won't be called everytime. Especially considering we already provide a generic `callback` option. Is it possible to simply use the `callback` for analytics purposes here, using the `authData` and open state as a flag for whether this was dismissed?

chickenn00dle (commented):
Nice work @miguelpeixe!

This works as expected. I just had one question about using a flag for open state before giving a thumbs up.

Also had a second non-blocking question/nit but feel free to ignore that one!

chickenn00dle (commented):
> add the following to line 24 of src/modal-checkout/analytics/ga4/utils/index.js for easier testing

Would be neat if we could make this logging permanent and only have it enabled when some development flag is present or something like that. Definitely not something needed in this PR, but thought I'd bring it up.

src/modal-checkout/modal.js
miguelpeixe: The `checkoutOpen` is also set to `true` when an anonymous reader intent to purchase triggers the auth modal. In this case, we can't leverage the checkout modal state because it's not yet open at this point.

We can rename the variable or have a different strategy, but I don't think going through each modal state would be practical for this.

src/modal-checkout/modal.js
miguelpeixe: My concern is that the `callback` is only called on success, so, as is, we don't have a way to hook into dismissals.

I think we can go 2 ways:

1. Make the `callback` also trigger on dismissals, including the success/error/dismiss as a callback argument
2. Change `callback` to `onSuccess` and preserve `onClose` (or rename to `onDismiss`)

WDYT?

miguelpeixe:
> Would be neat if we could make this logging permanent and only have it enabled when some development flag is present or something like that.

That can be handy. I usually use browser extensions to track GA behavior on a page. I just figured the console log would be more straightforward for this PR than suggesting an extension.

src/modal-checkout/modal.js
miguelpeixe: I updated the variable name and added a comment in 94fea3f2cd765f7707f5b9abd14b403d528bde47. I think this change makes it more clear that it's not about the checkout modal being open. Let me know if you think this works

src/modal-checkout/modal.js
chickenn00dle: > Change callback to onSuccess and preserve onClose (or rename to onDismiss)

I like the idea of onSuccess/onDismiss. I feel this makes it super clear where functionality for each should go. But truthfully I'm fine with either approach if you have a preference for `1`.

src/modal-checkout/modal.js
chickenn00dle: This works for me! Thanks for clarifying @miguelpeixe 👍

src/modal-checkout/modal.js
miguelpeixe: I also like 2 better 😄

miguelpeixe:
~f958d868c4f0912cc40f9415f1b943198eccb5cf fixes a missing dependency for the checkout form~. Reversed in c5bb48d99a4e7ab7ce66ca77ea51b51966bb6689 because that's already being taken care of in https://github.com/Automattic/newspack-blocks/pull/1909

src/modal-checkout/modal.js
miguelpeixe: Updated in 89aacb6ab1ac71b5c1f75d1e536507d2e03762ab and https://github.com/Automattic/newspack-plugin/pull/3481/commits/cda1b5c52c69700d02b48caf83dfd10ddb6e9ece. I've also introduced the `onError` callback and updated the newsletter signup modal to follow the same standard.

chickenn00dle (approved):
Looks good! Thanks for the changes @miguelpeixe!

PR #1880, opened 2024-09-12
fix (ras-acc): add styles for terms and conditions

laurelfulford:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-theme/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

This PR adds some styles for WooCommerce's built in terms and conditions checkbox, which appears on the second screen of the modal checkout.

See: 1207817176293825-as-1208279455320432

### How to test the changes in this Pull Request:

1. Create a page and title it "Terms and Conditions" or something similar. Give it some text content.
2. Under WP Admin > WooCommerce > Advanced, under the Page Settings header, use the Terms and conditions dropdown to find your page:

CleanShot 2024-09-11 at 18 42 26

3. Run through a checkout; note the appearance of the Terms and Conditions checkbox at the bottom of the second screen.

CleanShot 2024-09-11 at 18 45 34

4. Click the Terms and Conditions text to open the terms details and note the appearance.

CleanShot 2024-09-11 at 18 46 05

5. Apply this PR and run `npm run build`.
6. Repeat steps 3 and 4 and confirm that the Terms checkbox and details look better:

CleanShot 2024-09-11 at 18 44 18

CleanShot 2024-09-11 at 18 44 34

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #1910, opened 2024-10-18
fix(ras-acc): apply terms and conditions styles

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1207817176293825/1208579907739880

Looks like our terms and conditions styles weren't being applied correctly because we were targeting invalid classnames and nested elements. This PR adjusts our checkout styles to correctly target this section.

**NOTE:** I'm making some assumptions about what this should look like based on the existing styles since I didn't see this checkbox in the Figma. Let me know if this should look differently!

Without T&C set:
Screenshot 2024-10-18 at 13 49 25

With T&C set:
Screenshot 2024-10-18 at 13 52 04

1. Set up a terms and conditions page, then apply it in the customizer (WooCommerce > Checkout > Terms and conditions page)
2. As a reader, go through checkout and confirm the Terms and Conditions checkbox appears as shown in the 'With T&C set' screenshot above.
3. Remove the terms and conditions page via the customizer again
4. Repeat step 2, but this time confirm the Terms and Conditions checkbox appears as shown in the 'Without T&C set' screenshot above.

### Other information:

### DIFF
--- a/src/modal-checkout/checkout.scss
+++ b/src/modal-checkout/checkout.scss
@@ -556,7 +556,7 @@
 				gap: 0;
 
 				input {
-					margin: 0 var(--newspack-ui-spacer-base, 8px) 0 0;
+					margin: 0 var(--newspack-ui-spacer-2, 12px) 0 0;
 				}
 				// Override checkout form terms and condition text.
 				.woocommerce-terms-and-conditions-checkbox-text {
--- a/src/modal-checkout/checkout.scss
+++ b/src/modal-checkout/checkout.scss
@@ -531,6 +531,14 @@
 			color: var(--newspack-ui-color-neutral-60, colors.$newspack-ui-color-neutral-60);
 		}
 
+		.woocommerce-terms-and-conditions {
+			margin-top: var(--newspack-ui-spacer-5, 24px);
+
+			*:first-child {
+				margin-top: 0;
+			}
+		}
+
 		p.form-row {
 			background: var(--newspack-ui-color-neutral-5, #f7f7f7);
 			border-color: var(--newspack-ui-color-border, #ddd);
--- a/src/modal-checkout/checkout.scss
+++ b/src/modal-checkout/checkout.scss
@@ -553,6 +553,7 @@
 				height: 20px;
 				margin: 0;
 				padding: 0;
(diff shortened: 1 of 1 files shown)


laurelfulford (changes requested):
If it helps, I did get a thumbs up from Kevin for the styles used [here](https://github.com/Automattic/newspack-blocks/pull/1880) -- the main difference is the font size and lack of background.

Personally I like how it looks now! The only thing that stands out to me with the label part is the distance between the text and the asterisks. 

When the terms are open, the spacing is also a little funky -- the terms box is closer to the privacy policy text than the toggle: 

CleanShot 2024-10-18 at 11 40 37@2x

I missed this myself, but when you add a heading the spacing at the top of the terms also seems like a lot! I think removing the top margin from the first element should take care of most cases of that.

chickenn00dle:
Thanks for the review @laurelfulford! 

I've got a fix up for the terms box and abbreviation spacing in https://github.com/Automattic/newspack-blocks/pull/1910/commits/850d5aa4395d30430980ca5e384f56513aa00f32 and https://github.com/Automattic/newspack-blocks/pull/1910/commits/ac5e438bb053c52646aefd67b2553c6fc0d5df85

Let me know if this works!

laurelfulford (approved):
Thanks @chickenn00dle! This looks awesome overall! The only other thing I see is that the spacing next to the checkbox seems to be a bit smaller usual; it's only really noticeable when you have the checkboxes back to back like this:

CleanShot 2024-10-18 at 14 07 29

I think it's just a matter of changing `var(--newspack-ui-spacer-base,8px)` to `var(--newspack-ui-spacer-2,12px)` to match the checkbox above; marking as approved since it's a small tweak and things otherwise are solid!

chickenn00dle:
Thanks @laurelfulford!

> I think it's just a matter of changing var(--newspack-ui-spacer-base,8px) to var(--newspack-ui-spacer-2,12px) to match the checkbox above; marking as approved since it's a small tweak and things otherwise are solid!

Good catch! Fixed in https://github.com/Automattic/newspack-blocks/pull/1910/commits/6516aae412801ae2d1a4a1de9365fc095e95f9d0

PR #1912, opened 2024-10-21
fix(ras-acc): handle recaptcha v2 errors

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1207817176293825/1208344705374759/f

Part of https://github.com/Automattic/newspack-plugin/pull/3489.

This PR fixes how we handle recaptcha v2 errors. The recaptcha script currently returns a string error which our blocks script seems to not handle correctly. To fix this, we add a try/catch when handling the error message and if jquery fails to convert the string to a jquery object, we append it to a paragraph instead. We also add new onSuccess and onError callbacks to the recaptcha render call to handle rendering the recaptcha error and scrolling on success.

Screenshot 2024-10-21 at 17 06 55

BEFORE TESTING BE SURE TO CHECKOUT https://github.com/Automattic/newspack-plugin/pull/3489 AS WELL

**Recaptcha V2**
1. As admin enable recaptcha v2 via Newspack > Connections > reCAPTCHA
2. Temporarily comment out [this line](https://github.com/Automattic/newspack-plugin/blob/f9d4c5389ff6d1824844e64f91a70932efe0ecca/src/other-scripts/recaptcha/index.js#L171) and replace it with `onError?.( 'There was an error with reCAPTCHA. Please try again.' )` (and build) to force a v2 error.
3. As a reader, go through the checkout flow via any checkout or donation button on the site
4. Confirm you see a checkout form error appear with the message hardcoded from step 2 above.
5. Revert the change you made in step 2
6. Smoke test the following flows:
  - Registration flow via signin button in site header
  - Checkout flow as logged in reader
  - Registration + Checkout flow as new reader via any checkout or donation button

**Recaptcha V3**
1. As admin enable recaptcha v3 via Newspack > Connections > reCAPTCHA
2. Smoke test the following flows:
  - Registration flow via signin button in site header
  - Checkout flow as logged in reader
  - Registration + Checkout flow as new reader via any checkout or donation button

**No Recaptcha**
1. As admin disable recaptcha altogether via Newspack > Connections > reCAPTCHA
2. Smoke test the following flows:
  - Registration flow via signin button in site header
  - Checkout flow as logged in reader
  - Registration + Checkout flow as new reader via any checkout or donation button

### Other information:

### DIFF
--- a/src/modal-checkout/index.js
+++ b/src/modal-checkout/index.js
@@ -207,7 +207,6 @@ import { domReady } from './utils';
 				 * @param {string} error_message
 				 */
 				function handleFormError( error_message ) {
-					unblockForm( $form );
 					$form
 						.find( '.input-text, select, input:checkbox' )
 						.trigger( 'validate' )
@@ -247,6 +246,8 @@ import { domReady } from './utils';
 						}
 					};
 
+					clearNotices();
+
 					/**
 					 * The new "wc-block-components-notice-banner" does not provide a <li />
 					 * of errors when only one field failed validation.
@@ -264,8 +265,6 @@ import { domReady } from './utils';
 						} );
 					}
 
-					clearNotices();
-
 					// Handle generic errors.
 					if ( genericErrors.length ) {
 						$fieldToFocus = false; // Don't focus a field if validation returned generic errors.
@@ -280,6 +279,8 @@ import { domReady } from './utils';
 						window.scroll( { top: $fieldToFocus.offset().top - 100, left: 0, behavior: 'smooth' } );
 						$fieldToFocus.find( 'input.input-text, select, input:checkbox' ).trigger( 'focus' );
 					}
+
+					unblockForm( $form );
 
 					$( document.body ).trigger( 'update_checkout' );
 					$( document.body ).trigger( 'checkout_error', [ error_message ] );
@@ -455,11 +456,12 @@ import { domReady } from './utils';
 						// Initiate reCAPTCHA, if available.
 						if ( newspack_grecaptcha?.render ) {
 							$form.data( 'newspack-recaptcha', 'newspack_modal_checkout' );
-							newspack_grecaptcha.render(
-								$form.get(),
-								() => window.scroll( { top: 0, left: 0, behavior: 'smooth' } ),
-								( error ) => handleFormError( error )
-							);
+							const onSuccess = () => $form.get( 0 ).scrollIntoView( true, { behavior: 'smooth' } );
+							const onError = ( error ) => handleFormError( error );
+							newspack_grecaptcha.render( $form.get(), onSuccess, onError );
+							// Refresh reCAPTCHAs on Woo checkout update and error.
+							$( document ).on( 'updated_checkout', () => newspack_grecaptcha.render( $form.get(), onSuccess, onError ) );
+							$( document.body ).on( 'checkout_error', () => newspack_grecaptcha.render( $form.get(), onSuccess, onError ) );
 						}
 						if ( $coupon.length ) {
 							$coupon.show();
--- a/src/modal-checkout/index.js
+++ b/src/modal-checkout/index.js
@@ -456,7 +456,10 @@ import { domReady } from './utils';
 						// Initiate reCAPTCHA, if available.
 						if ( newspack_grecaptcha?.render ) {
 							$form.data( 'newspack-recaptcha', 'newspack_modal_checkout' );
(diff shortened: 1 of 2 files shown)


src/modal-checkout/index.js
```diff
@@ -450,7 +456,15 @@ import { domReady } from './utils';
...
-							newspack_grecaptcha.render( $form.get(), ( error ) => handleFormError( error ) );
+							const onSuccess = () => {
+								clearNotices();
+								$form.get( 0 ).scrollIntoView( { behavior: 'smooth' } );
+							}
+							const onError = ( error ) => handleFormError( error );
+							newspack_grecaptcha.render( $form.get(), onSuccess, onError );
+							// Refresh reCAPTCHAs on Woo checkout update and error.
+							$( document ).on( 'updated_checkout', () => newspack_grecaptcha.render( $form.get(), onSuccess, onError ) );
+							$( document.body ).on( 'checkout_error', () => newspack_grecaptcha.render( $form.get(), onSuccess, onError ) );
```
chickenn00dle, line 464: Moved these over from the plugin script since these bits are specific to modal checkout

src/modal-checkout/index.js
```diff
@@ -450,7 +456,15 @@ import { domReady } from './utils';
 						// Initiate reCAPTCHA, if available.
 						if ( newspack_grecaptcha?.render ) {
 							$form.data( 'newspack-recaptcha', 'newspack_modal_checkout' );
-							newspack_grecaptcha.render( $form.get(), ( error ) => handleFormError( error ) );
+							const onSuccess = () => {
+								clearNotices();
+								$form.get( 0 ).scrollIntoView( { behavior: 'smooth' } );
```
chickenn00dle, line 458: Moved scrolling the form over from the plugin script since this was specific to modal checkout

src/modal-checkout/index.js
```diff
@@ -247,12 +246,18 @@ import { domReady } from './utils';
...
 					 * The new "wc-block-components-notice-banner" does not provide a <li />
 					 * of errors when only one field failed validation.
 					 */
 					if ( ! error_message.includes( '<li' ) ) {
-						handleErrorItem( $( error_message ) );
+						try {
+							handleErrorItem( $( error_message ) );
+						} catch {
+							handleErrorItem( $( '<p />' ).append( error_message ) );
+						}
```
miguelpeixe: I suggest a small refactor here to avoid the use of try/catch:

```js
					if ( error_message.indexOf( '<' ) !== 0 ) {
						// If error_message is not an HTML string, wrap it in a <li />.
						handleErrorItem( $( '<li />' ).append( error_message ) );
					} else if ( ! error_message.includes( '<li' ) ) {
						// If not a list, handle as a single error.
						handleErrorItem( $( error_message ) );
					} else {
						// Handle multiple errors.
						const $errors = $( error_message );
						$errors.find( 'li' ).each( function () {
							handleErrorItem( $( this ) );
						} );
					}
```

I also changed the wrapper to `<li />` so the handler doesn't have to [generate a new node](https://github.com/Automattic/newspack-blocks/blob/e7457201337fe0f7a4c80ac0a241308b57d4e502/src/modal-checkout/index.js#L242-L244).

miguelpeixe (changes requested):
It's working great for the v2 invisible badge! I also tested what happens with the v2 checkbox version and I get an infinite loop of `There was an error with reCAPTCHA. Please try again.` alerts in the frontend.

Can we make this unsupported strategy fail better? It would be nice to detect that while saving the keys, but I'm not sure Google has an API for that.

Besides that, just another comment inline.

src/modal-checkout/index.js
chickenn00dle: Good idea! Updated in https://github.com/Automattic/newspack-blocks/pull/1912/commits/cf06a1d509e9a7e286c11832817ec7f00840a702

chickenn00dle:
Thanks for the review @miguelpeixe!

> Can we make this unsupported strategy fail better? It would be nice to detect that while saving the keys, but I'm not sure Google has an API for that.

I also didn't see a way to detect this when saving keys, but added a retry count to the recaptcha render to prevent this infinite failure when the API keys are incorrect: https://github.com/Automattic/newspack-plugin/pull/3489/commits/18d16b460a8448417fdeab818c613d8fb8af45f4

There is still some weirdness with the modal checkout, but I think this is fine given the issue is invalid API keys and checkout will never succeed in these cases.

miguelpeixe (approved):
Thank you for revising! The retry count is a good workaround.

PR #1916, opened 2024-10-21
fix: add check if product can be purchased

laurelfulford:
### All Submissions:

This PR uses WooCommerce's `purchaseable` check to see if a product can be purchased before it can be assigned to a Checkout Button block, or before the Checkout Button block is rendered on the front-end.

This change is to fix an edge case where you can assign a price-less product to the Checkout Button block, which causes the modal checkout to throw an error. Some products -- like NYP products -- don't seem to throw the same error even if they don't have a suggested or minimum price set; using the `purchaseable` check seems to account for this but I'd definitely appreciate thoughts about this approach!

See 1207817176293825-as-1208327956169322

1. Set up a few different products, including:
   * A regular simple product with a price.
   * A product with variations with prices.
   * A simple product without a price.
   * A product with variations, at least one without a price.
   * A regular NYP product, with a minimum or suggested price.
   * A NYP product without a minimum and suggested price.
2. View the /shop page, and confirm that some of your products can't be purchased (they should include a link to the individual product screen, but no purchase button). This should just be your product with no price, unless I've messed up a test case. 
3. Set up a checkout button block for each. Note that you can do this with the product without a price, but it will throw an error on the front-end about subscriptions, or just return a button:

CleanShot 2024-10-21 at 15 17 43

5. On a new page, try to add a checkout button block for each product. Note your price-less product won't be returned when you search for products in the sidebar. 
6. On the old page, confirm that the price-less product is still in the editor, but no longer renders on the front-end. 

### Other information:

### DIFF
--- a/src/blocks/checkout-button/edit.js
+++ b/src/blocks/checkout-button/edit.js
@@ -100,7 +100,9 @@ function ProductControl( props ) {
 			.then( products => {
 				const _suggestions = {};
 				products.forEach( product => {
-					_suggestions[ product.id ] = `${ product.id }: ${ product.name }`;
+					if ( product.purchasable ) {
+						_suggestions[ product.id ] = `${ product.id }: ${ product.name }`;
+					}
 				} );
 				setSuggestions( _suggestions );
 			} )
--- a/src/blocks/checkout-button/view.php
+++ b/src/blocks/checkout-button/view.php
@@ -34,11 +34,27 @@ function render_callback( $attributes, $content ) {
 	}
 	$product_id = $attributes['product'];
 
-	// Check if product can actually be purchased before rendering.
 	if ( function_exists( 'wc_get_product' ) ) {
 		$product = wc_get_product( $product_id );
+
+		// Check if product can actually be purchased before rendering.
 		if ( ! $product || ! $product->is_purchasable() ) {
 			return '';
+		}
+
+		// Check if product is NYP with no prices set at all.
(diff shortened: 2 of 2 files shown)


chickenn00dle (commented):
Looks good @laurelfulford!

My only concern is the NYP w/o a suggested or minimum price case. In RAS-ACC that will be handled by a new input to update the price within modal checkout, but on `trunk` there is no way to adjust this price. Should we also be hiding these in `trunk` then getting a PR up in RAS-ACC to unhide them?

I'm fine if the answer is no since this is such an edge case, but wanted to at least discuss it.

laurelfulford:
> My only concern is the NYP w/o a suggested or minimum price case.

Ahhh, that's a good point!

I've added a check for NYP on the front-end (don't show buttons for products without prices), but in the editor they're still available since you can set a price for the block once you pick a product. Hopefully this isn't too confusing!

This should probably need some more testing steps, like:

1. Checking if a button shows when the product has each a suggested or minimum price, or both.
2. Checking if a button shows when the product has a price set on the button itself (in the sidebar). 

Let me know if that makes sense, or if you have any questions -- thanks @chickenn00dle!!

chickenn00dle (approved):
Nice! Thanks for the changes @laurelfulford 🙌. This looks good now 🚢

laurelfulford:
Thanks @chickenn00dle! 🙌

laurelfulford:
Probably a bit of overkill, but I created a PR to merge trunk into epic/ras-acc [here](https://github.com/Automattic/newspack-blocks/pull/1918) just to make sure the NYP stuff got removed, and nothing else weird happened (the view.php files are VERY different!).

PR #1923, opened 2024-10-24
fix(modal-checkout): empty cart when modal checkout is abandoned

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1207817176293825/1208620127411764/f

This PR adds an ajax request to empty cart when modal checkout is abandoned.

Screenshot 2024-10-24 at 12 32 08

1. Via any checkout or donate button open modal checkout as a reader
2. In another tab (in the same session) open the cart page `test.site/cart`
3. Verify the item is present in the cart
4. Back in the modal checkout tab, close the checkout modal without completing checkout
5. Go back to the cart tab. On `epic/ras-acc` the item will still be present. On this branch, the cart should be empty.

### Other information:

### DIFF
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -316,7 +316,7 @@ final class Modal_Checkout {
 	 * Process abandon checkout for modal.
 	 */
 	public static function process_abandon_checkout() {
-		if ( is_admin() || ! defined( 'DOING_AJAX' ) ) {
+		if ( ! defined( 'DOING_AJAX' ) ) {
 			return;
 		}
 
@@ -324,7 +324,7 @@ final class Modal_Checkout {
 			return;
 		}
 
-		if ( ! check_ajax_referer( 'newspack_modal_checkout_nonce', 'security', false ) ) {
+		if ( ! check_ajax_referer( 'newspack_modal_checkout_nonce' ) ) {
 			wp_send_json_error( [ 'message' => __( 'Invalid nonce.', 'newspack-blocks' ) ] );
 			wp_die();
 		}
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -56,6 +56,8 @@ final class Modal_Checkout {
 		add_action( 'wp_ajax_modal_checkout_request', [ __CLASS__, 'process_checkout_request' ] );
 		add_action( 'wp_ajax_nopriv_modal_checkout_request', [ __CLASS__, 'process_checkout_request' ] );
 		add_action( 'wp', [ __CLASS__, 'process_checkout_request' ] );
+		add_action( 'wp_ajax_abandon_modal_checkout', [ __CLASS__, 'process_abandon_checkout' ] );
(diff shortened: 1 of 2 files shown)


includes/class-modal-checkout.php
```diff
@@ -310,6 +312,32 @@ function ( $item ) {
...
+	public static function process_abandon_checkout() {
+		if ( is_admin() || ! defined( 'DOING_AJAX' ) ) {
+			return;
+		}
+
+		if ( ! self::is_modal_checkout() ) {
+			return;
+		}
+
+		if ( ! check_ajax_referer( 'newspack_modal_checkout_nonce', 'security', false ) ) {
```
miguelpeixe: Non-blocking suggestion: use `nonce` or just the default `_wpnonce`/`ajax_nonce` as the request argument because it's standard practice in WP.

includes/class-modal-checkout.php
```diff
@@ -310,6 +312,32 @@ function ( $item ) {
 		}
 	}
 
+	/**
+	 * Process abandon checkout for modal.
+	 */
+	public static function process_abandon_checkout() {
+		if ( is_admin() || ! defined( 'DOING_AJAX' ) ) {
```
miguelpeixe: When doing ajax, `is_admin()` is always true. This check is preventing all `abandon_modal_checkout` requests to get through.

```suggestion
		if ( ! defined( 'DOING_AJAX' ) ) {
```

miguelpeixe (changes requested):
It works well when I'm already authenticated on the page load, but it doesn't work if I have to authenticate during the checkout process.

Besides potential issues with cached values on anonymous visits, nonces use authentication as part of validation. Once I authenticate the nonce provided on page load will no longer be valid.

To get around that, we could provide the nonce localized to the checkout script, then pull it in `modal.js`. WDYT?

includes/class-modal-checkout.php
chickenn00dle: Good call. Done in https://github.com/Automattic/newspack-blocks/pull/1923/commits/8b9930e666cfe35f003bf8ef1c90fb66cd8f01ec

includes/class-modal-checkout.php
chickenn00dle: Good catch! Fixed in https://github.com/Automattic/newspack-blocks/pull/1923/commits/8b9930e666cfe35f003bf8ef1c90fb66cd8f01ec

chickenn00dle:
Thanks for the review @miguelpeixe!

> Once I authenticate the nonce provided on page load will no longer be valid.

Ah whoops! I always forget about the transition in login state 🤦 . I moved the nonce to the checkout script and pulled in `modal.js` as you suggested: https://github.com/Automattic/newspack-blocks/pull/1923/commits/7a1c99ac92428484667f5e187da7a3f53ab58f95

miguelpeixe (approved):
Thank you for the revisions! Tested well with and without the auth flow.

PR #1918, opened 2024-10-22
fix: check for purchaseable products merged from trunk

laurelfulford:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-theme/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

[This PR](https://github.com/Automattic/newspack-blocks/pull/1916) changed code that was very different between trunk and epic/ras-acc, so I made this one to merge trunk into epic/ras-acc early, and to remove the NYP check we don't need in RAS-ACC.

See 1207817176293825-as-1208327956169322

### How to test the changes in this Pull Request:

Follow the steps to test from https://github.com/Automattic/newspack-blocks/pull/1916, minus the NYP comment -- NYP product should be allowed to be added to the Checkout Button block and the block should show on the front-end because readers can set a purchase price from the checkout.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #1925, opened 2024-10-25
fix: rework approach of removing invalid products from Checkout Block

laurelfulford:
### All Submissions:

This PR uses a different approach to add a check for whether a product should be useable with the Checkout Button block. 

On the front-end, the criteria is:
* The product need to be a variable product without a variation pre-selected, OR
* The product needs to have a price set (whether it's a regular product, or a variable product with variation selected), OR
* The product needs to be a NYP product. 

If products don't match, the button won't be visible on the front-end.

On the backend, the criteria for the Product selector is:
* The product needs to have a price set, OR
* The product needs to be a variable product, OR
* The product needs to be set to NYP.

If products don't match, they won't show up as suggestions for future buttons.

On the backend, the criteria for the variation selector is:
* The variation needs to have a price, OR
* The variation needs to be set to NYP.

If variations don't match, they won't show up as suggestions for future buttons. (Right now if a variation doesn't have a price, it won't show up in the variation picker step of the modal checkout).

See: 1207817176293825-as-1208327956169322

1. Set up a bunch of different products. These include:
* A simple product with a price
* A simple product without a price
* A NYP product with no prices set
* A NYP product with at least one price (suggested or minimum)
* A variable product with at least one NYP variation, and at least one variation without a price.
2. Prior to applying this PR, try adding once of each to a page. For the variable product, add two buttons, one that allows readers to pick the variation, and one with the variation set to the option without a price.
4. On the front-end, confirm the buttons assigned to price-less products (the simple product without a price, and the variable product set to a price-less variation) don't appear.
5. Create a new page. 
6. Repeat step 2; confirm that when you try to assign price-less product to a button, it doesn't show up in the search.
7. Confirm the NYP button without prices shows up in the search.
8. When you add your product with variations and try to assign a specific variation, confirm that the variation without a price doesn't appear, but that the NYP one does. 
9. Publish your page; confirm that all of the newly added buttons appear.
10. Smoke test other product types and settings, like subscriptions, sale prices, free trials...
11. Let me know if you think I'm missing anything!!

### Other information:

### DIFF
--- a/src/blocks/checkout-button/edit.js
+++ b/src/blocks/checkout-button/edit.js
@@ -92,6 +92,7 @@ function ProductControl( props ) {
 	const [ suggestions, setSuggestions ] = useState( {} );
 	const [ selected, setSelected ] = useState( false );
 	const [ isChanging, setIsChanging ] = useState( false );
+
 	function fetchSuggestions( search ) {
 		setInFlight( true );
 		return apiFetch( {
@@ -100,7 +101,7 @@ function ProductControl( props ) {
 			.then( products => {
 				const _suggestions = {};
 				products.forEach( product => {
-					if ( product.purchasable ) {
+					if ( '' !== product.price || getNYP( product ).isNYP ) { // Variable products will populate price with one of the variations prices.
 						_suggestions[ product.id ] = `${ product.id }: ${ product.name }`;
 					}
 				} );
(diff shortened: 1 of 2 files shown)


miguelpeixe (changes requested):
> Repeat step 2; confirm that when you try to assign price-less product to a button, it doesn't show up in the search.

I'm also unable to assign price-less products while on the `epic/ras-acc` branch 🤔 

> When you add your product with variations and try to assign a specific variation, confirm that the variation without a price doesn't appear, but that the NYP one does.

How do we handle NYP without any set price? It doesn't render great on the variation modal:

image

And in the checkout it's a zero money transaction:

image

laurelfulford:
> I'm also unable to assign price-less products while on the epic/ras-acc branch 🤔

🤦‍♀️ Gah, this is my bad! There's [a previous fix](https://github.com/Automattic/newspack-blocks/pull/1918) for this that's already committed, that's what's preventing being able to add price-less products to the editor. The only way to do it would be to remove [this `product.purchaseable` check](https://github.com/Automattic/newspack-blocks/blob/epic/ras-acc/src/blocks/checkout-button/edit.js#L103-L105).

> How do we handle NYP without any set price? It doesn't render great on the variation modal:

Ah, that's terrible 😞 I don't think this is a case we've actually tested before, that's a great catch!

The idea behind allowing NYP products with no prices is that they can still technically be purchased (since you can set the price on the final screen). But maybe we should block these as well? (That's how it would need to be handled in trunk since you can't currently set the price while checking out). 

I'm going to flag this for design in the Asana task to figure out the best UX, and then take it from there. Thanks for catching this, @miguelpeixe!

miguelpeixe:
> The idea behind allowing NYP products with no prices is that they can still technically be purchased

Ah, right!

image

We can figure out a design change to adjust the price-less NYP in the modal in a separate PR. This approach works well!

miguelpeixe (approved):

laurelfulford:
Thanks for the review, @miguelpeixe! 🙌

PR #1934, opened 2024-10-30
feat: add support for Newspack Guest Contributor in HPP blocks

leogermani:
### All Submissions:

Adds support for the Newspack's Guest Contributor role in the Homepage posts and Carousel blocks

1. Create a user with only the Guest Contributor role and assign a post to it
2. Add a Home page posts block
3. Search by that author name and confirm it finds the author and their posts
4. Do the same for the Carousel block

### Other information:

### DIFF
--- a/includes/class-newspack-blocks.php
+++ b/includes/class-newspack-blocks.php
@@ -745,6 +745,9 @@ class Newspack_Blocks {
 						'authors'   => $authors,
 						'coauthors' => $co_authors_names,
 					];
+
+					// Also, in these cases, never offload the query to Elastic Search.
+					$args['newspack_no_es_query'] = true;
 				}
 			}
 		}
--- a/src/blocks/author-profile/class-wp-rest-newspack-authors-controller.php
+++ b/src/blocks/author-profile/class-wp-rest-newspack-authors-controller.php
@@ -142,8 +142,14 @@ class WP_REST_Newspack_Authors_Controller extends WP_REST_Controller {
 				}
 			}
 		} else {
+			$role_in = [ 'Administrator', 'Editor', 'Author', 'Contributor' ];
+			// Support for Newspack's Guest Contributor Role.
+			if ( class_exists( 'Newspack\Guest_Contributor_Role' ) ) {
+				$role_in[] = Newspack\Guest_Contributor_Role::CONTRIBUTOR_NO_EDIT_ROLE_NAME;
+			}
+
 			$user_args = [
-				'role__in' => [ 'Administrator', 'Editor', 'Author', 'Contributor' ],
+				'role__in' => $role_in,
 				'offset'   => $offset,
 				'orderby'  => 'registered',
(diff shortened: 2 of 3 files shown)


dkoo (changes requested):
I'm not sure why it's happening, but when I have ElasticSearch enabled for HPP and Carousel blocks the posts aren't getting fetched for me on the front-end. It works as expected in the editor:

Screenshot 2024-11-04 at 2 37 00 PM

But the front-end shows 0 posts for both blocks. I confirmed that the query args are the same in both contexts, and I only have a single Guest Contributor user who's assigned to a single post (but these were created/assigned many weeks ago, so they should be indexed for ElasticSearch by now). Paging @miguelpeixe who worked on the ES integration in case it's obvious to you.

leogermani:
Looks like requests in the admin are not offloaded to ES.

I'm struggling to test this. @miguelpeixe could you have a look?

leogermani:
@dkoo To be fair, this PR does not touch the Query for posts. It only touches the query for authors in the Authors filter input...

miguelpeixe:
> Paging @miguelpeixe who worked on the ES integration in case it's obvious to you.

~~That was an issue in the first implementation of the ES, which has been fixed since https://github.com/Automattic/newspack-manager/pull/275. Can you confirm you have the latest manager?~~

I may have misunderstood the issue. It works in the editor but on the front-end. If you have ES enabled for Newspack Blocks that might be a new uncovered incompatibility. We should flag that in Asana for further exploration.

miguelpeixe:
> It only touches the query for authors in the Authors filter input...

The ES incompatibility might be in `trunk`, to any query filtering CAP authors.

leogermani:
> > It only touches the query for authors in the Authors filter input...
> 
> The ES incompatibility might be in `trunk`, to any query filtering CAP authors.

They are not CAP authors. It's just another user role. But they are regular wp users

leogermani:
> I may have misunderstood the issue. It works in the editor but on the front-end. 

afaik queries are not offloaded to ES in the admin, only in the front end

leogermani:
@dkoo can you have a look again, without the ES integration? I've updated this branch with the latest trunk, there were some fixes there. Things work ok on my end.

I'll work on another PR to make sure that the queries that have the SQL statement filtered will not be offloaded to ES.

EDIT: actually it was a one liner so I added in https://github.com/Automattic/newspack-blocks/pull/1934/commits/1d1b168a88cb9b73eaf7ef4f0d6f4bad98526c37 - depends on https://github.com/Automattic/newspack-manager/pull/301

dkoo (approved):
Working now with https://github.com/Automattic/newspack-manager/pull/301!

dkoo:
@leogermani looks like we need to account for the new query param in the unit tests

leogermani:
> @leogermani looks like we need to account for the new query param in the unit tests

Fixed

PR #1933, opened 2024-10-30
Lower block cache time to 2 minutes by default

claudiulodro:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-theme/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Note: This is a hotfix PR

In my testing on a live site, there is no noticeable performance difference with a 2 minute cache time. A 5 minute edge/batcache with a 5 minute block cache can potentially make it almost 10 minutes before a new post shows on the homepage (if the timing lines up poorly). This will prevent that situation.,

### How to test the changes in this Pull Request:

1. Verify block caching works. Wait 2 minutes. Verify the cache has expired.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1935, opened 2024-10-30
fix: name branch correctly

claudiulodro:
Trivial change under a `fix/` branch name to force a release since https://github.com/Automattic/newspack-blocks/pull/1933 did not trigger it


laurelfulford (approved):
Excellent code comment! :shipit:

PR #1950, opened 2024-11-12
fix(modal-checkout): correctly generate tiered donation price summaries

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1207817176293825/1208733956648866/1208751945335388/f

This PR fixes an issue where the logic for tiered donation price summary generation was never being run since it relied on no frequency input in the donation form for this layout.

We fix this by explicitly checking for a donation tier index value in the form data and if present running that logic first.

1. As a logged out reader, with forced registration enabled select any donation value to trigger the auth modal with a price summary at the top.
2. Confirm the price summary is correct
3. Repeat the above steps for all three donation layout types (tiered, untiered, frequency) and for multiple frequencies and values

### Other information:

### DIFF
--- a/src/modal-checkout/modal.js
+++ b/src/modal-checkout/modal.js
@@ -342,49 +342,51 @@ domReady( () => {
 				);
 
 				if ( donationTiers?.length ) {
-					const frequencyInputs = form.querySelectorAll(
-						`input[name="donation_value_${ frequency }"], input[name="donation_value_${ frequency }_untiered"]`
-					);
+					const donationTierIndex = formData.get( 'donation_tier_index' );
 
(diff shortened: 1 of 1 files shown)


laurelfulford (approved):
This seems to fix the issue in my testing, and works for the different layouts 🎉 

I ran into one issue with non-tiered Donate Blocks, but I can recreate it on epic/ras-acc too, so I can file it separately. 

(It's an edge case where the way we're checking if the `price` is in the `priceSummary` can return false positives, like if you have $6 as a low price and $60 as the high price for a frequency, both will display $60 in the registration popup because `6` appears in both -- $6 matches, but so does $60).

PR #1954, opened 2024-11-13
fix: also search for coauthor posts by term slug

leogermani:
### All Submissions:

When searching for posts from an author that is a WP User (not a guest author), we should also (actually maybe only) search by term slug.

This PR adds a few extra conditions to make sure we find posts by an author when searching by terms.

The specific case we are fixing here is from users that ended up with their Display Name as the term `name`, and thus the search term name using the user_login as input would never find anything. We also need to search by user_nicename and make sure we search in the slug field.

1. Activate CoAuthors Plus
2. Create a user and assign a few posts to it
3. Create a post and assign this user as a secondary co-author (a co-author, but not the author on the `post_author` field in the database
4. On a page, add a Home Page posts block and filter by this author
5. Make sure all the posts show up, including the one they are not the main author (on `trunk` it won't show up)
6. Add this somewhere:

```
// Disable guest authors before CoAuthors plus is initialized
add_action(
	'init',
	function() {
		add_filter( 'coauthors_guest_authors_enabled', '__return_false' );
	},
	9
);
```

7. Make sure the block still fetch the correct posts.
8. Remove the snippet to re-enable Guest Authors
9. Create a guest author
10. Create a post and assign only the guest author to it.
11. You can confirm and see that the `post_author` field in the DB still points to your admin
12. In the HPP block, filter posts by the guest author and confirm it works
13. Now filter posts by your admin, and confirm the post by the guest author is NOT returned

### Other information:

### DIFF
--- a/includes/class-newspack-blocks.php
+++ b/includes/class-newspack-blocks.php
@@ -710,6 +710,7 @@ class Newspack_Blocks {
 									}
 								} else {
 									$author_names[]  = $author_data->user_login;
+									$author_names[]  = $author_data->user_nicename;
 									$author_emails[] = $author_data->user_email;
 								}
 							}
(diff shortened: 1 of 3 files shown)


dkoo (approved):
Works as described!

PR #1961, opened 2024-11-15
fix: improve appearance of password in modal checkout

laurelfulford:
### All Submissions:

While this won't be our recommended setting -- it's better if publishers require registration on purchase in the Reader Activation settings -- this PR improves the appearance of the password field when "Send password setup link" is unchecked under the WooCommerce settings.

See 1207817176293825-as-1208733956648888

1. Under WooCommerce > Settings > Accounts & Privacy, uncheck "Send password setup link (recommended)"
2. Under Newspack > Engagement > Reader Activation > Advanced Settings, toggle off "Require sign in or create account before checkout".
3. In an incognito window, run through a checkout (ideally both with a product that requires shipping and one that does not) -- note the appearance of the password section:

CleanShot 2024-11-14 at 16 45 43

CleanShot 2024-11-14 at 16 46 06

6. Repeat step 3; confirm that the spacing looks better, and the extra header is gone. Bonus points: 

CleanShot 2024-11-14 at 16 38 50

CleanShot 2024-11-14 at 16 37 56

7. Try entering a password; you'll immediately get a message about it being too short, and confirm it doesn't look too wildly out of place.

CleanShot 2024-11-14 at 16 41 22

### Other information:

### DIFF
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -89,6 +89,8 @@ final class Modal_Checkout {
 		add_action( 'wp_ajax_process_name_your_price_request', [ __CLASS__, 'process_name_your_price_request' ] );
 		add_filter( 'option_woocommerce_woocommerce_payments_settings', [ __CLASS__, 'filter_woocommerce_payments_settings' ] );
 		add_action( 'init', [ __CLASS__, 'unhook_woocommerce_payments_update_billing_fields' ] );
+		add_action( 'wp_enqueue_scripts', [ __CLASS__, 'update_password_strength_message' ], 9999 );
+
 
 		/** Custom handling for registered users. */
 		add_filter( 'woocommerce_checkout_customer_id', [ __CLASS__, 'associate_existing_user' ] );
@@ -944,6 +946,23 @@ final class Modal_Checkout {
 		return add_query_arg(
 			$args,
 			$url
+		);
+	}
+
+	/**
+	 * Update the text used for the Password Strength message from WooCommerce.
+	 */
+	public static function update_password_strength_message() {
+		wp_localize_script(
+			'wc-password-strength-meter',
+			'pwsL10n',
+			array(
+				'mismatch' => __( 'Password mismatch', 'newspack-blocks' ),
+				'short'    => __( 'Password strength: Very weak', 'newspack-blocks' ),
+				'bad'      => __( 'Password strength: Weak', 'newspack-blocks' ),
+				'good'     => __( 'Password strength: Medium', 'newspack-blocks' ),
+				'strong'   => __( 'Password strength: Strong', 'newspack-blocks' ),
+			)
 		);
 	}
 
--- a/src/modal-checkout/checkout.scss
+++ b/src/modal-checkout/checkout.scss
@@ -206,6 +206,7 @@
 
 			.woocommerce-password-strength {
 				font-size: var(--newspack-ui-font-size-xs, 14px);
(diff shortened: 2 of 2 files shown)


laurelfulford:
@thomasguillot I'd appreciate your eyes on the password length message in the last screenshot. This fix is more of a fallback thing -- publishers should have registration through RAS enabled if they want people to set passwords -- but I wanted to make sure things didn't look too off if they didn't!

Just let me know if anything should be changed!

thomasguillot:
Screenshot 2024-11-15 at 13 22 01@2x

@laurelfulford a few things:

* Add an 8px margin between the input and the text.
* Add the correct line-height
* Add the same font-size and line-height to the `small` message, with colour set to neutral-60
* Add "Password strength:" before so we know what it is
* Very weak and weak should be error-50, Medium should be warning-50, and Strong success-50

laurelfulford:
Thanks @thomasguillot! I've applied everything -- the only thing that's not 1:1 is that there isn't a `warning-50` colour in the components Figma, just `-40` -- if that's not okay, just let me know which hex to use and I can add a `-50`!

Just let me know if anything needs adjusting further!

CleanShot 2024-11-15 at 10 41 02

CleanShot 2024-11-15 at 10 41 18

CleanShot 2024-11-15 at 10 42 06

CleanShot 2024-11-15 at 10 41 50

dkoo (changes requested):
The styles look good! I'm not sure how much we want to do here, but currently the form allows you to click "continue" and then even complete checkout with a weak password (e.g. even just a single letter). But then when you go to sign in using that weak password, it won't allow you to.

How possible would it be to detect the state of the password strength meter and either disable the "Continue" button until it shows a "strong" valid password, OR return an error from the server response after you click "Continue"? If the LOE seems high we can address in a separate task/PR and only if we decide that we need to support the Woo option that shows this field at all.

laurelfulford:
Thanks @dkoo! I totally missed that -- that's not great 😕 I'll see what we can do to block hitting 'Continue' when the password isn't good enough, I think that would be a much better experience! If it turns out to be a lot, I'll handle it in a separate PR.

laurelfulford:
This is looking to be more complicated than I expected (at least for me!). I'm going to merge this PR and work on disabling the Continue button separately.

laurelfulford:
@dkoo Just setting this back to 'needs review' to get a thumbs up on the above 🙂

I honestly think forcing the registration is probably the better option than allowing the password field to display (maybe we can use disabling the "Send password setup link" as a way to also turn on forced registration in RAS?). That might just the be the frustration taking though 😆 

I did kind of get it working, but I ended up fighting (I think?) our standard form validation, which kept un-disabling the Continue button. There's probably a nice way to do it if we can hook into the existing validation.

dkoo (approved):
@laurelfulford Sounds good, let's tackle that in a separate PR. 

One thing we could try is server-side validation instead: populate a hidden input field based on the validation status of the password checker: If it's failing validation, we populate it with some value like `1`, if it's passing we leave it empty. Then, on the server-side validation we do for checking the address fields, if we get that hidden field with any value, we return a "please enter a stronger password" error.

laurelfulford:
Thanks @dkoo! That sounds like a good idea 🙂 

I feel like that may similar to how it's being handled outside of the modal checkout -- if I run through a regular /shop > /cart checkout with the password email disabled and add a short password, it's not picked up by the front-end validation, but you get a non-specific error after sending the form and it doesn't let you complete the transaction.

PR #1943, opened 2024-11-11
feat(homepage-articles): rename block and reorganise settings

thomasguillot:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-theme/blob/master/.github/CONTRIBUTING.md)?
* [ ] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

You can read more about the logic behind this change here: P2-pfaSwu-iT

The main reasons are:

- The block's functionality extends beyond just Posts for the Homepage
- The sidebar settings are disorganised, making navigation through options difficult

__Before:__

homepage-posts-settings-before

__After:__

content-loop-settings-after
content-loop-styles

### How to test the changes in this Pull Request:

1. Add a bunch of Homepage post block to a page with various settings
2. Switch to this branch
3. Check if the blocks are still displaying the correct information and has kept the correct settings

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1962, opened 2024-11-15
feat(carousel): rename block and reorganise settings

thomasguillot:
### All Submissions:

Similar to #1943, this time for the "Post Carousel" block.

1. Add a bunch of Carousel Post blocks to a page with various settings
2. Switch to this branch
3. Check if the blocks are still displaying the correct information and has kept the correct settings

### Other information:


dkoo (changes requested):
This needs a little updating to more closely match changes made in #2000:

- Rename "Settings" panel to "Display"
- Rename "Loop" panel to "Content"
- Move "Display" panel below "Content" panel

These changes would bring the sidebar panels more in line with changes to the Content Loop block in #2000.

Aside from that, there are some additional changes we could consider to this block, either in this PR or another one. Consider these non-blocking suggestions rather than required changes.

- Add option to "Allow duplicate stories" or not, and make sure it works the same way in both editor vs. front-end ([see comment](https://github.com/Automattic/newspack-blocks/pull/2000#discussion_r1937800543) for details on how it's currently slightly broken)
- Consider moving attributes from `index.js` to a `block.json` file for registering block attributes, to bring it in line with other blocks in our codebase

thomasguillot:
@dkoo made some changes: https://github.com/Automattic/newspack-blocks/pull/1962/commits/65a1b0dace5f112f82a1344c82d29214421cb68c and https://github.com/Automattic/newspack-blocks/pull/1962/commits/4d9cc5c16da05018c73f8abb14e18ad5a5ec923b

dkoo (changes requested):
@thomasguillot Thanks for the reorganization and the new block.json file! The changes here look great.

While testing this, I ran into an issue where setting block attributes that result in zero posts can crash the editor. However, I confirmed this is reproducible on `trunk` as well so it's not due to any of these changes. Filed a bug report here: `1200550061930446/1209344644455228`

dkoo (approved):

PR #1968, opened 2024-11-19
fix: do not invalidate tax queries

leogermani:
### All Submissions:

On https://github.com/Automattic/newspack-blocks/pull/1954 we changed the tax_query adding a new clause by `slug`. We also expanded the possible values we are searching for. 

As a result, we might end up generating an invalid query for one of the searches (in slug or name), and that's fine. But our condition to consider a that the whole query is invalid now needs to account that we have two clauses and that it's ok for one of them to fail.

@eddiesshop can you help me here?
1. Make sure that CAP is active.
2. Make sure that Newspack Blocks is active.
3. Create a handful of posts, assign a random author, and publish.
4. Create a new WP_User. If possible, ensure that the `user_login` does not match `user_nicename`.
5. Create another handful of posts and assign the user from the previous step as the author.
6. Create a new post and add a Homepage Post Block, filtered to the author from Step 4.
7. Observe that posts from Step 3 are being displayed for this author.
8. Switch Newspack Blocks to this PR/branch.
9. Attempt to repeat Step 6. Notice however, that now only posts assigned to the author in Step 4 are being displayed.

### Other information:

### DIFF
--- a/includes/class-newspack-blocks.php
+++ b/includes/class-newspack-blocks.php
@@ -1290,8 +1290,17 @@ class Newspack_Blocks {
 		$exclude = $exclude['where'];
 		$authors = " ( {$wpdb->posts}.post_author IN ( $csv ) $exclude ) ";
 
-		// Make sure the authors are set, the tax query is valid (doesn't contain 0 = 1).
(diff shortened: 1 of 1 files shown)


eddiesshop (approved):
I approve this change.

PR #1973, opened 2024-11-26
Fix/checkout button deprecation

laurelfulford:
### All Submissions:

This PR reworks the deprecation code in the Checkout Button block:
* It moves the content from save.js to a deprecated.js file, to follow examples I found -- after I finally got everything working, I'm not sure this was needed but the reordering that came with it helped make sure the `save` in the updated version wasn't overriding the `save` passed to deprecated (which was part of the issue).
* It makes sure the deprecated version of the block still gets the 'Continue browsing' string, and not the newer 'Continue'. This means that they continue to have the label 'Continue browsing' which may not be ideal 😬 If there's a better way to do this that switches the string, let me know!

This is one of those fun ones where I have no idea why it works, but it seems to, so any nitpicky feedback is welcome!

See: 1207817176293825-as-1208842146441487

1. Switch to the /trunk branches of all of the plugins.
2. Set up a bunch of Checkout Button blocks with different settings, including having a few with different 'After purchase' settings (primarily the button label). Also change colours, widths, border radius, etc. Make sure at least one is in a group/columns block.
3. Save.
4. Switch the branches to epic/ras-acc.
5. Refresh the editor; note you have a bunch of errors around block validation:

CleanShot 2024-11-26 at 08 59 09

6. Apply this PR and run `npm run build`.
7. Refresh the editor again; confirm that the errors are gone and the buttons are previewing correctly. If you have the console open, you should also get a bunch of messages like this indicating the blocks were successfully updated:

CleanShot 2024-11-26 at 09 02 12

8. Try making a non-block edit and save (to rule out the issue Source reported: nested, non-recovered blocks disappeared if the page they were on was saved without fixing them). 
9. Confirm that all the blocks are still in place on the front-end and in the editor.
10. Test the buttons to make sure things work as expected -- in particular note if your settings are correct (like any button labels you've changed). 
11. Smoke test things by adding a few new buttons to the page and make sure everything works as expected.
12. Hammer anything else you can think of. 

### Other information:


chickenn00dle (changes requested):
Looks good for the most part @laurelfulford!

Screenshot 2024-11-26 at 16 22 23

Just one small issue where updated buttons don't respect the color set in the editor:

Editor:
Screenshot 2024-11-26 at 16 24 47

Front-end:
Screenshot 2024-11-26 at 16 24 30

laurelfulford:
Oh, good catch @chickenn00dle! 

This looks like an issue in RAS-ACC, full stop (I can recreate it on the epic branch) but I think it makes sense to fix in this PR since it's addressing other button-y things 🙂

laurelfulford:
The gradient issue should be fixed in https://github.com/Automattic/newspack-blocks/pull/1973/commits/9e6d6a7f8eaeeac5b7a2942d4b087554bc0c7bb3 -- it looks like we weren't looking for that style at all, so any of the default gradients baked into the theme wouldn't be applied on the front-end.

Best way to test would be to try applying a few of the default gradients, both with the default blue and again with some custom colours set in the Customizer, and make sure they're applied on the front-end 🙂 

Just let me know if you have any questions or if anything looks off!

chickenn00dle (approved):
Nice! Works as expected now!

laurelfulford:
Awesome, thanks @chickenn00dle! 🙌

PR #1984, opened 2024-12-03
fix: remove reCAPTCHA for Woo from modal checkout

laurelfulford:
### All Submissions:

This PR removes and dequeues code from the reCAPTCHA for WooCommerce plugin from our modal checkout.

See 1207817176293825-as-1208873550136264

Note: the way I went about removing the scripts might be terrible; it seemed kind of duplicative, but I couldn't think of a better way that didn't require spelling out each allowed script. 

1. Install reCAPTCHA for WooCommerce and enable it on your site.
2. Try running through a modal checkout; note you get an error.
3. Apply this PR.
4. Confirm that you can complete the modal checkout.
5. View source in the modal checkout and confirm you don't have anything referencing `recaptcha` from the plugin (a `g-recaptcha` div, plus the script, rcfwc.js).
6. Confirm that those elements still exist in the regular checkout and that the reCAPTCHA there doesn't throw errors.

### Other information:

### DIFF
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -133,10 +133,8 @@ final class Modal_Checkout {
 		add_filter( 'woocommerce_subscriptions_product_limited_for_user', [ __CLASS__, 'subscriptions_product_limited_for_user' ], 10, 3 );
 		add_filter( 'woocommerce_get_privacy_policy_text', [ __CLASS__, 'woocommerce_get_privacy_policy_text' ], 10, 2 );
 
-		// Remove any hooks related to reCAPTCHA for WooCommerce.
-		if ( self::is_modal_checkout() ) {
-			add_action( 'plugins_loaded', [ __CLASS__, 'remove_recaptcha_hooks' ] );
-		}
+		// Remove any hooks that aren't supported by the modal checkout.
+		add_action( 'plugins_loaded', [ __CLASS__, 'remove_hooks' ] );
 	}
 
 	/**
@@ -761,7 +759,7 @@ final class Modal_Checkout {
 		foreach ( $wp_scripts->queue as $handle ) {
 			$allowed = false;
 			foreach ( $allowed_assets as $allowed_asset ) {
-				if ( false !== strpos( $handle, $allowed_asset, 0 ) ) {
+				if ( 0 === strpos( $handle, $allowed_asset, 0 ) ) {
 					$allowed = true;
 					break;
 				}
@@ -773,7 +771,7 @@ final class Modal_Checkout {
 		foreach ( $wp_styles->queue as $handle ) {
 			$allowed = false;
 			foreach ( $allowed_assets as $allowed_asset ) {
-				if ( false !== strpos( $handle, $allowed_asset, 0 ) ) {
+				if ( 0 === strpos( $handle, $allowed_asset, 0 ) ) {
 					$allowed = true;
 					break;
 				}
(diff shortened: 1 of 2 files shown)


includes/class-modal-checkout.php
```diff
@@ -744,19 +749,25 @@ public static function dequeue_scripts() {
...
+			'rcfwc-js',
+		];
+
 		/**
 		 * Filters the allowed assets to render in the modal checkout
 		 *
 		 * @param string[] $allowed_assets Array of allowed assets handles.
 		 */
 		$allowed_assets = apply_filters( 'newspack_blocks_modal_checkout_allowed_assets', $allowed_assets );
+		$skip_assets    = apply_filters( 'newspack_blocks_modal_checkout_skip_assets', $skip_assets );
```
chickenn00dle: Hmm, rather than adding another array for skipped assets, can we instead fix the check for `allowed_assets` below so that this script and any others with similar situations where the allowed asset string is found but the script itself is actually not expected?

includes/class-modal-checkout.php
```diff
@@ -744,19 +749,25 @@ public static function dequeue_scripts() {
...
 
 		global $wp_scripts, $wp_styles;
 
 		foreach ( $wp_scripts->queue as $handle ) {
 			$allowed = false;
 			foreach ( $allowed_assets as $allowed_asset ) {
-				if ( false !== strpos( $handle, $allowed_asset ) ) {
+				if ( false !== strpos( $handle, $allowed_asset ) && ! in_array( $handle, $skip_assets ) ) {
```
chickenn00dle: See my comment above. This would be where we want to fix the check for `allowed_assets`. I think what we want to do here is ensure the `strpos` is always equal to 0 to indicate the script begins with the `allowed_asset` string rather than just existing anywhere within it.

includes/class-modal-checkout.php
```diff
@@ -779,6 +790,19 @@ public static function dequeue_scripts() {
 		}
 	}
 
+	/**
+	 * Remove hooks related to reCAPTCHA for WooCommerce
+	 */
+	public static function remove_recaptcha_hooks() {
```
chickenn00dle: Question: How common is this extension? It feels strange to have to account for specific 3rd party plugin hooks like this within our own plugin. It also doesn't seem very sustainable for other cases that might be impacting checkout down the line.

Is it possible to do this more dynamically using an approach similar to how we handle allowed scripts?

chickenn00dle (changes requested):

includes/class-modal-checkout.php
laurelfulford: > How common is this extension?

It's active on 28 sites now (and installed but inactive on others) -- definitely more than I would have expected! 

> Is it possible to do this more dynamically using an approach similar to how we handle allowed scripts?

Like pass a filterable array of hooks to remove? I can dig into it!

laurelfulford:
Ok I think I have everything sorted, @chickenn00dle! 

As an extra precaution, I did some error logging for the scripts/styles getting removed before and after this change, and re-added a couple -- one (the woocommerce styles one) was causing issues. I didn't see any issues caused by selectWoo, but better safe than sorry! (It's Woo's accessible fork of select2, which was also being enqueued [I think] as a fallback).

For the remove_action part, I added an array with the hook and callback that can be filtered using `newspack_blocks_modal_checkout_remove_hooks`. 

Just let me know if anything needs changing -- thanks!

includes/class-modal-checkout.php
```diff
@@ -779,6 +784,54 @@ public static function dequeue_scripts() {
...
+				'hook'     => 'woocommerce_before_checkout_billing_form',
+				'callback' => 'rcfwc_field_checkout',
+			],
+			[
+				'hook'     => 'woocommerce_after_checkout_billing_form',
+				'callback' => 'rcfwc_field_checkout',
+			],
+			[
+				'hook'     => 'woocommerce_review_order_before_submit',
+				'callback' => 'rcfwc_field_checkout',
```
chickenn00dle: Another (non-blocking question) I had a look through the code and am seeing we have this array of fields to filter out during validation in the modal checkout js: https://github.com/Automattic/newspack-blocks/blob/f398d8c69ba1322ae0b57b28a27a5e561c3b3815/src/modal-checkout/index.js#L618-L620

For this hook, which I'm assuming is adding a checkout field, is this something that can be removed using the same strategy?

We'd still need to remove hooks associated with any backend logic of course, but if we can stick to the same strategy for bypassing field validation that'd be great!

includes/class-modal-checkout.php
```diff
@@ -779,6 +784,54 @@ public static function dequeue_scripts() {
...
+				'callback' => 'rcfwc_checkout_check',
+			],
+		];
+
+		/**
+		 * Filters the hooks to remove from the modal checkout.
+		 *
+		 * @param string[] $remove_list Array of hooks to remove.
+		 */
+		$remove_list = apply_filters( 'newspack_blocks_modal_checkout_remove_hooks', $remove_list );
```
chickenn00dle, line 837: Another non-blocking comment. This is definitely better. But in the end we are still needing to specify the `rcfwc_` callbacks here. This is fine if this is the only one, but I can't imagine we don't run into similar issues in the future with other extensions.

I haven't dug too deep into it, but it's possible to view all hooks associated with a specific action via the `$wp_filter` global. Is it possible to do something like create an allow list of plugins, and check the added filters for the checkout processing hook(s), then remove if not in the allow list?

I'm happy to go with this quick fix specific to recaptcha for woocommerce, but really feel this should be improved so we don't have to keep accounting for one-off plugins like this.

chickenn00dle (changes requested):
Hmm, I tested with the changes but am still seeing the recaptcha error on my end:
Screenshot 2024-12-04 at 10 58 59

Not sure if its the same extension but I am testing with this one: https://woocommerce.com/products/recaptcha-for-woocommerce/

I also left two small comments that are non-blocking (sort of). Let me know what you think!

includes/class-modal-checkout.php
laurelfulford: > For this hook, which I'm assuming is adding a checkout field, is this something that can be removed using the same strategy?

I'm not sure of the ins and outs of how this works, but it's adding a `div`, like:

`<div class="g-recaptcha" data-sitekey="6LfrhJEqAAAAAPuv8JoMdAqEM6G-TKQxc2Es_37A"></div>`

I double-checked, and it doesn't look like anything related to reCAPTCHA is included in `$form.serializeArray()`, which is what we're checking `removeFromValidation` for.

includes/class-modal-checkout.php
laurelfulford: I think for now (and in the interest of getting this fixed and rolled out during alphas), my vote would be to keep it simple and revisit when we need to add more plugins 😅 I know something more abstracted would be ideal but I'm worried about cutting this too close to the wire.

laurelfulford: I'm going to spin up a separate issue to capture this and see what I can do with it! 🙂

laurelfulford:
I've updated the code to actually work with the correct ReCaptcha for WooCommerce plugin 😆 

Anything related to the plugin should be dequeued/removed in the modal checkout, but still be present in the regular checkout. The plugin uses the prefix `i13` for scripts and elements it inserts.

includes/class-modal-checkout.php
```diff
@@ -791,34 +791,40 @@ public static function remove_hooks() {
...
-			[
-				'hook'     => 'woocommerce_checkout_process',
-				'callback' => 'rcfwc_checkout_check',
-			],
-		];
 
+		$remove_list = [];
+
+		// reCaptcha for WooCommerce.
+		if ( method_exists( 'I13_Woo_Recpatcha', '__construct' ) ) {
```
chickenn00dle: non-blocking nit: we can just do a check for the class with `class_exists`.

chickenn00dle (approved):
Looks good. Thanks @laurelfulford

includes/class-modal-checkout.php
laurelfulford: Thanks Raz! Fixed in [ad577d6](https://github.com/Automattic/newspack-blocks/pull/1984/commits/ad577d6e040e72dacb494191da2166c3ec4376a3).

PR #1985, opened 2024-12-03
fix(modal-checkout): handle paypal

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1207817176293825/1208897296735259/f

We have a handful of publishers using modal checkout with paypal however RAS-ACC breaks this experience. This PR ensures paypal can render and be used in modal checkout.

Screenshot 2024-12-04 at 15 47 36

1. Install and set up paypal payments (See https://woocommerce.com/document/woocommerce-paypal-payments/#testing-in-sandbox for steps to get set up and note you will need to enable a few settings before paypal buttons become enabled for subscription products.)
2. Make sure PayPal is set up as one of our payment gateways
3. On `release`, go through checkout via any donate or checkout button and confirm you can checkout with paypal
4. On this branch, repeat the last step again confirming you can checkout with paypal
5. Repeat these steps for the various paypal buttons

Note: Some of the paypal checkout flows are funky on `release` already. This PR does not fix this funkiness, but instead just makes sure there is feature parity with paypal working generally.

### Other information:

### DIFF
--- a/src/modal-checkout/checkout.scss
+++ b/src/modal-checkout/checkout.scss
@@ -75,7 +75,7 @@
 
 			// PayPal payments buttons.
 			#ppc-button-ppcp-gateway {
-				margin-bottom: var(--newspack-ui-spacer-5, 24px);
+				margin-bottom: var(--newspack-ui-spacer-2, 12px);
 			}
 		}
 
--- a/src/modal-checkout/checkout.scss
+++ b/src/modal-checkout/checkout.scss
@@ -73,8 +73,13 @@
 				margin-bottom: var(--newspack-ui-spacer-2, 12px);
 			}
 
-			// PayPal payments buttons.
+			// PayPal Payments buttons.
 			#ppc-button-ppcp-gateway {
+				margin-top: var(--newspack-ui-spacer-5, 24px);
+			}
+			#ppc-button-ppcp-gateway,
+			#ppc-button-applepay-container,
+			#ppc-button-googlepay-container {
 				margin-bottom: var(--newspack-ui-spacer-2, 12px);
 			}
 		}
--- a/src/modal-checkout/checkout.scss
+++ b/src/modal-checkout/checkout.scss
@@ -72,6 +72,11 @@
 			h3 {
 				margin-bottom: var(--newspack-ui-spacer-2, 12px);
 			}
+
+			// PayPal payments buttons.
+			#ppc-button-ppcp-gateway {
+				margin-bottom: var(--newspack-ui-spacer-5, 24px);
+			}
 		}
 
 		// Override label styles from theme
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -745,10 +745,9 @@ final class Modal_Checkout {
 			'stripe',
 			'select2',
 			'selectWoo',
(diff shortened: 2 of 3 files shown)


src/modal-checkout/checkout.scss
```diff
@@ -72,6 +72,11 @@
 			h3 {
 				margin-bottom: var(--newspack-ui-spacer-2, 12px);
 			}
+
+			// PayPal payments buttons.
+			#ppc-button-ppcp-gateway {
+				margin-bottom: var(--newspack-ui-spacer-5, 24px);
```
laurelfulford:
```suggestion
				margin-bottom: var(--newspack-ui-spacer-2, 12px);
```

I noticed some spacing weirdness depending on the button settings -- we might not be able to get them all in one go, but:

* The PayPal button has 24px space between it and the Back button; the Complete Order button only had 12px.

CleanShot 2024-12-04 at 14 52 08

I'm not getting any space above the button when the terms are visible: 

CleanShot 2024-12-04 at 16 17 21

When enabled, both the Apple Pay and Google Pay buttons touch the 'Back' button.

CleanShot 2024-12-04 at 16 25 18

laurelfulford (commented):
The Paypal button works well in isolation in my testing; I've noted some visual issues but I think we can handle them separately.

The other buttons I managed to get to appear were:

Google Pay -- this successfully let me run a test transaction, and showed me the success message.

Apple Pay -- this briefly showed a payment processing message and then closed, never changing the appearance of the modal checkout. As discussed, the same issue exists on release so I think we're OK there.

Debit/Credit card payment fields: there are a number of display issues with this option and it doesn't seem to have a test mode, so I can't run a transaction with it. I think we can handle those separately though, since that isn't really the primary use case for Paypal:

CleanShot 2024-12-05 at 12 17 21

CleanShot 2024-12-05 at 12 17 46

Venmo Payment: I was only able to get this button when I switched my store to be US-based. On transaction, I got a "Venmo Sandbox - Venmo is working! Tap the button below and sign in with PayPal Sandbox Account" message, which basically let me pay through Paypal. I don't get the point, but it worked! 😆 

I wasn't able to get the black "Debit or Credit Card" appearing at the bottom of the modal checkout at all like in your screenshot. For this is might be worth having an American test, but this otherwise worked for me!

I'm leaving this as a comment instead of an approval in hopes someone US based can test that last Debit/Credit card option. 

Of the visual feedback, I think the space above the PayPal button would be nice to fix, but the rest can be handled separately.

chickenn00dle:
@dkoo I hope it's okay, but I'm assigning you for review to get a US based engineers eyes on this per @laurelfulford's last comment.

src/modal-checkout/checkout.scss
chickenn00dle: Thanks @laurelfulford!

I added some styles to fix these issues. Let me know if this works. 

I actually have a hard time getting the google button to appear so haven't been able to visually confirm that one.

laurelfulford:
@chickenn00dle The spacing looks better around the Paypal button -- thanks @chickenn00dle! 🎉 

Nothing's outstanding for me!

dkoo (approved):
Tests well with both PayPal and Venmo!

PR #1989, opened 2024-12-06
feat: add CSS class to variation buttons for tracking

laurelfulford:
### All Submissions:

This PR adds a unique CSS class only used by variation picker buttons in the modal checkout to make them easier to apply tracking to. (I am ABSOLUTELY open to feedback on the name since we won't want to change it!!)

See: 1207817176293825-as-1208919248362608

1. Apply this PR.
2. Set up a checkout button block for a product with variations.
3. Run through the checkout and inspect the buttons on the first step; make sure they have the CSS class `newspack-modal-variation-selection`.

### Other information:

### DIFF
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -641,7 +641,7 @@ final class Modal_Checkout {
 										<div class="variation"><?php echo esc_html( $variation_name ); ?></div>
 										<form data-product="<?php echo esc_attr( wp_json_encode( $product_data ) ); ?>">
 											<input type="hidden" name="newspack_checkout" value="1" />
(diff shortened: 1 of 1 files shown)


chickenn00dle (approved):
Looks good!

My only two non-blocking suggestion would be if we want to ensure this is unique, let's make it an `id` instead of a `class` and maybe go with `newspack-modal-checkout-vairaion-selection`

laurelfulford:
Thanks @chickenn00dle! I updated the class to `newspack-modal-checkout-variation-selection`. 

The downside with using an ID is that there will be multiples of these together in a modal -- they won't actually be unique from each other. The class should be enough so conversion tracking for Parsely and others won't need to rely on the parent elements to select these buttons, just this class 🙂

PR #1997, opened 2024-12-10
fix: remove CAP compatibility mode

leogermani:
### All Submissions:

This PR removes the so-called "Compatibility mode" for CoAuthors Plus.

This means that if CAP is enabled, we will rely only on the `author` taxonomy to search for posts.

Compatibility mode means that we were combining the search by author terms with the search by `post_author`, resulting in a very complex and fragile query. The reason for compatibility mode to exist is to make sure we are going to find posts when CAP plugin was recently activated and existing posts have not been populated with `author` terms yet.

This can be done by executing `wp co-authors-plus create-author-terms-for-posts`.

By doing this, we are fixing a number of bugs in the query. For example, before this PR, it was not possible to filter by author AND category (or any other taxonomy)

Since this plugin completely rewrites the way we filter posts by authors, we need to test a few different scenarios

#### Without CAP

Deactivate CAP and perform a few filters by author. Confirm it works as expected

#### With CAP

* Note*: There are some bugs or strange behaviors in the way we fetch the authors list to build the filter. For example, when you link a Guest Author to a WP user, you can't search by the guest author name, you have to search by the wp user name - if both names are completely different, the search won't work. This PR does not touch that.

1. regular searches

* Create a post with more than one author.
* On a page, add a HPP block and filter by one or more of the authors of that post. Make sure the post is returned.
* Combine the search with another taxonomy and make sure it works

2. coauthor

Set more than one author to a post and then try to search by a coauthor (one of the authors that is not the "main" author (the one in the `post_author` column in the DB). Make sure the block finds the post

3. Guest Author

* Create a couple of guest authors.
* Assign a guest author as one of the authors of a post
* try to filter by this guest author and make sure it finds the post

4. Solo Guest Author
* Create a post and assign it only to the guest author
* in a HPP block, filter by your admin user (not the guest author)
* Confirm that post does NOT show up 
* (we test this because your admin user is still listed as the `post_author` in the db table)

5. Linked guest author
* Link one of your guest authors to an existing WP user that has posts
* in the HPP block, filter by this guest author (see note above on bugs for the authors selectors - to avoid that, create the guest author with a similar name)
* Make sure the posts by the linked wp user are returned
* Now create a new post and assign it only to this guest author
* in the HPP block, try to filter by the real wp user. confirm that both posts by the user and guest author are returned

### Other information:

### DIFF
--- a/includes/class-newspack-blocks.php
+++ b/includes/class-newspack-blocks.php
@@ -27,26 +27,6 @@ class Newspack_Blocks {
 	const SQL_PATTERN = '/\/\* newspack-blocks \*\/(.|\n)*\/\* \/newspack-blocks \*\//';
 
 	/**
-	 * Class property to store user IDs and CAP guest author names for building
-	 * custom SQL statements. In order to allow a single WP_Query to filter by
-	 * both WP users and CAP guest authors (a taxonomy), we need to directly
-	 * modify the JOIN and WHERE clauses in the SQL query.
-	 *
-	 * If this property is false, then the custom statements will be stripped
-	 * from all SQL clauses. If it's an array with `authors` and `coauthors`
-	 * keys, the custom statements will be added to the SQL query.
-	 *
-	 * Example array:
-	 * [
-	 *     'authors'   => [], // Array of numeric WP user IDs.
-	 *     'coauthors' => [], // Array of CAP guest author name slugs.
-	 * ]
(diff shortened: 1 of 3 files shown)


dkoo (approved):
Just works, and the code is a lot easier to parse without the query/clause filtering!

PR #2000, opened 2024-12-12
feat(homepage-articles): rename block and reorganise settings

thomasguillot:
### All Submissions:

See https://github.com/Automattic/newspack-blocks/pull/1943 for details

I just a few more changes based on the feedback received in p2

See https://github.com/Automattic/newspack-blocks/pull/1943 for details

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/edit.tsx
+++ b/src/blocks/homepage-articles/edit.tsx
@@ -33,19 +33,18 @@ import {
 	AlignmentControl,
 } from '@wordpress/block-editor';
 import {
+	BaseControl,
 	Button,
 	ButtonGroup,
 	PanelBody,
-	PanelRow,
+	Path,
+	Placeholder,
 	RangeControl,
+	Spinner,
+	SVG,
 	Toolbar,
 	ToggleControl,
 	TextControl,
-	Placeholder,
-	Spinner,
-	BaseControl,
-	Path,
-	SVG,
 } from '@wordpress/components';
 import { withDispatch, withSelect } from '@wordpress/data';
 import { compose } from '@wordpress/compose';
@@ -59,6 +58,7 @@ import {
 	postFeaturedImage,
 	pullLeft,
 	pullRight,
+	textColor as typeScaleIcon,
 } from '@wordpress/icons';
 
 let IS_SUBTITLE_SUPPORTED_IN_THEME: boolean;
@@ -99,6 +99,12 @@ const squareIcon = (
 		/>
 	</SVG>
 );
+
+const typeScaleIconSmall = (
+	<SVG xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24">
+		<Path d="M13 8H11.6667L9 15.3333H10.2667L11 13.3333H13.8L14.5333 15.3333H15.8L13 8ZM11.3333 12.3333L12.3333 9.06667L13.4667 12.3333H11.3333Z"	/>
+	</SVG>
+)
 
 class Edit extends Component< HomepageArticlesProps > {
 	renderPost = ( post: Post ) => {
@@ -267,7 +273,7 @@ class Edit extends Component< HomepageArticlesProps > {
 	};
 
 	renderInspectorControls = () => {
-		const { attributes, setAttributes, textColor, setTextColor } = this.props;
+		const { attributes, setAttributes } = this.props;
 
 		const {
 			authors,
@@ -277,7 +283,6 @@ class Edit extends Component< HomepageArticlesProps > {
 			includeSubcategories,
 			customTaxonomies,
 			columns,
-			colGap,
 			postType,
 			showImage,
 			showCaption,
@@ -293,7 +298,6 @@ class Edit extends Component< HomepageArticlesProps > {
 			readMoreLabel,
 			excerptLength,
 			showSubtitle,
-			typeScale,
 			showDate,
 			showAuthor,
 			showAvatar,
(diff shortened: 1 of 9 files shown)


src/blocks/homepage-articles/edit.tsx
```diff
@@ -434,107 +398,163 @@ class Edit extends Component< HomepageArticlesProps > {
...
+						<RangeControl
+							label={ __( 'Max number of words in excerpt', 'newspack-blocks' ) }
+							value={ excerptLength }
+							onChange={ ( value: number ) => setAttributes( { excerptLength: value } ) }
+							min={ 10 }
+							max={ 100 }
+							__next40pxDefaultSize
+						/>
+					) }
+					{ ! showFullContent && (
```
dkoo, line 479: **Bug:** If the display setting is "Full Post", this hides the toggle control for showing the "more" link, but if that's already toggled on, it still appears below the full post content.

To fix this I think we'd just need to change [this condition in the template](https://github.com/Automattic/newspack-blocks/blob/update/homepage-posts-block-improvements/src/blocks/homepage-articles/templates/article.php#L156) from:

```php
			if ( $post_link && ( $attributes['showReadMore'] ) ) :
```

to:

```php
			if ( $post_link && ! $attributes['showFullContent'] && $attributes['showReadMore'] ) :
```

src/blocks/homepage-articles/edit.tsx
```diff
@@ -434,107 +398,163 @@ class Edit extends Component< HomepageArticlesProps > {
...
+								variant={ ! showExcerpt && showFullContent && 'primary' }
+								aria-pressed={ ! showExcerpt && showFullContent }
+								onClick={ () => {
+									setAttributes( {
+										showFullContent: ! showFullContent,
+										showExcerpt: showExcerpt ? false : showExcerpt
+									} )
+								} }
+							>
+								{ __( 'Full Post', 'newspack-blocks' ) }
```
dkoo, line 465: Non-blocking suggestion:

```suggestion
								{ __( 'Full Content', 'newspack-blocks' ) }
```

src/components/editor-panels.js
```diff
@@ -50,7 +49,7 @@ export const PostTypesPanel = ( { attributes, setAttributes } ) => {
 	} );
 
 	return (
-		<PanelBody title={ __( 'Post Types', 'newspack-blocks' ) }>
+		<PanelBody title={ __( 'Post Types', 'newspack-blocks' ) } initialOpen={ false }>
 			<CheckboxesGroup
```
dkoo, line 64: This may be true of `trunk` too, but if you de-select all options, it behaves the same as selecting "Posts" only. Maybe we should force the checkbox for "Posts" to be selected in this scenario.

src/blocks/homepage-articles/edit.tsx
```diff
@@ -387,45 +381,15 @@ class Edit extends Component< HomepageArticlesProps > {
...
-												</Button>
-											);
-										} ) }
-									</ButtonGroup>
-								</PanelRow>
-							</BaseControl>
-						</Fragment>
-					) }
+					<ToggleControl
+						label={ __( 'Allow duplicate stories', 'newspack-blocks' ) }
```
dkoo, line 385: Not related to these changes, so no action needed on this. But noting so I don't forget.

If you add a Post Carousel block above the Content Loop and set both to NOT allow duplicate stories, the stories shown in the editor vs. the front-end will differ:

- In the editor, the Content Loop block will show the stories shown in the Post Carousel block
- On the front-end, the Content Loop will NOT show the stories shown in the Post Carousel block

src/components/autocomplete-tokenfield.js
```diff
@@ -181,6 +181,7 @@ class AutocompleteTokenField extends Component {
 					onChange={ tokens => this.handleOnChange( tokens ) }
 					onInputChange={ input => this.debouncedUpdateSuggestions( input ) }
 					label={ label }
+					__next40pxDefaultSize
```
dkoo, line 184: Not a requirement, but while we're touching this code, should we replace this `AutocompleteTokenField` component with the one imported from `newspack-components`? I tested this locally and it seemed to work without any issues. It results in a slightly smaller visual size for the autocomplete fields:

Screenshot 2025-01-31 at 12 26 57 PM

If we decide to do this, we can delete this file and its SCSS file and change the import in `query-controls.js`.

src/components/query-controls.js
```diff
@@ -255,11 +257,33 @@ class QueryControls extends Component {
 		return (
 			<>
 				{ enableSpecific && (
-					<ToggleControl
-						checked={ specificMode }
-						onChange={ onSpecificModeChange }
-						label={ __( 'Choose Specific Posts', 'newspack-blocks' ) }
-					/>
+					<BaseControl
+						label={ __( 'Type', 'newspack-blocks' ) }
```
dkoo: Non-blocking suggestion:

```suggestion
						label={ __( 'Mode', 'newspack-blocks' ) }
```

src/blocks/homepage-articles/edit.tsx
```diff
@@ -434,107 +398,163 @@ class Edit extends Component< HomepageArticlesProps > {
...
-										checked={ infiniteScroll }
-										onChange={ () => setAttributes( { infiniteScroll: ! infiniteScroll } ) }
-									/>
-								) }
+								<ToggleControl
+									label={ __( 'Infinite scroll', 'newspack-blocks' ) }
+									checked={ infiniteScroll }
+									disabled={ ! moreButton }
+									onChange={ () => setAttributes( { infiniteScroll: ! infiniteScroll } ) }
+								/>
```
dkoo, line 419: Here's a suggestion that I thought of while testing this with an existing instance of the Homepage Posts block. I had forgotten I had created this instance of the block while testing #1968, so it had author-related query attributes that are no longer valid. These attributes are no longer handled by the UI, so the "Authors" field appeared blank, but it was still querying posts by CAP author terms that were stored in the block attributes, and I couldn't figure out why this was happening until I debug-logged the query arguments on the back-end. 

All this to say, maybe if any of these Content query attributes are set to something other than the defaults, we could consider showing a button to restore default settings for this panel? Otherwise, the only way to fix a scenario like the above is to delete and recreate the block (losing all other display-related or style attributes you had set), or enter Code mode and delete the invalid attributes (no one but developers would probably be comfortable doing this). WDYT? Could be done in a separate PR if you prefer.

dkoo (changes requested):
@thomasguillot this is still really clean and the changes are much-needed and more user-friendly than before. I ran into a minor issue and had some non-blocking suggestions which you're welcome to dismiss if you disagree.

src/components/editor-panels.js
thomasguillot: Done here: https://github.com/Automattic/newspack-blocks/pull/2000/commits/92b1e3d5d16756481036b06fd116583e13d1e9c6

Let me know if the logic/code make sense.

src/blocks/homepage-articles/edit.tsx
thomasguillot: Unfortunately, "Content" is too long.
Screenshot 2025-02-06 at 13 58 56@2x

thomasguillot: Unless I remove "Full"?

Screenshot 2025-02-06 at 14 33 50@2x

src/blocks/homepage-articles/edit.tsx
thomasguillot: Oops! https://github.com/Automattic/newspack-blocks/pull/2000/commits/d5feefe8479875a205cbaf541e7e862fe81507b1

src/blocks/homepage-articles/edit.tsx
thomasguillot: This should probably be its own issue?

src/components/autocomplete-tokenfield.js
thomasguillot: I think I'd want to revisit the `AutocompleteTokenField` from `newspack-components` first. It's probably slightly outdated (enhanced why it doesn't support `__next40pxDefaultSize`)

src/blocks/homepage-articles/edit.tsx
thomasguillot: I have a list of enhancements:

* When using the Query Loop type, place the Filters (Authors, Categories, etc.) in a [ToolsPanel](https://href.li/?https://wordpress.github.io/gutenberg/?path=/docs/components-experimental-toolspanel--docs) (similar to the Query block). This would allow users to select only the filters they need and make resetting easier.
* Add background color support to the block
* Support typography settings instead of the custom Type Scale setting
* Replace the Grid Spacing with Core’s Block Spacing
* Use [ToggleGroupControl](https://href.li/?https://wordpress.github.io/gutenberg/?path=/docs/components-experimental-togglegroupcontrol--docs) instead of the custom ButtonGroup I created
* Move the “Read More” link text setting inline, like the “Load more posts” button

And I think using the ToolsPanel would help restoring these settings to default.

dkoo: Those enhancements sound great! Seems appropriate to tackle that in a separate PR.

src/blocks/homepage-articles/edit.tsx
dkoo: Logged in `1200550061930446/1209344566417561`

src/blocks/homepage-articles/edit.tsx
dkoo: I'm okay with keeping it "Full post" since it's clearer to me than just "Content"

src/components/editor-panels.js
dkoo: Works for me!

src/components/editor-panels.js
```diff
@@ -2,30 +2,40 @@
  * WordPress dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { Spinner, CheckboxControl, PanelBody, PanelRow } from '@wordpress/components';
+import { BaseControl, CheckboxControl, PanelBody, Spinner } from '@wordpress/components';
 import { useSelect } from '@wordpress/data';
 
-const CheckboxesGroup = ( { options, values, onChange } ) => {
+const CheckboxesGroup = ( { options, values, onChange, defaultRequiredSlug } ) => {
```
dkoo: Nit: offer a default `null` value to show that this prop isn't required

```suggestion
const CheckboxesGroup = ( { options, values, onChange, defaultRequiredSlug = null } ) => {
```

src/components/editor-panels.js
```diff
@@ -2,30 +2,40 @@
...
 	if ( ! Array.isArray( options ) ) {
 		return <Spinner />;
 	}
-	return options.map( ( { name, slug } ) => (
-		<PanelRow key={ slug }>
+	return options.map( ( { name, slug } ) => {
+		const isDefaultType = slug === defaultRequiredSlug;
+		const otherTypesSelected = values.length > 1 || ( values.length === 1 && values[0] !== defaultRequiredSlug );
+		const isDisabled = isDefaultType && !otherTypesSelected;
```
dkoo: Nit: explicitly account for `defaultRequiredSlug` prop not being passed, and simplify check for other selected options:

```suggestion
		const isDefault = defaultRequiredSlug && slug === defaultRequiredSlug;
		const otherOptionsSelected = defaultRequiredSlug && values.some( value => value !== defaultRequiredSlug );
		const isDisabled = isDefault && ! otherOptionsSelected;
```

src/components/editor-panels.js
```diff
@@ -2,30 +2,40 @@
...
 					if ( value && cleanPostType.indexOf( slug ) === -1 ) {
 						cleanPostType.push( slug );
 					} else if ( ! value && cleanPostType.indexOf( slug ) > -1 ) {
 						cleanPostType.splice( cleanPostType.indexOf( slug ), 1 );
 					}
+					// If no post types would be selected, force the default required one
+					if ( cleanPostType.length === 0 ) {
+						cleanPostType.push( defaultRequiredSlug );
+					}
 					onChange( cleanPostType );
```
dkoo: Nit: tweak variable name and comment to clarify that this component can be used for data types other than post types, too (since it's used for the "additional post statuses" panel as well):

```suggestion
					const cleanOptions = [ ...new Set( values ) ];
					if ( value && cleanOptions.indexOf( slug ) === -1 ) {
						cleanOptions.push( slug );
					} else if ( ! value && cleanOptions.indexOf( slug ) > -1 ) {
						cleanOptions.splice( cleanOptions.indexOf( slug ), 1 );
					}
					// If no options would be selected, force the default required one
					if ( defaultRequiredSlug && cleanOptions.length === 0 ) {
						cleanOptions.push( defaultRequiredSlug );
					}
					onChange( cleanOptions );
```

src/components/editor-panels.js
dkoo: I have some nitpicky/non-blocking suggestions below for these changes, let me know what you think

dkoo (approved):
@thomasguillot thanks for the changes and feedback. Pre-approving this with some non-blocking suggestions—feel free to accept or reject, and merge when you decide!

thomasguillot:
Thank you so much for your suggestions (and review), @dkoo ❤️

PR #2003, opened 2024-12-16
Release Dec 16

miguelpeixe:
A new release.


dkoo (approved):
Noting that the author query changes might mean a breaking change for some sites until they run the cron job to add missing author terms.

PR #2007, opened 2024-12-17
fix(modal-checkout): force checkout registration option

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1205234045751551/1208932231357058/f

This PR forces the value of the WC checkout registration setting to true in modal checkout context. This is because we always register new accounts when one doesn't exist and when this setting is disabled, modal checkout breaks.

Screenshot 2024-12-17 at 15 27 56

1. Go to WooCommerce > Settings > Accounts & Privacy in WP Admin
2. Disable checkout registration via the `During checkout` checkbox pictured above
3. As a reader, open modal checkout via any checkout or donate button
4. On `trunk` modal checkout will get stuck loading. On this branch it wont

### Other information:

### DIFF
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -171,6 +171,7 @@ final class Modal_Checkout {
 		add_filter( 'googlesitekit_adsense_tag_blocked', [ __CLASS__, 'is_modal_checkout' ] );
 		add_filter( 'jetpack_active_modules', [ __CLASS__, 'jetpack_active_modules' ] );
 		add_filter( 'woocommerce_checkout_update_order_review_expired', [ __CLASS__, 'is_not_modal_checkout_filter' ] );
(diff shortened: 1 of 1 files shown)


laurelfulford (approved):
This works as described!

I noticed one "off" thing, but I think it's more to do with having weird test site settings:

When the option is toggled off you lose the ability to change the account creation options. If we get feedback about this (readers being asked to create usernames and passwords on that first screen) we might want to toggle these both on, too, in cases where the "Allow customers to create an account" toggle is off and you're in the modal checkout: 

CleanShot 2024-12-18 at 13 20 16

What the checkout does with both of those disabled:

CleanShot 2024-12-18 at 13 22 04

Given that the original issue was much bigger -- broken checkout! -- we can table this as a future enhancement, or wait and see if we actually have people running into this issue 🙂

chickenn00dle:
Thanks @laurelfulford!

> Given that the original issue was much bigger -- broken checkout! -- we can table this as a future enhancement, or wait and see if we actually have people running into this issue 🙂

This works for me!

PR #2008, opened 2024-12-17
fix(modal-checkout): add custom modal checkout validation

chickenn00dle:
### All Submissions:

This PR moves the modal checkout validation request to newspack blocks. We still call `WC_Checkout::process_checkout`, but beforehand we override the needs payment flag so that step 1 of modal checkout doesn't validate payment method (not needed at this point).

1. Smoke test modal checkout using various supported payment methods, as a logged in and logged out reader:
 - Stripe
 - WCPay
 - PayPal Payments

### Other information:

### DIFF
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -118,6 +118,8 @@ final class Modal_Checkout {
 		add_action( 'wp', [ __CLASS__, 'process_checkout_request' ] );
 		add_action( 'wp_ajax_abandon_modal_checkout', [ __CLASS__, 'process_abandon_checkout' ] );
 		add_action( 'wp_ajax_nopriv_abandon_modal_checkout', [ __CLASS__, 'process_abandon_checkout' ] );
+		add_action( 'wp_ajax_validate_modal_checkout', [ __CLASS__, 'validate_checkout_request' ] );
+		add_action( 'wp_ajax_nopriv_validate_modal_checkout', [ __CLASS__, 'validate_checkout_request' ] );
 
 		add_filter( 'wp_redirect', [ __CLASS__, 'pass_url_param_on_redirect' ] );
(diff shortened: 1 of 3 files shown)


laurelfulford (changes requested):
Thanks for tackling this, @chickenn00dle! 

I worked through all the payment gateways you listed, and the only issue I ran into is PayPal Payments now shows a "Proceed to PayPal" button: 

CleanShot 2024-12-18 at 18 35 14

When I switch back to trunk, the button goes away. The Yellow button works as expected; the Proceed to PayPal button technically works, but it's kind of weird - it loads the form inside of the checkout modal, and then redirects to a full screen to authenticate, then into a non-modal order confirmation. 

Just let me know if this doesn't match what you're seeing and I'll see if I can recreate it elsewhere!

chickenn00dle:
Thanks for the review @laurelfulford!

> When I switch back to trunk, the button goes away. The Yellow button works as expected; the Proceed to PayPal button technically works, but it's kind of weird - it loads the form inside of the checkout modal, and then redirects to a full screen to authenticate, then into a non-modal order confirmation.

Good catch! 

I was checking `$wp_script->src` when iterating through `$wp_styles` so the paypal css assets were missing 🤦 .

Fixed in https://github.com/Automattic/newspack-blocks/pull/2008/commits/93392a5d372547044f4020f1605ca601fa5a9340

includes/class-modal-checkout.php
```diff
@@ -401,6 +399,26 @@ public static function process_abandon_checkout() {
 		wp_die();
 	}
 
+	/**
+	 * Process modal checkout validation.
+	 */
+	public static function validate_checkout_request() {
```
chickenn00dle, line 405: @miguelpeixe, I'd love your opinion on this approach for "hijacking" the form validation from Woo. In particular, do you think its safe to rename the `woocommerce-process-checkout-nonce` on the front end, then reassign in the backend before triggering `WC_Checkout::process_checkout`?

chickenn00dle:
Seems this is no longer needed, so gonna close this one!

PR #2032, opened 2025-01-24
fix: improve formatting of donation amount options

dkoo:
### All Submissions:

Fixes display of number amounts in the Donate block in editor and front-end contexts. Eventually we may want to refactor this block to fetch rendered HTML via PHP in the future, but that's beyond the scope of this fix.

1. Add a Donate block, set to configure manually + tiered and enter very large numbers in all amount inputs (ensuring that some numbers are integers and others have fractional amounts), like so:

Screenshot 2025-01-24 at 3 13 01 PM

2. Observe that in the editor, the amounts are shown without formatting, e.g. `500000` instead of `500,000` for both Frequency and Tiers layouts.

Screenshot 2025-01-24 at 3 15 37 PM
Screenshot 2025-01-24 at 3 16 19 PM

3. Observe that on the front-end, the amounts are shown either without formatting (Frequency layout) OR with formatting, but that integers are shown with trailing `00` decimals (Tiers layout):

Screenshot 2025-01-24 at 3 16 06 PM
Screenshot 2025-01-24 at 3 16 19 PM

4. Check out this branch, and confirm that the numbers are properly formatted and that integers do not include the trailing `00` decimals, in both front-end and editor contexts:

Screenshot 2025-01-24 at 3 15 53 PM
Screenshot 2025-01-24 at 3 16 36 PM

### Other information:

---
  - https://app.asana.com/0/0/1208923545061655

### DIFF
--- a/includes/class-newspack-blocks.php
+++ b/includes/class-newspack-blocks.php
@@ -1415,13 +1415,13 @@ class Newspack_Blocks {
 	/**
 	 * Get a formatted HTML string containing amount and frequency of a donation.
 	 *
-	 * @param float|string $amount Amount.
-	 * @param string       $frequency Frequency.
-	 * @param bool         $hide_once_label Whether to hide the "once" label.
+	 * @param float  $amount Amount.
+	 * @param string $frequency Frequency.
+	 * @param bool   $hide_once_label Whether to hide the "once" label.
 	 *
 	 * @return string
 	 */
-	public static function get_formatted_amount( $amount = 'AMOUNT_PLACEHOLDER', $frequency = 'day', $hide_once_label = false ) {
+	public static function get_formatted_amount( $amount = null, $frequency = null, $hide_once_label = false ) {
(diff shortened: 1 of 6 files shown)


leogermani (approved):
works as described.

Not sure if you wanted to address anything related with this, but if the number are really large, it will still break the layout, making the donate box wider than it should:

image

dkoo:
> if the number are really large, it will still break the layout, making the donate box wider than it should

@leogermani 557f06b should fix this. It still doesn't look great, but at least it doesn't break the layout:

Screenshot 2025-01-30 at 2 20 03 PM

faa8cfb also fixes an issue I was seeing with the regex that was causing the Tiers-based layout to not show the correct amounts in the editor.

leogermani (approved):
I think that's better.

I wonder if this PR solves the issue that publisher reported with large numbers or if we need more. In any case, it's worth merging this

PR #2038, opened 2025-02-03
fix(checkout-button): handle missing saved product

miguelpeixe:
### All Submissions:

1200550061930446-as-1209290415945917/f

When a Checkout Button block is saved with a non-existing product ID it stays in a broken state, unable to search for products.

This PR handles this case by resetting the block attribute and displaying an error message:

image

1. While on the release branch, add a checkout button block to a page and select a product
2. In the code editor, modify the ID to a random number (that doesn't match a product ID)
3. Back to the code editor, try to select a product, and confirm you get stuck with the spinner
4. Checkout this branch, repeat step 2 and confirm you get an error message as in the image above and you're able to select a new product and save
5. Refresh the page and confirm the selected product persists

### Other information:

---
  - https://app.asana.com/0/0/1209290415945917

### DIFF
--- a/src/blocks/checkout-button/edit.js
+++ b/src/blocks/checkout-button/edit.js
@@ -10,7 +10,7 @@ import { debounce, invert } from 'lodash';
 /**
  * WordPress dependencies
  */
-import { __ } from '@wordpress/i18n';
+import { __, sprintf } from '@wordpress/i18n';
 import {
 	InspectorControls,
 	RichText,
@@ -92,6 +92,7 @@ function ProductControl( props ) {
 	const [ suggestions, setSuggestions ] = useState( {} );
 	const [ selected, setSelected ] = useState( false );
 	const [ isChanging, setIsChanging ] = useState( false );
+	const [ productError, setProductError ] = useState( '' );
 
 	function fetchSuggestions( search ) {
 		setInFlight( true );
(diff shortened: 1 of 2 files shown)


src/blocks/checkout-button/edit.js
```diff
@@ -117,8 +118,19 @@ function ProductControl( props ) {
...
 				setSelected( product );
+				setProductError( '' );
 				props.onProduct( product );
 			} )
+			.catch( () => {
+				props.onChange( '' );
+				setProductError(
+					sprintf(
+						// translators: %s: product ID.
+						__( 'Product with ID %s was not found. Select a different product.', 'newspack-blocks' ),
```
dkoo:
```suggestion
						__( 'Could not find a product with ID %s. Please select a different product.', 'newspack-blocks' ),
```

dkoo (changes requested):
This is a good chance, but I'm not sure we should reset the product ID attribute in this scenario. Doing so makes it very easy to miss the error message since it will only appear one time, and on the front-end there's no difference between rendering a Checkout Button with an invalid product ID vs. no product ID at all (in both cases, the block doesn't render anything). What's the harm in keeping the invalid attribute so that the error message continues to display until you choose a new, valid product?

Aside from that I just have a non-blocking suggestion on the wording of the error message itself.

miguelpeixe:
Thanks, @dkoo! I agree we should keep the attribute until a new product is selected and tweaked the approach in 434f4f4bd0a167271353d2397b888f987d722d43.

I'm also proposing a small change in 756cdea962431f9b9d5151d6cb237e3d546f769d as well. There's more info in p1738607315332379/1738591216.983469-slack-C01C378QWRX, but the gist is that in the front-end we filter to match the string exactly, which is how the `FormTokenField` component works. In the backend it does a fuzzy search, so there can be an edge case in which the backend query does not include the desired result when using the exact string.

I'm proposing we enforce quotation marks (`"search string"`) for the backend, so it matches the front-end behavior.

dkoo (approved):
Works well! And I think the change to do an exact search for the search string is a good one.

PR #2049, opened 2025-02-08
feat: toggle transaction details

laurelfulford:
### All Submissions:

This PR makes the Transaction Details section of the Modal Checkout collapsed by default, and makes it toggle opened/closed.

I originally tried to approach this within the Newspack UI code (https://github.com/Automattic/newspack-plugin/pull/3734 and https://github.com/Automattic/newspack-plugin/pull/3734), but the way we destroy/re-add this table made it hard to get that approach playing nicely and not re-collapsing/requiring a bit of a rewrite.

This approach is strictly tied to the markup in the modal checkout, but let me know if the original approach would've been better!

See 1200550061930446-as-1208982074785832

1. Apply this PR and run `npm run build`.
2. Under Reader Revenue settings, enable the option to cover the transaction fees, but leave it unselected by default.
3. Run through a checkout using monthly or annual Donation product.
4. On the second screen, confirm you see the Transaction Details heading with a chevron:

CleanShot 2025-02-07 at 17 18 31

5. Inspect the heading (which should actually be a button) and confirm it has the attribute `aria-expanded` set to false, and the attribute `aria-controls` set to the ID that holds the transaction details table (`order_review_content`).
6. Click the header and confirm that the table expands nicely.
7. Inspect the heading and confirm the `aria-expanded` attribute has switched to `true`. 

CleanShot 2025-02-10 at 12 18 47

8. Try checking/unchecking the transaction fees checkbox, and confirm the table remains visible, and that the `aria-expanded` attribute on the heading stays as `true`.
9. Trying clicking the 'Back' then 'Continue' button to go to the first screen of the checkout and back again, and confirm the table remains visible, and that the `aria-expanded` attribute remain `true`.
10. Toggle the Transaction Details closed again, and repeat some of the above (toggling the transaction fees, clicking 'Back' and 'Continue) to make sure the `aria-expanded` attribute remains false. 
11. Run through a new checkout using the One-Time donation product.
12. On the second screen, confirm that the Transaction Details are not visible by default. 
13. Toggle on the option to cover transaction fees, and confirm the Transaction Details heading appears.
14. Click the Transaction Details header and confirm the table opens, and that it has the attribute `aria-expanded` set to `true`.
15. Try toggling off and on the transaction fee coverage and confirm the section disappears, but when it reappears the table is still open and that the heading's `aria-expanded` attribute is still `true`. 
16. Trying clicking the 'Back' then 'Continue' button to go to the first screen of the checkout and back again, and confirm the table remains visible, and that the `aria-expanded` attribute is still `true`. 
17. Toggle the Transaction Details closed again, and repeat some of the above (toggling the transaction fees, clicking 'Back' and 'Continue) to make sure the `aria-expanded` attribute remains false. 

### Other information:


laurelfulford:
This is nearly done, I just need to make it accessible then I'll mark it ready for review!

laurelfulford:
This should be ready to go! 

@thomasguillot I've included you for a design review just to make sure things look as expected (especially with how the animation feels!).

thomasguillot:
I've slightly tweaked the transitions. https://github.com/Automattic/newspack-blocks/pull/2049/commits/797c4580ad1e3a5953f5d8b7acbdff8da104e772

My main worry is how the scroll behaves. It's quite bouncy. I wonder if the user clicks this they are automatically scrolled down to the `#transaction-details`... Anyway it's not a blocker at all, just some I've noticed.

bouncin

dkoo:
I think the bounciness will be fixed by this PR: https://github.com/Automattic/newspack-blocks/pull/2052

There's currently a resize listener on the iframe element that attempts to scroll to to the top of the content whenever the iframe resizes. I think the intention is to scroll to the top of the page when transitioning between modal checkout screens, but it ends up getting triggered by anything that could cause a resize, including things like accordions.

laurelfulford:
Thanks @thomasguillot and @dkoo! 

https://github.com/Automattic/newspack-blocks/pull/2052 has been merged, so I've updated this PR from trunk and it's definitely helped with the weird jerkiness -- in my testing, the modal's scroll position is staying put as you toggle open and closed the transaction details, and toggle on and off covering fees.

chickenn00dle (approved):
Looks good. Great work @laurelfulford!

PR #2058, opened 2025-02-18
fix(carousel): avoid editor crash on empty

naxoc:
### All Submissions:

This fixes a bug where the Content Carousel block would "crash" in the editor and just show a spinner and not be responsive to changes made from the sidebar.

Before applying the patch, trigger the bug:

1. Create a Content Carousel on a post in the editor
2. In the "Content" panel, set a post category or tag - I just used "Science"
3. In the "Post Types" panel, select a different post type in addition to "Posts" that can't have any post categories or tags – I just used "Page"
4. Deselect "Posts" and observe editor crash and JS console error

Now apply the patch:
* Check that toggling between having post and page selected under post type will not crash the block 
* Check that the "Sorry, no posts were found." text is displayed when only page is selected.
* Check that there are no console errors
* Check that the block behaves as expected (shows the content you asked it to).

### Other information:

---
  - https://app.asana.com/0/0/1209344644455228

### DIFF
--- a/src/blocks/carousel/view.php
+++ b/src/blocks/carousel/view.php
@@ -34,7 +34,7 @@ function newspack_blocks_render_block_carousel( $attributes ) {
 
 	$article_query = new WP_Query( Newspack_Blocks::build_articles_query( $attributes, apply_filters( 'newspack_blocks_block_name', 'newspack-blocks/carousel' ) ) );
 	if ( false === $article_query->have_posts() ) {
-		return;
+		return '';
 	}
 
 	$counter = 0;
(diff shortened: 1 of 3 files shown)


src/blocks/carousel/view.php
```diff
@@ -34,7 +34,7 @@ function newspack_blocks_render_block_carousel( $attributes ) {
 
 	$article_query = new WP_Query( Newspack_Blocks::build_articles_query( $attributes, apply_filters( 'newspack_blocks_block_name', 'newspack-blocks/carousel' ) ) );
 	if ( false === $article_query->have_posts() ) {
-		return;
+		return '';
```
naxoc, line 37: The docs said it return string, so I changed that to ''.

dkoo (approved):
Works well and avoids the crash!

PR #2067, opened 2025-03-05
fix: remove coming soon mode message from the modal checkout

laurelfulford:
### All Submissions:

If you enable WooCommerce's new 'Coming Soon' mode (which is on by default for new sites), the modal checkout will display the 'Coming Soon' message underneath the checkout form.

This PR fixes the issue by disabling Coming Soon mode in the modal (so it displays as usual). Another option could be to disable the actual checkout and only display the message, but without the header/footer (which is what happens on the regular /checkout screens).

See 1200550061930446-as-1209426735338090

1. If it's not on already, turn on "Coming Soon" mode in Woo. You can see if it's on or off by checking the badge in the top-left corner of the admin toolbar, and clicking it to change the settings if it says "Live". You can also access this setting under WP Admin > Woo > Settings > Site visibility:

CleanShot 2025-03-05 at 09 53 00@2x

CleanShot 2025-03-05 at 09 55 03@2x

2. In an incognito window, go to the /shop page and confirm you get a full-page 'Coming Soon' message.
3. Try running through a checkout with the modal checkout and confirm you get the same message, under the form:

CleanShot 2025-03-05 at 09 56 36@2x

4. Apply this PR.
5. Repeat steps 2 and 3; confirm that you still get the 'Coming Soon' message on the /shop page, but not in the modal checkout. 

### Other information:

### DIFF
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -212,6 +212,9 @@ final class Modal_Checkout {
 
 		// Remove any hooks that aren't supported by the modal checkout.
 		add_action( 'wp_loaded', [ __CLASS__, 'remove_hooks' ] );
+
+		// Exclude the modal checkout from 'Coming Soon' mode.
+		add_action( 'plugins_loaded', [ __CLASS__, 'disable_coming_soon' ] );
 	}
 
 	/**
@@ -978,6 +981,22 @@ final class Modal_Checkout {
 			$priority = has_action( $remove['hook'], $remove['callback'] );
 			remove_action( $remove['hook'], $remove['callback'], $priority );
 		}
+	}
+
+	/**
+	 * Exclude the Modal Checkout from 'Coming Soon' mode.
+	 */
+	public static function disable_coming_soon() {
+		if ( ! self::is_modal_checkout() ) {
+			return;
+		}
+		add_filter(
+			'woocommerce_coming_soon_exclude',
+			function() {
+				return true;
+			},
+			10
+		);
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


includes/class-modal-checkout.php
```diff
@@ -980,6 +983,22 @@ public static function remove_hooks() {
...
 
+	/**
+	 * Exclude the Modal Checkout from 'Coming Soon' mode.
+	 */
+	public static function disable_coming_soon() {
+		if ( ! self::is_modal_checkout() ) {
+			return;
+		}
+		add_filter(
+			'woocommerce_coming_soon_exclude',
```
chickenn00dle: Is there a reason we don't just use this filter in `init` above?

```
add_filter( 'woocommerce_coming_soon_exclude', [ __CLASS__, 'disable_coming_soon' ] );
...
public static function disable_coming_soon() {
    return self::is_modal_checkout();
}
```

chickenn00dle (changes requested):
Works as expected, but I left some feedback.

Also noting this doesn't prevent the store notice from appearing when enabled via Appearance > Customize > WooCommerce > Store Notice

But that one is much less layout breaking so I'm fine if we don't want to remove that one.

includes/class-modal-checkout.php
laurelfulford: Gah, good call @chickenn00dle! 🤦‍♀️ I spent way too much time trying to get other hooks working related to the Coming Soon mode, I should've put more thought into this one. Updated in [092dd11](https://github.com/Automattic/newspack-blocks/pull/2067/commits/092dd11c52a8ffd59680766fd2c71788f4c44060)!

laurelfulford:
Thanks @chickenn00dle!

> Also noting this doesn't prevent the store notice from appearing when enabled via Appearance > Customize > WooCommerce > Store Notice

This seems to be a separate animal, and will appear whether or not 'Coming Soon' mode is enabled. We could probably make it look nicer, but I think it makes sense to tackle that separately 🙂

chickenn00dle (approved):
Looks good!

laurelfulford:
Thanks @chickenn00dle!

PR #2074, opened 2025-03-14
fix(donate-block): handle decimals in custom amount input

adekbadek:
### All Submissions:

Adds decimals handling to the custom value of the Donate block. 

1. On `trunk`, load a page with a Donate block and choose a custom amount. Input a value with decimals – e.g. `10.23`, and observe it won't pass browser validation
2. Switch to this branch, try again, observe that a value with decimals (using a `.`!) is accepted
3. In WooCommerce settings, change the value of the "Currency options" -> "Number of decimals" and observe the form is validated accordingly

### Other information:

---
  - https://app.asana.com/0/0/1209683142184116

### DIFF
--- a/src/blocks/donate/frontend/class-newspack-blocks-donate-renderer-frequency-based.php
+++ b/src/blocks/donate/frontend/class-newspack-blocks-donate-renderer-frequency-based.php
@@ -140,6 +140,10 @@ class Newspack_Blocks_Donate_Renderer_Frequency_Based extends Newspack_Blocks_Do
 
 		$configuration = self::get_configuration( $attributes );
 
+		// Calculate the value for the "step" property of the custom amount donation value `input` element.
+		$decimals = get_option( 'woocommerce_price_num_decimals', 2 );
+		$input_element_step = '0.' . str_repeat( '0', $decimals - 1 ) . '1';
+
 		ob_start();
 
 		/**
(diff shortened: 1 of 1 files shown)


naxoc:
I tested this with Danish currency options in Woo, and I got some wrong results on checkout. Danish uses `.` as a thousands separator and `,` as decimal separator:

Inputting `15.50`:
CleanShot 2025-03-17 at 13 13 19@2x
(It will not allow me to input a `,`)

Becomes:
CleanShot 2025-03-17 at 13 14 18@2x

(here are the settings in woo)
CleanShot 2025-03-17 at 13 14 59@2x

adekbadek:
That's an interesting condition, thanks for finding that! So if the thousand separator is a `.`, it will be conflated with the decimal separator (which for HTML is always a `.`). Fixed in [f06fb47](https://github.com/Automattic/newspack-blocks/pull/2074/commits/f06fb473958cc64d043c4c7fe2b50999f2eb863a) – if the separator is a `.`, the input field won't allow decimals because that's confusing.

In the future we might consider building a JS-powered custom input, which will handle all cases.

naxoc:
I'm not entirely sure I understand the `$input_element_step` variable, but the commit does make it so I can't input decimals with a comma. Most of Europe uses that, so we should probably support it, but this is good for now. 

I do wonder if we could use something like `wc_format_localized_decimal()`? Not sure if we know we have Woo available in the block though :)

adekbadek:
The problem is, according to the HTML spec, [a float can only use a dot](https://html.spec.whatwg.org/dev/common-microsyntaxes.html#valid-floating-point-number). Therefore, an `input[type="number"]` with a `,` decimal separator is invalid HTML. The solution would be to use a standard `input[type="text"]` element, with JS handling the validation and input filtering. Since this issue came up only once, the demand does not seem to justify the additional effort.

So, this PR will add handling for `.` as the decimal separator only.

naxoc (approved):
> The problem is, according to the HTML spec, [a float can only use a dot](https://html.spec.whatwg.org/dev/common-microsyntaxes.html#valid-floating-point-number).

Dang. TIL! 

Yeah, this PR makes sense. We can fix the other issue at some point later 👍

PR #2076, opened 2025-03-17
fix: better handle deleted terms

leogermani:
### All Submissions:

When a category or tag is deleted, the block might still be filtering by it, but nothing is displayed in its settings. If you want to clean up and remove that filter from the block, the only way to do it is to go through the source code.

Usually when a publisher deletes a category they don't realize it impacts the block. They see a block returning no posts and they can't tell why that's happening. They open the editor and the filter is clear. There's no easy way to fix the issue.

This PR adds a "deleted category/tag" placeholder to surface the issue and allow the editor to fix it.

Cleaning this up on the fly would be too expensive. And magically fixing this when loading the Editor could not work very well because the user would not have any visual feedback of what just happened and therefore not even click the save button.

1. Add a Content Loop block and filter by category
2. Delete the category (alternatively, you can just edit the source code and modify the filter to a category ID that doesn't exist)
3. refresh the editor and see that you see the "deleted category" placeholder
4. Confirm that you can delete and save the block and that you will see posts again.
5. Repeat it with tags

### Other information:

### DIFF
--- a/src/components/query-controls.js
+++ b/src/components/query-controls.js
@@ -125,49 +125,75 @@ class QueryControls extends Component {
 		);
 	};
 	fetchSavedCategories = categoryIDs => {
-		return apiFetch( {
-			path: addQueryArgs( '/wp/v2/categories', {
+		return apiFetch({
+			path: addQueryArgs('/wp/v2/categories', {
 				per_page: 100,
 				_fields: 'id,name',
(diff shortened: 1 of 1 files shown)


laurelfulford (approved):
Works as described! This is a nice clean way to fix this -- thanks @leogermani!

PR #2087, opened 2025-03-25
fix: correct checkout button block width in WP 6.8

laurelfulford:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-theme/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

There's a couple button block styling changes coming in WordPress 6.8 that negatively impact the Checkout Button block. This PR should fix that!

### How to test the changes in this Pull Request:

1. On a site running WP 6.7.2, add a few different checkout button blocks to a page, including:
    * One with a default width/alignment
    * One each with different width setting (100%, 75%, 50%, and 25%)
    * One each with different alignments (centred and wide). Overall you should get a page that looks like this:

CleanShot 2025-03-25 at 14 56 53

2. Switch the site to the latest RC of WP 6.8 (you can do this by installing the [Beta Tester plugin](https://wordpress.org/plugins/wordpress-beta-tester/), switching the settings to "Bleeding edge" and "Beta/RC Only", and upgrading your site under WP Admin > Dashboard > Updates).
3. Refresh your page - the button with the default settings and the button that's centred should now be full width:

CleanShot 2025-03-25 at 15 01 51

4. Apply this PR and run `npm run build`.
5. Confirm that the buttons now match the original screenshot from 6.7.2 on the front end and in the editor:

CleanShot 2025-03-25 at 15 04 17

6. Roll back your site to WP 6.7.2 (if you're using the Beta Tester plugin, you can do this on the WP Admin > Dashboard > Updates screen again, by clicking 'Re-install version 6.7.2'). 
7. Confirm the CSS changes don't actually change anything for WP 6.7.2 on the front-end and in the editor.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #2090, opened 2025-03-28
fix: correct Content Loop button width in WP 6.8

laurelfulford:
### All Submissions:

Similar to https://github.com/Automattic/newspack-blocks/pull/2087, changes to the default button styles in WordPress 6.8 have made the Content Loop 'Load More' button full-width.

This PR tweaks the styles so the appearance doesn't change.

See 1209694846565575-as-1209815966418365

1. On a test site running WP 6.8 RC, add a Content Loop block to the editor and toggle on the "Show Load more posts button" option.
2. View in the editor and front-end, and note it's full width:

CleanShot 2025-03-28 at 12 17 26

4. Retest in the editor on front-end and confirm the button is only as wide as it needs to be:

CleanShot 2025-03-28 at 12 10 24

5. Repeat the testing steps on a site running WP 6.7.2 to confirm it doesn't break anything.

### Other information:


naxoc (commented):
I tested this in FF and I got this really tall button:
CleanShot 2025-03-31 at 12 46 58@2x

Not sure if that would mess with other things, but also adding `height: auto;` made the button look normal for me.

laurelfulford:
That is wild @naxoc!! I use Firefox as my main browser, but I can't recreate this 😕 

Adding a `height: auto` shouldn't cause any issues, though! I added it to both the editor and front end styles -- just let me know if that doesn't do the trick for you!

naxoc:
Weird! I still get some wonkyness in FF. 

It looks like there is a pseudo-element in that block that makes the space the button is in too big:

CleanShot 2025-04-01 at 12 16 23@2x

laurelfulford:
> Weird! I still get some wonkyness in FF.

The good news (??) is that that's an unrelated issue -- there's a fix for that one here: https://github.com/Automattic/newspack-blocks/pull/2089 😅 It's the `.editor-style-wrapper` class that's the issue, it's on the main editor container as well and picking up some style changes there. This happened in really specific cases before 6.8, but something's changed that's made it more pronounced.

naxoc (approved):

PR #2101, opened 2025-04-08
feat: add body class depending on theme type

thomasguillot:
### All Submissions:

If we have a block theme active (e.g. our very own [newspack-block-theme](https://github.com/Automattic/newspack-block-theme)), this function adds an `is-block-theme` class to the body (editor and front-end).

The issue comes from the fact that some CSS was targeting the block theme using WP’s `theme-{stylesheet}` functionality. However, I’ve noticed that the `newspackstaging` environment is removing this class, and I’m not entirely sure why. In any case, this approach seems fragile, so we’re switching to using a more reliable classname to determine if it’s a block theme.

1. Switch to this branch
2. Switch your theme to a block theme
3. Check the editor and the front-end if you see the class applied
4. Add a colums block and check if the gap (32px) is ignored and instead uses the gap defined by the block
5. Switch to a classic theme and check, you shouldn't see any specific class applied
6. And check the columns block, the gap should be 32px

### Other information:

### DIFF
--- a/includes/class-newspack-blocks.php
+++ b/includes/class-newspack-blocks.php
@@ -85,16 +85,17 @@ class Newspack_Blocks {
 	/**
 	 * Body class.
 	 *
-	 * @param array $classes Array of body class names.
-	 * @return array Modified array of body class names.
+	 * @param string|array $classes Array or string of body class names.
+	 * @return string|array Modified array or string of body class names.
 	 */
 	public static function add_body_classes( $classes ) {
-		$block_theme = 'is-block-theme';
-
-		if ( wp_is_block_theme() && is_admin() ) {
-			$classes .= $block_theme;
-		} elseif ( wp_is_block_theme() ) {
-			$classes[] = $block_theme;
+		if ( wp_is_block_theme() ) {
+			// Handle string (admin) vs array (frontend) cases.
+			if ( is_string( $classes ) ) {
+				$classes .= ' is-block-theme ';
+			} else {
+				$classes[] = 'is-block-theme';
+			}
 		}
 
 		return $classes;
--- a/includes/class-newspack-blocks.php
+++ b/includes/class-newspack-blocks.php
@@ -29,6 +29,8 @@ class Newspack_Blocks {
 		add_post_type_support( 'page', 'newspack_blocks' );
 		add_action( 'jetpack_register_gutenberg_extensions', [ __CLASS__, 'disable_jetpack_donate' ], 99 );
 		add_filter( 'the_content', [ __CLASS__, 'hide_post_content_when_iframe_block_is_fullscreen' ] );
+		add_filter( 'body_class', [ __CLASS__, 'add_body_classes' ] );
+		add_filter( 'admin_body_class', [ __CLASS__, 'add_body_classes' ] );
 
 		/**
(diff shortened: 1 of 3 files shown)


includes/class-newspack-blocks.php
```diff
@@ -80,6 +82,24 @@ function( $classes ) {
...
+	 * @return array Modified array of body class names.
+	 */
+	public static function add_body_classes( $classes ) {
+		$block_theme = 'is-block-theme';
+
+		if ( wp_is_block_theme() && is_admin() ) {
+			$classes .= $block_theme;
+		} elseif ( wp_is_block_theme() ) {
+			$classes[] = $block_theme;
+		}
```
dkoo: Simplifying to this worked for me:

```suggestion
		if ( wp_is_block_theme() ) {
			$classes[] = 'is-block-theme';
		}
```

dkoo (approved):
Works! Just a small nitpick below.

includes/class-newspack-blocks.php
thomasguillot: Did it apply the `'is-block-theme` to the editor as well? In my test it didn't that's why I had to add `if ( wp_is_block_theme() && is_admin() ) {`

dkoo: It did in my test site:

Screenshot 2025-04-11 at 12 24 56 PM

thomasguillot:
@dkoo In https://github.com/Automattic/newspack-blocks/pull/2101/commits/a977c1b68f3c792220f8987a84e39dbe037a069d I've changed the logic a bit.

But if I just replace the whole thing with 
```php
if ( wp_is_block_theme() ) {
    $classes[] = 'is-block-theme';
}
```
Then I get an error in the editor in my local instance:
```
Fatal error: Uncaught Error: [] operator not supported for strings in /var/www/html/wp-content/plugins/newspack-blocks/includes/class-newspack-blocks.php:93 Stack trace: #0 /var/www/html/wp-includes/class-wp-hook.php(324): Newspack_Blocks::add_body_classes(' theme-newspack...') #1 /var/www/html/wp-includes/plugin.php(205): WP_Hook->apply_filters(' theme-newspack...', Array) #2 /var/www/html/wp-admin/admin-header.php(252): apply_filters('admin_body_clas...', '') #3 /var/www/html/wp-admin/edit-form-blocks.php(376): require_once('/var/www/html/w...') #4 /var/www/html/wp-admin/post.php(187): require('/var/www/html/w...') #5 {main} thrown in /var/www/html/wp-content/plugins/newspack-blocks/includes/class-newspack-blocks.php on line 93
```
That's because `admin_body_class` is [expecting](https://developer.wordpress.org/reference/hooks/admin_body_class/) a `string` and not an `array`. I don't understand how you were able to have the correct class applied to the editor on your test site.

includes/class-newspack-blocks.php
thomasguillot: [See](https://github.com/Automattic/newspack-blocks/pull/2101#issuecomment-2797931817)

dkoo (approved):
Gotcha, I'm not sure what's going on with the test site either, but it's still working with the latest changes, so let's go with that!

PR #2102, opened 2025-04-08
fix: ensure transaction total updates on Complete Transaction button for Braintree

laurelfulford:
### All Submissions:

This PR tweaks how the modal checkout button's text is updated: rather than adding and checking for the `button-label-updated` class, it relies on whether the `span.cart-price` element exists, which would also be inserted at the same time the `button-label-updated` CSS class was added.

This is to work around an issue with the Braintree payment gateway, which seems to cause issues with the `button-label-updated` class sticking.  

While testing this I found a separate issue where the Complete Transaction dollar amount is wiped out every time you change payment methods -- this happens with any payment gateway. I'll dig into that separately!

See 1200550061930446-as-1209884963637057

1. Set up a test site with the Braintree Credit Card gateway enabled (but not Braintree PayPal).
2. Enable coupons under WooCommerce > Settings, and add at least one coupon under WooCommerce > Marketing > Coupons.
3. Edit the includes/modal-checkout.php file to add `braintree_credit_card` to the `$supported_gateways` array, or make a mini-plugin to filter them: 

```
function add_gateway( $gateways ) {
	$gateways[] = 'braintree_credit_card';
	return $gateways;
}
add_filter( 'newspack_blocks_modal_checkout_supported_gateways', 'add_gateway' );
```
4. Set up a checkout button block.
5. In an incognito window, run through a checkout with your checkout button block. 
6. On the second screen, add your coupon code and click 'Apply'. Note that the 'Complete transaction: $X.XX' dollar amount on the place order button doesn't update.
7. Click 'Apply' again, and confirm the button text does update, even though you also get an error about the coupon already being applied. 
8. Apply this PR.
9. Repeat steps 5 - 7 and confirm that the coupon is applied to the Complete Transaction button when you first click 'Apply'. 
10. Switch entirely to a different payment gateway (like Stripe) and repeat steps 5-7, making sure that your coupon is applied. Note that I found I had to fully disable the Braintree plugin to turn it off, simply toggling it off under WooCommerce > Settings > Payments did not want to stick. 

### Other information:

### DIFF
--- a/src/modal-checkout/index.js
+++ b/src/modal-checkout/index.js
@@ -170,13 +170,16 @@ import { domReady } from './utils';
 					}
 				}
 
+				/**
+				 * Update Place Order button text.
+				 */
 				$( document ).on( 'updated_checkout', function() {
 					// Update "Place Order" button to include current price.
 					let processOrderText = newspackBlocksModalCheckout.labels.complete_button;
 					if ( ! processOrderText ) {
 						return;
 					}
-					if ( $( '#place_order' ).hasClass( 'button-label-updated' ) ) {
+					if ( $( '#place_order' ).has( $( 'span.cart-price' ) ) ) {
 						// Modify button text to include updated price.
 						const tree = $( '<div>' + processOrderText + '</div>' );
 						// Update the HTML in the .cart-price span with the new price, and return.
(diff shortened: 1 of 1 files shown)


laurelfulford:
Reworked as a hotfix here: https://github.com/Automattic/newspack-blocks/pull/2104

PR #2091, opened 2025-03-28
feat: display authors present in custom bylines

allysonsouza:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-theme/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Is adds a conditional check to `Newspack_Blocks::prepare_authors()` to check if custom bylines are in use, and then it returns an array of authors that are present on the custom byline online, not all authors from coauthors plus.

### How to test the changes in this Pull Request:

1. Look at a page where an article block is in use
2. Set a custom byline to one or more of the articles being displayed
3. Compare if the authors present in the custom byline are the ones that are being displayed in the block

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #2105, opened 2025-04-11
feat(content-loop): support custom bylines

miguelpeixe:
### All Submissions:

Implement custom byline support to the Content Loop block.

This PR also refactors how avatars and bylines are rendered in the editor for the Content Loop and Carousel blocks. It unifies the rendering strategy, passing the full markup instead of author data for react parsing. The single strategy improves extensibility and maintainability.

Supersedes #2091 

1. Checkout https://github.com/Automattic/newspack-plugin/pull/3911
2. Create a new category and publish 4 posts in this category with the following author setup:
   1. Single author with a custom byline
   3. 3 authors without custom byline
   4. 3 authors with a custom byline that only mentions 2 of them
   5. 3 authors with a custom byline **but toggled off**
6. Draft a new page and add a Content Loop block
7. Filter the block by the created category and confirm the bylines rendered as expected in the editor
8. Publish and visit the page
9. Confirm the bylines reflect the editor

### Other information:

### DIFF
--- a/includes/class-newspack-blocks-api.php
+++ b/includes/class-newspack-blocks-api.php
@@ -74,57 +74,6 @@ class Newspack_Blocks_API {
 	}
 
 	/**
-	 * Get author info for the rest field.
-	 *
-	 * @param array $object_info The object info.
-	 * @return array Author data.
-	 */
-	public static function newspack_blocks_get_author_info( $object_info ) {
-		$authors = [];
-		if ( class_exists( 'Newspack\Bylines' ) && Newspack\Bylines::is_enabled() ) {
-			$authors = Newspack\Bylines::get_post_byline_authors();
-		}
-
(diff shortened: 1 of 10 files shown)


leogermani (changes requested):
When using the custom byline, the post date sticks together with the last author name

image

In this branch, the Carousel block throws an error.

Question: Looking at the code I wonder if we should offer a filter, instead of calling the Bylines method directly... something like we did in the theme -> https://github.com/Automattic/newspack-theme/blob/trunk/newspack-theme/inc/template-tags.php#L92

miguelpeixe:
I got into a bit of a rabbit hole looking into implementing filters and decided to refactor how the editor renders the avatars and bylines.

Filters weren't practical before this refactor because the editor doesn't use the markup used in the view. It should now be unified and filterable, which is simpler and better to maintain.

leogermani (approved):

PR #2115, opened 2025-04-28
fix: the warning generated by get_term_classes because the category might be null

arthur791004:
### All Submissions:

Closes p1745618925471759-slack-CRWCHQGUB
Related to https://github.com/Automattic/jetpack/pull/43271

Fix the warning Warning: Attempt to read property "slug" on null generated by the get_term_classes function. You can see the error from logstash on Slack.

1. Weirdly, I cannot reproduce the issue even if I access the same request URL from the log...
2.
3.

### Other information:

### DIFF
--- a/includes/class-newspack-blocks.php
+++ b/includes/class-newspack-blocks.php
@@ -839,14 +839,18 @@ class Newspack_Blocks {
 		$tags = get_the_terms( $post_id, 'post_tag' );
 		if ( ! empty( $tags ) ) {
 			foreach ( $tags as $tag ) {
-				$classes[] = 'tag-' . $tag->slug;
+				if ( ! empty( $tag->slug ) ) {
+					$classes[] = 'tag-' . $tag->slug;
+				}
 			}
 		}
 
 		$categories = get_the_terms( $post_id, 'category' );
 		if ( ! empty( $categories ) ) {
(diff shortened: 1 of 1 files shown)


includes/class-newspack-blocks.php
```diff
@@ -839,22 +839,28 @@ public static function get_term_classes( $post_id ) {
...
+				}
 			}
 		}
 
 		foreach ( self::get_custom_taxonomies() as $tax ) {
 			$terms = get_the_terms( $post_id, $tax['slug'] );
 			if ( ! empty( $terms ) ) {
 				foreach ( $terms as $term ) {
-					$classes[] = $term->taxonomy . '-' . $term->slug;
+					if ( $term && isset( $term->taxonomy ) && isset( $term->slug ) ) {
```
fgiannar:
```suggestion
					if ( isset( $term->taxonomy, $term->slug ) ) {
```

Not a blocker, just an FYI that this check would suffice without having to check `$term`. Also, just in case it makes sense, using a `! empty` check here might also make sense to not account empty slugs.

arthur791004: Cool. I didn't know it's enough 👍

leogermani (approved):
Weird that we would ever have empty values there, but the check looks ok and harmless

PR #2117, opened 2025-05-06
feat: update styles for the Subscription Confirmation checkbox

laurelfulford:
### All Submissions:

This PR should be tested with https://github.com/Automattic/newspack-plugin/pull/3961. It adds three things: 

* It adds some styles for spacing for the Subscription Confirmation checkboxes in the modal checkout. 
* It adds a new `<span>` element when a modal checkout label is for a required field and for a checkbox form element. This is to help it play nicer with our Newspack UI styles which use grid; without it, separate HTML elements -- like required asterisk and link -- ended up following the grid.
* It adds a check to block the modal checkout from being submitted when the Subscription Confirmation or Terms & Conditions checkboxes are present but not checked. There's also regular Woo validation, but this makes the requirement a little more up front. 

See 1200550061930446-as-1208929637924603

1. Please follow the testing steps in https://github.com/Automattic/newspack-plugin/pull/3961.

### Other information:

### DIFF
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -193,6 +193,9 @@ final class Modal_Checkout {
 		add_action( 'wp_ajax_get_cart_total', [ __CLASS__, 'get_cart_total_js' ] );
 		add_action( 'wp_ajax_nopriv_get_cart_total', [ __CLASS__, 'get_cart_total_js' ] );
 
+		// Wrap required checkbox text in a span so it works nicely with the Newspack UI grid layout.
+		add_filter( 'woocommerce_form_field_checkbox', [ __CLASS__, 'wrap_required_checkbox_text' ], 10, 4 );
+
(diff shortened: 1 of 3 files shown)


adekbadek (commented):
> The "Place Order" button should also be greyed out while this checkbox isn't selected

Didn't happen for me:

image

I see `handleSubscriptionConfirmation` function executed on the first step of the checkout modal (name & email fields), but then not on the actual checkout page.

laurelfulford:
Thanks @adekbadek! I think I was only testing after I filled in my email 🤦‍♀️ this should be fixed in [456de11](https://github.com/Automattic/newspack-blocks/pull/2117/commits/456de11d2e80116155a6dce3a95594047145063e).

claudiulodro (approved):

adekbadek (approved):
Works as expected now!

PR #2121, opened 2025-05-13
feat(modal-checkout): checkout from "My Account" dashboard

miguelpeixe:
### All Submissions:

Introduces the `newspackOpenModalCheckout( options );` function so the modal checkout can be initiated programmatically by 3rd parties.

`options.title` will determine the checkout title, not overriding other modal titles the modal checkout may require (thank you/newsletters).

`options.actionType` will set a custom `action_type` for the `checkout_completed` activity, which currently is either `checkout_button` or `donation`.

`options.afterSuccess` sets the after success `url`, `behavior`, and `buttonLabel` for the checkout.

`options.onCheckoutComplete` is a callback to be called on the `checkout_complete` RAS activity with the activity `data`

`options.onClose` is a callback to be called on the newly introduced `checkout-closed` custom event

This required a few changes to the modal checkout structure, which I commented inline.

Follow instructions at https://github.com/Automattic/newspack-plugin/pull/3973, but also make sure this introduces no regressions to regular modal checkout flows by donating and purchasing a subscription as a reader.

### Other information:


includes/class-modal-checkout.php
```diff
@@ -1111,29 +1112,23 @@ public static function get_checkout_template( $template ) {
...
-			);
 		}
+		return array_filter(
+			[
+				'after_success_behavior'     => isset( $request_params['after_success_behavior'] ) ? sanitize_text_field( wp_unslash( $request_params['after_success_behavior'] ) ) : '', // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+				'after_success_url'          => isset( $request_params['after_success_url'] ) ? sanitize_url( wp_unslash( $request_params['after_success_url'] ) ) : '', // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+				'after_success_button_label' => isset( $request_params['after_success_button_label'] ) ? sanitize_text_field( wp_unslash( $request_params['after_success_button_label'] ) ) : '', // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+				'action_type'                => isset( $request_params['action_type'] ) ? sanitize_text_field( wp_unslash( $request_params['action_type'] ) ) : '', // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+			]
+		);
```
miguelpeixe, line 1106: This was slightly refactored to avoid code repetition

includes/tracking/class-data-events.php
```diff
@@ -163,16 +163,31 @@ public static function build_js_data_events( $product_id, $cart_item = null, $or
...
+				$data_order_details['subscription_ids'] = array_values(
+					array_map(
+						function( $subscription ) {
+							return $subscription->get_id();
+						},
+						$subscriptions
+					)
+				);
+			}
+		}
```
miguelpeixe: Adding the `subscription_ids` to the order details so it can get picked up by modal checkout callers via the `checkout_completed` activity.

src/modal-checkout/utils.js
```diff
@@ -40,3 +40,97 @@ export function createHiddenInput( name, value = null ) {
...
+}
+
+/**
+ * Get details from the data-order-details attribute given an element Node or ID.
+ *
+ * @param {Node|string} element HTML element or its ID to get order details from.
+ *
+ * @return {Object} Order details.
+ */
+export function getOrderDetails( element ) {
```
miguelpeixe: Moved from the analytics umbrella because it serves more than analytics. Anything can pick up the RAS activity it feeds

src/modal-checkout/utils.js
```diff
@@ -40,3 +40,97 @@ export function createHiddenInput( name, value = null ) {
...
+}
+
+/**
+ * Get details from the data-order-details attribute given an element Node or ID.
+ *
+ * @param {HTMLElement|string} element HTML element or its ID to get order details from.
+ *
+ * @return {Object} Order details.
+ */
+export function getOrderDetails( element ) {
```
dkoo: As noted in https://github.com/Automattic/newspack-plugin/pull/3973, the payload for data events related to checkout seems to have changed, especially when purchasing variable products. Here's a comparison of data events from buying the same variable subscription product via the same Checkout Button, on `trunk` vs. this branch:

```js

[
    // trunk
    {
        "action": "checkout_completed",
        "data": {
            "action_type": "checkout_button",
            "currency": "USD",
            "is_variable": true,
            "product_id": 662,
            "product_type": "membership",
            "recurrence": "month",
            "referrer": "/support/",
            "price": "25",
            "product_price_summary": "All Access - Monthly: $25.00 / month",
            "variation_id": "663"
        },
        "timestamp": 1747942779181
    },
    // feat/modal-checkout-my-account
    {
        "action": "checkout_completed",
        "data": {
            "amount": "25.00",
            "action_type": "checkout_button",
            "currency": "USD",
            "product_id": "662",
            "product_type": "membership",
            "referrer": "/support/",
            "recurrence": "month",
            "variation_id": "",
            "order_id": 6775
        },
        "timestamp": 1747942692518
    }
]
```

dkoo (changes requested):
Very nice refactor! I really like the addition of a global `newspackOpenModalCheckout` function that allows other plugin code to trigger modal checkouts. I noticed some differences in the data events activity payload for checkouts that we may want to examine more closely to avoid discontinuity in analytics, but other than that this is testing well.

src/modal-checkout/utils.js
miguelpeixe: Hmmm I think that's because I [changed the node](https://github.com/Automattic/newspack-blocks/blob/e711857f99fb0212f56b3c581a97243cbf8e05c0/src/modal-checkout/modal.js#L107) it picks up from. In trunk, it's the modal checkout container div, in here it's from the thank you paragraph. I did that because this other one includes order details (not only product details).

I'll work on unifying that data. Maybe look for a way to not have it printed in 2 different places and just update the same source on checkout completion.

dkoo: Yeah, I remember I ran into some issues with fetching some data from the thank you screen. We might need to extend our custom template a bit further to receive more data from the checkout screens.

miguelpeixe: A fetch/ajax request may be better suited. It doesn't have to be UI-blocking.

miguelpeixe: Also just realized that part of that payload is not just from `build_js_data_events()`, it's also from here: 

https://github.com/Automattic/newspack-blocks/blob/43c331555f4ecac69ff19ea85800b89c8c9e4a1e/includes/class-modal-checkout.php#L736-L746

We should unify that...

miguelpeixe: I consolidated this data in a13264ae6c70798f57e2d0490c3601cfd7355976.

The method was already covering product, cart, and order contexts, so I renamed it to `get_checkout_data()` and refactored it. This allows for the same method to serve the variation modal so we reduce code duplication.

This same data is also generated [here](https://github.com/Automattic/newspack-blocks/blob/1462233821b4f41ef6b80f6929243dc1b805cdb1/src/blocks/checkout-button/view.php#L148-L168) and [here](https://github.com/Automattic/newspack-blocks/blob/2eeaa8933cd4ac877389362aa0815e36d50bc729/src/blocks/donate/frontend/class-newspack-blocks-donate-renderer-tiers-based.php#L67-L71), which is not great.

I'm working on consolidating that as well. Moving the `get_checkout_data()` outside the "tracking" namespace should be better, too.

I'm not pursuing a different fetching strategy. Having a single source of truth for the data should be fine for now.

miguelpeixe:
It took a larger changeset than I anticipated, but it improves our management of the data that flows through the checkout modal.

While working on it, I caught a few bugs that have also been fixed here and in https://github.com/Automattic/newspack-plugin/pull/3973:

 - Opening the modal checkout or pre-checkout auth modal always attaches a new ESC event listener that's never removed. This end up triggering multiple `closeCheckout()` calls and consequently repeated GA events if the modal is opened more than once in the page (fixed in  de679f52addcb32ab12a8876e6abc22c3809646d and https://github.com/Automattic/newspack-plugin/pull/3988/commits/fdc1d638f0791229157df9a226a71181350a8feb)
 - The custom product price summary was being implemented only to the auth modal, not the checkout itself:

| Auth | Checkout | 
| --- | --- |
| image | image | 

This was fixed by consolidating how the price summary is handled for all product types, and applying it:

https://github.com/Automattic/newspack-blocks/blob/14498e332dbffa8991feaae376e2cbc8e442e2a5/src/modal-checkout/modal.js#L121-L122

The modal checkout data should now be consolidated into a single `checkoutData`. This object is formed with custom JSON set to the `data-checkout` attribute and form inputs. There's a noticeable simplification of the `handleCheckoutSubmit()` function, which now carries fewer conditional logic for specific states and should rely more on the `checkoutData` as source of truth.

Because this inevitably touches all checkout flows, they should be tested, including GA4 events:

 - With "Require sign-in or create account before checkout" (Audience -> Configuration -> Checkout & Payment) and as an anonymous reader:
   - Purchase a subscription
   - Make a single donation using tiered block
   - Make a single donation using untiered block
   - Make a recurring donation using tiered block
   - Make a recurring donation using untiered block
- Repeat the above while logged in as a reader
- Visit My Account dashboard and perform a subscription early renew via modal checkout

src/modal-checkout/modal.js
```diff
@@ -126,14 +100,12 @@ domReady( () => {
 			}
 			iframe._ready = true;
 		}
+		const productDetails = container.querySelector( '#modal-checkout-product-details' );
```
dkoo, line 106: This is being called immediately when I open any modal checkout, but before the contents have loaded, resulting in some JS errors:

```
Uncaught TypeError: Cannot read properties of null (reading 'querySelector')
    at HTMLIFrameElement.handleIframeReady (modal.js?ver=4.12.0:754:38)
    at openCheckout (modal.js?ver=4.12.0:1156:18)
    at HTMLFormElement.handleCheckoutFormSubmit (modal.js?ver=4.12.0:1029:7)

[...]

Uncaught TypeError: Cannot read properties of null (reading 'querySelector')
    at HTMLIFrameElement.handleIframeReady (modal.js:122:20)
```

Once the modal contents load I see this code running again, when it won't result in any errors. I think we can avoid the errors if we bail early if either `container` or `productDetails` can't be found.

includes/modal-checkout/class-checkout-data.php
```diff
@@ -0,0 +1,277 @@
...
+		}
+
+		$data = [
+			'amount'           => $amount,
+			'action_type'      => self::get_action_type( $product_id ),
+			'currency'         => function_exists( 'get_woocommerce_currency' ) ? \get_woocommerce_currency() : 'USD',
+			'product_id'       => strval( $product_id ? $product_id : '' ),
+			'product_type'     => $product_type,
+			'price_summary'    => self::get_price_summary( $name, $amount, $recurrence ),
+			'summary_template' => self::get_price_summary( $name, '{{PRICE}}', $recurrence ),
```
dkoo, line 231: This is a neat way to build the price summary, but it's superfluous to send to GA or any other integrations using this data. Can we strip it once `price_summary` has been set in the data object?

includes/modal-checkout/class-checkout-data.php
```diff
@@ -0,0 +1,277 @@
...
+			'referrer'         => $referrer ? str_replace( home_url(), '', $referrer ) : '', // Keeps format consistent for Homepage with Donate and Checkout Button blocks.
+			'recurrence'       => $recurrence,
+			'variation_id'     => strval( $variation_id ? $variation_id : '' ),
```
dkoo, line 234: Can we strip `variation_id` or any other keys with empty values before sending to GA4?

dkoo (changes requested):
@miguelpeixe thank you for these revisions and for starting the standardization of checkout data! In general I like the consolidation of the transaction details into the new `Checkout_Data` class. I think this will be much more readable and easier to follow in the future compared with how it was handled before.

I'm running into a handful of issues that I think are related to these changes. No big surprise since this is an unexpectedly big refactor of a feature that touches many different areas.

1. When doing checkouts from a paywall/content gate, the `np_modal_checkout_interaction` events don't always seem to have `gate_post_id` params attached. I think this might be due to changes in how URL params are handled/passed from screen to screen within the modal iframe? In particular, the `action=open` event did have the param, but `action=loaded` and `action=continue` when transitioning between checkout screens did not.  I tested against `trunk` and there, all `np_modal_checkout_interaction` events did have a `gate_post_id` param.

2. [NON-BLOCKING, but mentioning here because it's related to the next item] This is not related to these changes, but noting in case we want to handle this edge case in the future. Variable, non-subscription products are not handled by our Checkout Button block. Attaching a variable non-subscription product to a Checkout Button and clicking the button as a reader will always result in a checkout error: `Please choose product options by visiting <a href="[product page URL] title="Product Name">Product Name</a>.`

3. Related to the above, on `trunk` if as a reader I click a Checkout Button with a variable non-subscription product, I'll see the error message in the modal checkout window with a "Go back" button to close. On this branch, the modal checkout window will get stuck in a loading state and the error is logged to Logstash, but never shown to the reader. This is also replicable if you set a product to be limited to one per customer. On `trunk`, I see the error message below if I try to purchase a subscription product that I already own, but on this branch the modal checkout window hangs in a loading state.

Screenshot 2025-05-27 at 4 06 51 PM
Screenshot 2025-05-27 at 3 46 01 PM

Other feedback inline.

src/modal-checkout/modal.js
miguelpeixe: Whoops! Fixed in a435ca47ab5ae4d73ea39379f5901ccc31bdc958

includes/modal-checkout/class-checkout-data.php
miguelpeixe: We need to keep the `price_summary` because it'll be reused every time a modal is rendered on the same page.

How about controlling the keys that are sent instead? Implemented in e74c77cce69def34c4e4acbf5afbf445b8f5e96f

includes/modal-checkout/class-checkout-data.php
miguelpeixe: Implemented as part of the proposal to control the sent keys: e74c77cce69def34c4e4acbf5afbf445b8f5e96f

miguelpeixe:
The `gate_post_id` param has been fixed in 5b08f3d6151a4adae53a889e353c8806176440dc. Because we're now fetching that information from the cart item data, we no longer need to worry about passing it via URL, so this has been removed in d734a65e17b663ad76317930ae000b62c6a12e25.

I also identified an issue with `newspack_popup_id` that was also present in `trunk`. A hotfix is proposed in https://github.com/Automattic/newspack-blocks/pull/2134 and https://github.com/Automattic/newspack-popups/pull/1441. The fix is also implemented here (0a00a46a2e5c3994706b0601ea6009a259ca951d) so it should work in this branch as well.

src/modal-checkout/modal.js
miguelpeixe: The issue with error handling failing has also been fixed by a435ca47ab5ae4d73ea39379f5901ccc31bdc958

includes/modal-checkout/class-checkout-data.php
dkoo: I like this approach! Makes it much easier to see and control which keys ultimately get sent where

src/modal-checkout/analytics/ga4/utils/index.js
```diff
@@ -12,29 +12,42 @@ export const getEventPayload = ( action, extraParams = {} ) => {
 	return { ...extraParams, action };
 };
 
+/**
+ * Checkout data keys that can be included in the event payload.
+ *
+ * @type {string[]}
+ */
+const eventKeys = [
```
dkoo, line 20: We also need `action` for the action parameter, e.g. `seen`, `loaded`, `continue`, etc. as well as all [custom event params that we add in the main plugin](https://github.com/Automattic/newspack-plugin/blob/feat/my-account-modal-checkout-resubscribe/includes/plugins/google-site-kit/class-googlesitekit.php#L196).

dkoo (changes requested):
This is all working well now! We're just missing a handful of GA event params in our whitelist—details below. After that this should be good to merge.

src/modal-checkout/analytics/ga4/utils/index.js
miguelpeixe: > as well as all [custom event params that we add in the main plugin](https://github.com/Automattic/newspack-plugin/blob/feat/my-account-modal-checkout-resubscribe/includes/plugins/google-site-kit/class-googlesitekit.php#L196).

Were these part of the payload before? I've never seen them here

miguelpeixe: Added `action` in 83727183c10cc9c468ded9cda183e4a634a40a42

dkoo: > Were these part of the payload before? I've never seen them here

They're only enabled if you add a `NEWSPACK_GA_ENABLE_CUSTOM_FE_PARAMS` environment constant. Originally we sent them with every event, but because some publishers have custom dimensions and events in their own GA setups, they sometimes caused our GA events to exceed the parameter limit per event. They're now opt-in only via this constant.

miguelpeixe: Ah, gotcha! We don't need to worry about it, then. That's hooked into `googlesitekit_gtag_opt`, which adds the custom parameters some other way. We don't manually send these via `window.gtag( 'event' )`.

Seems to work fine for me after enabling the flag.

dkoo: Aha, you're right! I thought I wasn't seeing them, but it's just because I was testing on a page without categories. Thanks for looking into this.

dkoo (approved):
Looking good, thanks for the extensive refactoring here!

PR #2123, opened 2025-05-21
fix(homepage-articles): add style to head

adekbadek:
### All Submissions:

The HPB aka Content Loop block was rendering it's styles inline, in a `<style>` tag. That did not jive with how the Tabs block (or maybe some other blocks, too) render content nested in inner blocks. To be on the safe side, this PR moves the styles rendering to the document head.

1. Insert the Content Loop block on a page
2. Ensure a post with multiple authors is included – the styles for multiple authors are among those concerned here
3. Observe the post with multiple authors it renders with expected

### Other information:

---
  - https://app.asana.com/0/0/1210279806949908

### DIFF
--- a/src/blocks/homepage-articles/view.php
+++ b/src/blocks/homepage-articles/view.php
@@ -232,22 +232,40 @@ function newspack_blocks_render_block_homepage_articles( $attributes ) {
 		return;
 	}
 
+	// If the block is rendered in wp_kses_post() call, ensure the `time` tag is allowed.
+	\add_filter(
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Pre-approving because I don't see any issues on the front-end (which is presumably the more important context for this fix), but the styles in the editor still look wonky to me:

Screenshot 2025-05-22 at 4 10 26 PM

adekbadek:
This would not be related to the changes in this PR. Can you reproduce on `trunk`? If so, this is a new issue.

PR #2135, opened 2025-05-29
fix: update supported gateway check to improve accuracy

laurelfulford:
### All Submissions:

Our supported payment gateway check is sometimes being run too early; it defaults to `false` (no unsupported gateways) so depending when it was run when there are unsupported gateways, it would return a mix of incorrect `false` values and correct `true` values.

This update tries to run the check a little earlier and save the value, to improve the check's accuracy. It also changes the default to `true` so if the payment gateway check can't be run in time for whatever reason, it will default to unsupported gateways being set. 

Note: please test each checking step (where you test the modal) in a couple incognito windows, just to try to spook out whether it's just intermittently working 😅

1. Set up a few products, like a simple product and a variable product.
2. Using those products, set up at least three checkout button blocks: one with the simple product, one with the variable product with one variable pre-selected, and one with the variable product that lets the reader pick.
3. Set up Stripe, and under its settings, enable at least one unsupported third party payment option, like Klarna, Alipay, etc.
4. In an incognito window, run through the modal checkout. It may take a few failures, but you should eventually hit an attempt where the modal checkout opens in the browser window, taking it over, not in the modal checkout. 
5. Apply this PR.
6. Repeat step 4; you should consistently get the spinner animation on the button for the simple product and variable production with the variation picked, then bringing you to the regular /checkout page. For the variable product with no variation selected, you should get the variation picker modal, then the spinner animation on the button to bring you to the /checkout page. 
7. Go to WooCommerce > Settings > Payments > Stripe, and uncheck the unsupported gateways. 
8. Repeat the testing steps, and confirm that each option opens the modal checkout.

[Of the original testing steps](https://github.com/Automattic/newspack-blocks/pull/2009), please also check:

1. If the Donate block (both regular and tiered) correctly animates the buttons and sends them to the /checkout when a unsupported payment gateway is enabled. 
2. If the Donate block opens the modal when no unsupported gateways are enabled. 
3. When registration is required to make a purchase and you have an unsupported payment gateway, it still opens the sign in modal, then brings you to /checkout.
4. When registration is required to make a purchase and you don't have an unsupported payment gateway enabled, it correctly opens the sign in modal, then moves to the modal checkout. 

### Other information:


leogermani (approved):
Looks good!

laurelfulford:
Thanks @leogermani! I totally forgot about this one so it's probably less urgent than the hotfix suggests (we fixed it on the one affected site by disabling the unsupported gateway), but I'll merge it as one anyway!

PR #2139, opened 2025-06-03
fix(homepage-articles): revert #2123

miguelpeixe:
### All Submissions:

p1748973619796749-slack-C015W6BES8J

Reverts #2123.

The solution doesn't interact well with block caching. When blocks are cached, the block render callback is never called, and the inline styles are never enqueued.

1. While on release, visit your site anonymously and confirm that when blocks are cached, the `newspack-blocks-homepage-articles-inline` style tag is not added to the markup
2. Checkout this branch, refresh and confirm the inline styles are rendered as expected (as part of the block markup)

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/view.php
+++ b/src/blocks/homepage-articles/view.php
@@ -232,40 +232,22 @@ function newspack_blocks_render_block_homepage_articles( $attributes ) {
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


laurelfulford (approved):
LGTM! :shipit:

PR #2145, opened 2025-06-10
fix(homepage-articles): add styles outside of the block markup

adekbadek:
### All Submissions:

The HPB aka Content Loop block was rendering it's styles inline, in a `<style>` tag. That did not jive with how the Tabs block (or maybe some other blocks, too) render content nested in inner blocks. To be on the safe side, this PR moves the styles rendering to the document head.

This is a re-implementation of #2123 after it had to be reverted in #2139. The issue with the original approach was that the enqueueing of the CSS was done only if the block was not cached. This resulted in an experience in which the first request would get the the CSS enqueued, but the subsequent requests would not, because the HTML was served from cache and thus the `newspack_blocks_render_block_homepage_articles` function (enqueueing the CSS) was bypassed. 

1. Insert the Content Loop block on a page
2. Ensure a post with multiple authors is included – the styles for multiple authors are among those concerned here
3. Set the type scale of the block to the maximum value
4. Observe the post with multiple authors it rendered as expected
5. `wp config set NEWSPACK_LOG_LEVEL 2`, so that HPB caching is logged
6. Reload the page a few times and observe that the block is served from cache, but the styles persist

### Other information:

---
  - https://app.asana.com/0/0/1210279806949908

### DIFF
--- a/includes/class-newspack-blocks-caching.php
+++ b/includes/class-newspack-blocks-caching.php
@@ -212,6 +212,10 @@ class Newspack_Blocks_Caching {
 		if ( ! self::should_cache_block( $block_data ) ) {
 			return $block_html;
 		}
+
+		// Equeue the styles needed to render the blocks.
+		newspack_blocks_enqueue_block_homepage_articles_styles();
+
 		if ( ! self::$can_serve_all_blocks_from_cache ) {
 			return $block_html;
 		}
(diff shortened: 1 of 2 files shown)


laurelfulford (approved):
Thanks for updating this one, @adekbadek -- it looks good!

PR #2162, opened 2025-07-07
fix(modal-checkout): return array when getting event handlers

miguelpeixe:
### All Submissions:

When getting the event handlers to be used here:

https://github.com/Automattic/newspack-blocks/blob/b82874360a20938cc199eec2c2a348bbc8df27db/src/modal-checkout/index.js#L469-L473

The function can return `undefined` if the element has events but not the event being requested. It's a rare edge case only seen in a race condition, where a `click` event handler is registered, but no `submit` event handler has been registered yet.

This crashes the code above with `Uncaught TypeError: Cannot read properties of undefined (reading 'slice')`.

1. While on the release branch, tweak lines 219-225 of `src/modal-checkout/index.js` to run behind `setTimeout`:

```js
setTimeout(() => {
  setEditingDetails( true );
  if ( ! $gift_options.length ) {
    // Perform initial validation so it can skip 1st step if possible.
    validateForm( true, setReady );
  } else {
    setReady();
  }
}, 10);
```

2. Start the modal checkout as an anonymous user (register a new reader in the process if your site requires you to)
3. If the checkout doesn't crash, try again or tweak the timeout delay until you hit it. Mine was 10, but also intermittent.
4. Checkout this branch, reapply the setTimeout, and confirm that there's no crash when there are event handlers but none in the `submit` event
5. Complete the checkout flow and confirm it behaves as expected

### Other information:

### DIFF
--- a/src/modal-checkout/index.js
+++ b/src/modal-checkout/index.js
@@ -29,7 +29,7 @@ import { domReady } from './utils';
 			if ( ! event ) {
 				return events;
 			}
-			return $._data( element, 'events' )[ event ];
+			return events[ event ] || [];
 		}
 
 		function clearNotices() {


claudiulodro (approved):
I wasn't able to replicate the race condition failure locally, so not sure if you want a second opinion, but I can confirm that Checkout works fine with this PR and the code change doesn't seem to be doing anything unreasonable. :)

PR #2166, opened 2025-07-16
Fix/author list roles and exclusion

leogermani:
### All Submissions:

Fixes two issues in the Author List plugin

* It doesn't properly filter custom roles
* Exclude authors (cherry picking author names) is not working

1. You can test this with any custom role, like the ones added by Yoast or Woocommerce
2. Make sure you have authors in one of those custom roles
3. Test https://github.com/Automattic/newspack-plugin/pull/4090
4. Add an Author List block and filter by that role
5. Confirm it works as expected
6. Test excluding one of the visible authors
7. Confirm they are correctly excludeed

### Other information:

### DIFF
--- a/includes/class-newspack-blocks.php
+++ b/includes/class-newspack-blocks.php
@@ -240,7 +240,7 @@ class Newspack_Blocks {
 			if ( class_exists( 'WP_REST_Newspack_Author_List_Controller' ) ) {
 				$localized_data['can_use_cap']    = class_exists( 'CoAuthors_Guest_Authors' );
 				$author_list_controller           = new WP_REST_Newspack_Author_List_Controller();
-				$localized_data['editable_roles'] = $author_list_controller->get_editable_roles();
+				$localized_data['editable_roles'] = array_values( $author_list_controller->get_editable_roles() );
 			}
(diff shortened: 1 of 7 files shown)


rbcorrales:
I tested this with the custom "Shop manager" role and after checking out the branch, it's showing the user after selecting that role.

However, I'm not sure if this needs additional changes, because if I try to exclude that specific user with the custom role searching by name, it never gets listed (the `/wp-json/newspack-blocks/v1/authors?search=customer&offset=0&fields=id%2Cname&_locale=user` XHR call returns an empty array):

image

leogermani:
Thanks @rbcorrales , you found another bug.

It led me to a small refactor, but I think things look better now. Please test again (need to update the other plugin branch as well).

And can I ask you to test something else? It's important that we ensure that previous instances of the block are not affected. So create a few blocks in trunk and then apply this branch.

In my testing, some weird things happen when you try to edit the block... it will get in a broken state. BUT it will work as usual if you don't touch them, which I think is good enough

rbcorrales (approved):
The latest version works. Thanks for addressing this!

PR #2189, opened 2025-08-21
fix(modal-checkout): support stripe optimized checkout

chickenn00dle:
### All Submissions:

Closes https://linear.app/a8c/issue/NPPM-2175/testing-woocommerce-stripe-gateway-optimized-checkout-suite

This PR adds support for the new Stripe Optimized Checkout in our modal checkout. We do this by adjusting our payment method template override to more closely match what is expected by Stripes scripts

Screenshot 2025-08-25 at 13 48 04

- Enable stripe payment gateway, and ensure Optimized Checkout is enabled (WooCommerce > Payments > Manage (next to Stripe) > Settings)
- As a reader, trigger modal checkout via any checkout or donation button
- Smoke test modal checkout with:
  - Optimized Checkout disabled
  - Multiple stripe payment options enabled
  - Multiple gateways enabled
  - Only stripe enabled
  - Only any other gateway enabled

### Other information:

### DIFF
--- a/src/modal-checkout/checkout.scss
+++ b/src/modal-checkout/checkout.scss
@@ -412,94 +412,20 @@
 	.wc_payment_methods {
 		margin: var(--newspack-ui-spacer-5, 24px) 0;
 
-		// Largely copied and reorganized from newspack-ui/forms/_labels.scss
-		// since WooCommerce forms are different from our standard forms.
-		span.newspack-ui__input-card {
-			border: 1px solid var(--newspack-ui-color-border, #ddd);
-			border-radius: var(--newspack-ui-border-radius-m);
-			color: var(--newspack-ui-label-color, #000);
-			cursor: pointer;
-			display: block;
-			font-family: var(--newspack-ui-font-family, system-ui);
-			font-size: var(--newspack-ui-font-size-s, 16px);
-			font-weight: 600;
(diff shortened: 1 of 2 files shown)


miguelpeixe (changes requested):
Everything worked smoothly except for the button to "Save payment information":

https://github.com/user-attachments/assets/5a0c4aab-40b3-440c-838f-e24a21a59b96

After clicking, the form glitches a bit and clears the input values. This doesn't happen with the optimized checkout turned off.

chickenn00dle:
Good catch @miguelpeixe.

I think this is actually a bug on the WooCommerce Gateway Stripe side and is specific to classic checkout which our modal checkout uses. I set up a test site with only woo and stripe active to confirm:

https://github.com/user-attachments/assets/784a9fdb-99fc-4d32-8f68-550fd9748add

If you can confirm on your end as well, I'll get a bug report opened in the stripe repo.

miguelpeixe (approved):
Ah, that's unfortunate... Thank you for looking into it!

Approving, as this is not an issue with our code.

chickenn00dle:
Just noting I opened a bug report for the stripe gateway issue here: https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4629

PR #2202, opened 2025-09-10
feat: better error messages for limited products

leogermani:
### All Submissions:

Improves the error message when a user tries to buy a subscription they already have. Currently, the error message only covers the scenario that subscriptions are limited to one active sub per user. This PR also adds a message when the limitation is one of any status.

1. Setup a subscription
2. As a reader, buy that subscription via our checkout modal
3. As an admin, edit the Product, in the Advanced tab, in the  "Limit subscription" option, set it to "Limit to one active subscription"
4. In a new window, try to buy the subscription again with the same email used on step 2. Confirm you see the existing error message:

image

5.  As an admin, edit the Product, in the Advanced tab, in the  "Limit subscription" option, set it to "Limit to one of any status"
6. In a new window, try to buy the subscription again with the same email used on step 2. Confirm you see the new error message:

image

7. Does the new message looks good? Any suggestions to the copy?

### Other information:

### DIFF
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -1533,7 +1533,7 @@ final class Modal_Checkout {
 	 * @return string
 	 */
 	public static function get_subscription_limited_message_any() {
-		return __( "You're already a subscriber! You can only have one subscription at a time. If you wish to renew an expired subscription, please Sign In and visit the subscriptions page on your account.", 'newspack-blocks' );
+		return __( "You're already a subscriber! You can only have one subscription at a time. If you wish to renew an expired subscription, please sign in and visit the subscriptions page on your account.", 'newspack-blocks' );
 	}
 
 	/**
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -1515,7 +1515,7 @@ final class Modal_Checkout {
 	}
 
 	/**
-	 * Alternative error message to show when limiting a subscription product's purchase.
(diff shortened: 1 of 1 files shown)


includes/class-modal-checkout.php
```diff
@@ -1527,6 +1527,15 @@ public static function get_subscription_limited_message() {
 		);
 	}
 
+	/**
+	 * Alternative error message to show when limiting a subscription product's purchase to one of any status.
+	 *
+	 * @return string
+	 */
+	public static function get_subscription_limited_message_any() {
+		return __( "You're already a subscriber! You can only have one subscription at a time. If you wish to renew an expired subscription, please Sign In and visit the subscriptions page on your account.", 'newspack-blocks' );
```
dkoo: Nit:

```suggestion
		return __( "You're already a subscriber! You can only have one subscription at a time. If you wish to renew an expired subscription, please sign in and visit the subscriptions page on your account.", 'newspack-blocks' );
```

It would also be great if we could omit the "Could not complete this transaction. Please contact us for assistance" part of the message in this scenario, but I'd consider that a bonus.

dkoo (approved):
Does the job of improving the messaging in this specific scenario. Just one non-blocking nitpick about the capitalization—it's subjective. 

Also, I notice if I try to buy a second subscription of the product while signed in, I still get this more cryptic error message. Is it possible to clarify this one, too?

Screenshot 2025-09-12 at 10 38 12 AM

PR #2206, opened 2025-09-12
feat(checkout-button): add justification support

chickenn00dle:
### All Submissions:

This PR adds justification support to the checkout button block to be more inline with how the regular button block works:

Screenshot 2025-09-12 at 16 08 20

While working on https://github.com/Automattic/newspack-plugin/pull/4176 I noticed the checkout button block doesn't support justification like the regular button block does.

1. Add a checkout button to any page or post
2. Notice the justification settings in both the block controls and the sidebar
3. Select different justification settings and confirm the button moves in the editor
4. Save the different settings and view on the front end

### Other information:

### DIFF
--- a/src/blocks/checkout-button/block.json
+++ b/src/blocks/checkout-button/block.json
@@ -99,7 +99,16 @@
 			"__experimentalDefaultControls": {
 				"radius": true
 			}
-		}
+		},
+		"layout": {
+		  "default": {
+		    "type": "flex"
+      },
+      "allowJustification": true,
+      "allowVerticalAlignment": false,
+      "allowOrientation": false,
+      "allowSwitching": false
+    }
 	},
 	"styles": [
 		{
--- a/src/blocks/checkout-button/edit.js
+++ b/src/blocks/checkout-button/edit.js
@@ -207,7 +207,7 @@ function ProductControl( props ) {
 
 function CheckoutButtonEdit( props ) {
 	const { attributes, setAttributes, className } = props;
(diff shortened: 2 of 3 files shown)


dkoo (changes requested):
The new justification attributes work! But it seems to break the existing `width` settings in the main block attributes:

Screenshot 2025-09-15 at 10 56 19 AM

It still works in the editor preview, but seems to be ignored on the front-end. As an aside, maybe this panel should be moved to the "Styles" sidebar instead of the main attributes sidebar? But that's non-blocking for this PR.

miguelpeixe (changes requested):
Justification is a feature for the Buttons block, not an individual Button block. This approach requires the parent element to have `display: flex;`, which we don't have control over.

The block does support "align center", which IMO should cover the most common use case:

Image

I noticed that [this recent change](https://github.com/Automattic/newspack-blocks/pull/2196) broke the center alignment in the editor, but it works in the frontend. I just pushed this PR to fix the styles in the editor: https://github.com/Automattic/newspack-blocks/pull/2207

miguelpeixe:
Whoops, sorry @dkoo, I was mid-review and didn't see yours before submitting mine.

dkoo:
Should this PR be closed in favor of #2207, then?

chickenn00dle:
> Should this PR be closed in favor of https://github.com/Automattic/newspack-blocks/pull/2207, then?

WFM. Closing in favor of https://github.com/Automattic/newspack-blocks/pull/2207

PR #2209, opened 2025-09-18
Support for iframe auto-resizing

claudiulodro:
### All Submissions:

Closes NPPD-878. This PR adds support for iframes that send resize messages from the child page to the parent page. Currently iframes have a static vertical size, and this PR lets them have dynamic vertical sizes if the embedded page supports that.

1. You can use https://community.bendsource.com/bend/Responsive/Components/Landing/CalendarPicksFrame.html as a sample URL, because these embeds definitely support the necessary messages. Before applying this PR, create an iframe block with this URL. Resize the screen to a variety of sizes. Observe that at some sizes, the contents of the iframe overflow the container of the iframe:

Screenshot 2025-09-18 at 3 00 35 PM

2. Apply this PR. Refresh the page and go through the screen sizes again. Observe that the height of the iframe changes dynamically so that it never overflows:

Screenshot 2025-09-18 at 3 01 54 PM

### Other information:

### DIFF
--- a/src/blocks/iframe/view.php
+++ b/src/blocks/iframe/view.php
@@ -123,14 +123,27 @@ function newspack_blocks_get_iframe_html( $mode, $src, $height, $width, $is_full
 
 		// Add a listener for dynamic resizing if the iframe supports it.
 		window.addEventListener("message", function(event) {
-			if(event.data.height && iframe.contentWindow === event.source) {
+			// Reject messages from untrusted origins.
+			if (event.origin !== new URL(iframe.src).origin || iframe.contentWindow !== event.source) {
+				return; 
+			}
+
+			var iframeHeight = 0;
+			if (event.data && event.data.height){
+				if (typeof event.data.height === "number") {
+					iframeHeight = event.data.height;
+				} else if (typeof event.data.height === "string") {
+					iframeHeight = Number(event.data.height);
+				}
+			}
+			if (!isNaN(iframeHeight) && iframeHeight > 0) {
 				// Remove height from the iframe's parent element if needed.
-				if (iframe.parentElement) {
+				if (iframe.parentElement && iframe.parentElement.style.height !== "auto") {
 					iframe.parentElement.style.height = "auto";
 				}
 
 				// Set the new height dynamically.
-				iframe.style.height = event.data.height + "px";
+				iframe.style.height = iframeHeight + "px";
 			}
 		});
 	</script>
--- a/src/blocks/iframe/view.php
+++ b/src/blocks/iframe/view.php
@@ -121,6 +121,18 @@ function newspack_blocks_get_iframe_html( $mode, $src, $height, $width, $is_full
 			clearInterval(timerId);
 		}
 
+		// Add a listener for dynamic resizing if the iframe supports it.
+		window.addEventListener("message", function(event) {
+			if(event.data.height && iframe.contentWindow === event.source) {
+				// Remove height from the iframe's parent element if needed.
+				if (iframe.parentElement) {
+					iframe.parentElement.style.height = "auto";
+				}
+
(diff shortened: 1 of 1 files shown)


src/blocks/iframe/view.php
```diff
@@ -121,6 +121,18 @@ function newspack_blocks_get_iframe_html( $mode, $src, $height, $width, $is_full
 			clearInterval(timerId);
 		}
 
+		// Add a listener for dynamic resizing if the iframe supports it.
+		window.addEventListener("message", function(event) {
```
rbcorrales, line 125: This code accepts messages from any origin. Can we add origin validation to prevent XSS and malicious DOM manipulation?

```js
if (event.origin !== new URL(iframe.src).origin) {
    return; // Reject messages from untrusted origins.
}
```

This only partially addresses a possible vulnerability, since it would depend on the embedded source not being compromised, but it's better than leaving it open.

src/blocks/iframe/view.php
```diff
@@ -121,6 +121,18 @@ function newspack_blocks_get_iframe_html( $mode, $src, $height, $width, $is_full
...
+			if(event.data.height && iframe.contentWindow === event.source) {
```
rbcorrales: The code doesn't validate the type or bounds of the height value, which could lead to unexpected behavior or potential exploitation. Sample possible validation (you can also parse to int first, if the message is a string, but I think in the example you shared it's already an integer):

```js
if (event.data &&
    typeof event.data.height === 'number' &&
    event.data.height > 0 &&
    event.data.height <= 2000 &&
    iframe.contentWindow === event.source) {
...
}
```

src/blocks/iframe/view.php
```diff
@@ -121,6 +121,18 @@ function newspack_blocks_get_iframe_html( $mode, $src, $height, $width, $is_full
...
+				// Remove height from the iframe's parent element if needed.
+				if (iframe.parentElement) {
+					iframe.parentElement.style.height = "auto";
+				}
```
rbcorrales: Here we can  add condition to avoid unnecessary DOM updates:

```js
if (iframe.parentElement && iframe.parentElement.style.height !== "auto") {
    iframe.parentElement.style.height = "auto";
}
```

rbcorrales (changes requested):
The change works as expected, but please see my comments about it.

src/blocks/iframe/view.php
claudiulodro: Fixed in https://github.com/Automattic/newspack-blocks/pull/2209/commits/12a87f1b03c6c2f762aa9a65e8d14b0ae437953b

src/blocks/iframe/view.php
claudiulodro: Fixed in https://github.com/Automattic/newspack-blocks/pull/2209/commits/12a87f1b03c6c2f762aa9a65e8d14b0ae437953b. 

I didn't do a max height for two reasons:
1. A lot of people use this feature to embed a whole premium article type of thing which could be many screens tall.
2.  The `<= 2000` was triggering [this long standing bug](https://core.trac.wordpress.org/ticket/50461) which caused the next `&&` to get rendered as `&#038;&#038;` and created syntax errors :)

src/blocks/iframe/view.php
claudiulodro: Also fixed in https://github.com/Automattic/newspack-blocks/pull/2209/commits/12a87f1b03c6c2f762aa9a65e8d14b0ae437953b

claudiulodro:
Thanks for the excellent review, @rbcorrales! I've made your suggested changes

rbcorrales (approved):
Thanks for addressing the comments!

PR #2214, opened 2025-09-23
fix: clarify NRH dependency for Campaign ID sidebar attribute

dkoo:
### All Submissions:

Renames the Donate block's "Campaign" sidebar panel to "News Revenue Hub Settings" and hides it unless the active platform is NRH, to clarify the purpose of this block attribute.

Closes NPPM-2244.

1. On `trunk`, add a Donate block to a post/page and add a Campaign ID value to it. Save.
2. Check out this branch.
3. In **Audience > Configuration > Checkout & Payment**, set your reader revenue platform to something other than News Revenue Hub.
4. Edit the existing Donate block and add a new Donate block to a post or page. Confirm that there's no "Campaign" or "News Revenue Hub Settings" sidebar panel when the Donate block is selected.
5. Set your reader revenue platform to News Revenue Hub.
6. Edit the existing Donate block and add a new one, and confirm that the "News Revenue Hub Settings" panel appears and can still be edited and saved after refreshing the editor. Confirm that the Donate block you created in step 1 still has the same Campaign ID you entered when you last saved it.

### Other information:

### DIFF
--- a/src/blocks/donate/edit/index.tsx
+++ b/src/blocks/donate/edit/index.tsx
@@ -423,17 +423,19 @@ const Edit = ( { attributes, setAttributes, className }: EditProps ) => {
 						label={ __( 'Button Color', 'newspack-blocks' ) }
 					/>
 				</PanelBody>
-				<PanelBody title={ __( 'Campaign', 'newspack-blocks' ) } initialOpen={ false }>
-					<TextControl
-						label={ __( 'Campaign ID', 'newspack-blocks' ) }
-						value={ attributes.campaign || '' }
-						onChange={ ( value: string ) =>
-							setAttributes( {
-								campaign: value,
-							} )
-						}
-					/>
-				</PanelBody>
+				{ settings.platform === 'nrh' && (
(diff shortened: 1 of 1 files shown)


src/blocks/donate/edit/index.tsx
```diff
@@ -423,17 +423,19 @@ const Edit = ( { attributes, setAttributes, className }: EditProps ) => {
...
-						label={ __( 'Campaign ID', 'newspack-blocks' ) }
-						value={ attributes.campaign || '' }
-						onChange={ ( value: string ) =>
-							setAttributes( {
-								campaign: value,
-							} )
-						}
-					/>
-				</PanelBody>
+				{ settings.platform === 'nrh' && (
```
chickenn00dle, line 426: Non-blocking question: If a donation block already contains a campaign id, should we still show this setting even if nrh is not the platform?

chickenn00dle (approved):
Works as expected. Left a non-blocking question, but this is good to go!

src/blocks/donate/edit/index.tsx
dkoo: I think no, because it only has any effect if NRH is the platform.

PR #2220, opened 2025-09-29
fix(iframe-block): validate file extensions before allowing upload

dkoo:
### All Submissions:

This plugs a security vulnerability in the iframe block that allows files to be uploaded with an unsupported file extension.

Closes NPPM-2274.

1. On `release`, add an iframe block to a post or page.
2. Create a .zip archive containing the required `index.html` plus second valid file of a supported type, but renamed with an unsupported extension such as `.php`.
4. Upload the .zip to the iframe block.
5. In the `wp-content/uploads` folder, observe that the .zip contents are extracted to a subdirectory inside a `newspack_iframes` directory, including the file named with `.php` because the file contents are validated but not the extension.
6. Check out this branch and repeat the upload. This time confirm that the file with `.php` in the name is not extracted.

### Other information:

### DIFF
--- a/src/blocks/iframe/class-wp-rest-newspack-iframe-controller.php
+++ b/src/blocks/iframe/class-wp-rest-newspack-iframe-controller.php
@@ -206,6 +206,19 @@ class WP_REST_Newspack_Iframe_Controller extends WP_REST_Controller {
 	}
 
 	/**
+	 * Validate the file extension of a file.
+	 *
+	 * @param string  $file_name The name of the file.
+	 * @param boolean $in_archive If true, the file is inside an archive, which can contain HTML/CSS/JS files.
+	 *
+	 * @return boolean
+	 */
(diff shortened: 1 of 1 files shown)


chickenn00dle (approved):
This works as expected, however I wonder if this [validate_file](https://github.com/Automattic/newspack-blocks/blob/33afbebf4e7447574a615b091902653d08a0cd4d/src/blocks/iframe/class-wp-rest-newspack-iframe-controller.php#L187) method is not a better place to check the file extension?

It looks to be called prior to processing upload files here: https://github.com/Automattic/newspack-blocks/blob/216ce7880b376536a44af04402484b05f0ae95b4/src/blocks/iframe/class-wp-rest-newspack-iframe-controller.php#L262

I also noted there are no errors in the editor to indicate a file has been rejected so theres no way to know if anything needs to be fixed as the uploader.

That said, since this is a critical fix, I won't let these be blockers and will approve as is.

PR #2244, opened 2025-11-07
fix: content loop load more for custom taxs

leogermani:
### All Submissions:

NPPM-2373

Fixes the "load more" button in the content Loop block for Custom taxonomies.

When you add a block that has a filter for custom taxonomies like Brands, and a "Load more posts" button, when you click the button, the posts that are loaded do not respect the filter for the custom taxonomy.

This basically reverts part of the changes introduced in https://github.com/Automattic/newspack-blocks/pull/1454

It seems safe to just remove it, as the `$attributes` var is only being used to build the "more posts" link and in the templates. The link is what was broken, and the templates don't look for custom taxonomies for anything... I have no idea why I added this code in the first place.

1. Perform all the tests described in https://github.com/Automattic/newspack-blocks/pull/1454
2. Add a block with the option to "Show Load More posts button"  ON and a filter by BRAND
3. Visit the block on the front end and make sure the posts that are loaded respect the Brand filter.
4. Also run the same test for the Infinite Scroll option

### Other information:

### DIFF
--- a/src/blocks/homepage-articles/view.php
+++ b/src/blocks/homepage-articles/view.php
@@ -323,19 +323,6 @@ function newspack_blocks_render_block_homepage_articles( $attributes ) {
 		$styles = 'color: ' . $attributes['customTextColor'] . ';';
 	}
 
-	// Handle custom taxonomies.
-	if ( isset( $attributes['customTaxonomies'] ) ) {
-		$custom_taxes = $attributes['customTaxonomies'];
-		unset( $attributes['customTaxonomies'] );
-		if ( is_array( $custom_taxes ) && ! empty( $custom_taxes ) ) {
-			foreach ( $custom_taxes as $tax ) {
-				if ( ! empty( $tax['slug'] ) && ! empty( $tax['terms'] ) ) {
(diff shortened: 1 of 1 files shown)


src/blocks/homepage-articles/view.php
wil-gerken, line 1: Hi @leogermani, we weren't able to identify anything that needs the custom taxonomy split out like this, so the full removal seems great.

That said, if you think there is a concern, perhaps we should just remove the unset that is causing the issue. That would keep the intent of this original code in place. 

`unset( $attributes['customTaxonomies'] ); `

Any strong opinion either way?

leogermani: Yeah, that's not a bad approach to be super safe... but if we're really sure we are not using it anywhere I'd rather remove it. Let's do some testing for regressions.

Have you had a chance to test it with the infinite scroll option?

wil-gerken: Yes, I did test infinite scroll, and it works perfectly. They use the same $articles_rest_url so this fix addresses them both nicely.

Leaner code sounds good to me. This all looks great.

wil-gerken (approved):

PR #2245, opened 2025-11-07
fix: don’t show trial info in modal checkout if not applicable

dkoo:
### All Submissions:

Related https://github.com/Automattic/newspack-plugin/pull/4287. If a user who already purchased a free trial tries to purchase the same product, they shouldn't be able to get a free trial. This PR ensures that in that scenario, the checkout modal won't show the free trial info in the product info or transaction details.

Closes NPPM-2302.

See https://github.com/Automattic/newspack-plugin/pull/4287.

### Other information:

### DIFF
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -1829,7 +1829,7 @@ final class Modal_Checkout {
 	 *
 	 * @return false|int User ID if found by email address, false otherwise.
 	 */
-	private static function get_user_id_from_email() {
+	public static function get_user_id_from_email() {
 		$billing_email = filter_input( INPUT_POST, 'billing_email', FILTER_SANITIZE_EMAIL );
 		if ( $billing_email ) {
 			$customer = \get_user_by( 'email', $billing_email );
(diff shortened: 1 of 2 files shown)


includes/class-modal-checkout.php
```diff
@@ -1829,7 +1829,7 @@ public static function ensure_base_default_customer_address( $option_value ) {
 	 *
 	 * @return false|int User ID if found by email address, false otherwise.
 	 */
-	private static function get_user_id_from_email() {
+	public static function get_user_id_from_email() {
```
leogermani, line 1832: Was this intentional?

leogermani (approved):

includes/class-modal-checkout.php
dkoo: Yes, in the corresponding Plugins PR if the user is not logged in we attempt to check the user's history by email: https://github.com/Automattic/newspack-plugin/pull/4287/files#diff-ae452f417e51fa1a6021e09104bbc718889cf86eae8b4f68ad7b719e502d443dR139

This is the part that only works intermittently, though—I think the order in which the hooks are firing may not be consistent. If the free trial product is not limited in any way, and the user is not logged in while attempting to start another free trial with an email address that already has one, then the free trial may or may not be disabled during the purchase. It's blocked more consistently if the product is limited to one active subscription per user.

dkoo:
Waiting to merge this one as it goes along with https://github.com/Automattic/newspack-plugin/pull/4287.

PR #2247, opened 2025-11-12
feat(modal-checkout): processing payment screen

miguelpeixe:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-theme/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

NPPD-941

Implements a  "Processing Payment" spinner screen when placing the order in the modal checkout:

https://github.com/user-attachments/assets/e5676d2b-b968-4299-874b-0d6657377eca

### How to test the changes in this Pull Request:

1. Checkout this branch and https://github.com/Automattic/newspack-plugin/pull/4300
2. Start a modal checkout flow via a Checkout Button block
3. Enter an invalid card number (`4000000000000002`) and place the order
4. Confirm the new Processing Payment screen renders and exits upon error
5. Enter a valid card number (`4242424242424242`) and place the order
6. Confirm the new Processing Payment screen renders with the correct message interval
7. Confirm the payment succeeds
8. Without reloading the page, start the modal checkout again
9. Confirm the spinner **does not** include "Processing payment" text

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #2255, opened 2025-11-17
fix: handle checkout form inline errors and expired sessions

dkoo:
### All Submissions:

#### Recover from expired checkout sessions

Attempts to load a new modal checkout session if the current checkout session becomes expired or invalid. This should help users to pick up where they left off if their checkout session hits a dead end. Requires https://github.com/Automattic/newspack-plugin/pull/4310 to prevent the checkout session from attempting to reload the session within the modal checkout iframe, which will result in invalid nonce values and an expired session.

#### Handle inline checkout errors

#2247 added a "processing payment..." overlay when attempting to complete a transaction via modal checkout. This includes error handling for if the request fails on the server and returns an error to the AJAX request, but it doesn't handle inline errors thrown by and shown inline by the payment gateway form, which happens in an iframe element and prevents the checkout AJAX request from proceeding.

In this scenario, the checkout form will [trigger a `checkout_place_order` jQuery event](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/legacy/js/frontend/checkout.js#L772), but will not trigger any other event to indicate that the iframe's form validation failed, so we need to manually check the form's checkout processing state and infer an error state if it stops processing before we receive the `checkout_place_order_success` event.

Closes NPPM-2394.

When testing, make sure your Woo and Newspack settings allow for anonymous checkouts without registering or logging in first.

#### Recover from expired checkout sessions

1. On `trunk`, start a modal checkout as an anonymous reader.
2. On the second screen of the modal, attempt to complete checkout with a card number which will be declined ([see Stripe test card numbers](https://docs.stripe.com/testing#declined-payments)). Observe that the modal checkout goes back to the first screen, and that after this you're unable to proceed past this screen or complete checkout.
3. Check out this branch and https://github.com/Automattic/newspack-plugin/pull/4310.
4. In a new anonymous browser session, repeat steps 1-2:
5. After checking out with the declined card number, the checkout shows the proper error message:

Screenshot 2025-11-17 at 2 54 05 PM

6. After resubmitting this form, confirm that you briefly see an error message stating "We ran into a problem processing this request. Trying again..." before the modal checkout reloads itself back to the second screen with the same transaction details already loaded.
7. After the checkout modal reloads, confirm that you can successfully complete the transaction with a valid card number.

#### Handle inline checkout errors

1. On `trunk`, start a modal checkout as an anonymous reader.
2. On the second screen of the modal, attempt to complete checkout without filling in any payment details. Observe that the modal hangs in the "processing payment..." overlay state and never resolves (you can also use dev tools to hide the overlay and see that there are inline form errors hidden underneath it).
5. Check out this branch, repeat in a new anonymous browser session, and confirm that after a brief moment, the overlay disappears and allows you to see and correct the inline form errors.

### Other information:


laurelfulford (changes requested):
Hey @dkoo! I ran into some weirdness (description and video below) - just let me know if you can't recreate, and I can try on a JN site! 

**Recover from expired checkout sessions**

This issue is _technically_ fixed (you get the new error message and if you enter a working test credit card the second time it lets you complete the transaction), but it doesn't act like you described:

After entering a credit card that won't work, I either get this (briefly showing the second screen and error before being getting booted to the first screen), or only get the first screen again and no visible error.

https://github.com/user-attachments/assets/0bc38c8f-4f25-46dd-8339-09186699d391

On trunk, this issue also doesn't present the same for me (so maybe there's something there): instead of kicking me back to the first screen, I get the same infinite loading I get when trying to submit empty credit card information.

**Handle inline checkout errors**

This works as described!

laurelfulford:
Hey @dkoo! I retested these PRs on [a fresh JN site](https://potential-shearwater-christmas.jurassic.ninja/) and was able to narrow this down a bit more:

The issue I found only happens when trying to purchase a giftable subscription:
* It happens whether or not you check the "This purchase is a gift" checkbox. 
* If you do check the "This purchase is a gift" checkbox, when it kicks you back to the first screen it is no longer checked. The giftee's email is still saved, though, and appears when you re-check it.

If I test it with a non-giftable subscription, I don't think it's quite 1:1 what what you described, but it's a lot closer! 

* When I enter a failing credit card number, I briefly see the error before the second screen is reloaded.
* I don't see the "We ran into a problem processing this request. Trying again..." message at all, but I'm unclear what's triggering that (entering a working credit card number works fine).

This video shows what I get (it also shows that I don't know how to format postal codes!):

https://github.com/user-attachments/assets/1c2dd8dd-4115-4190-b4c8-767023636fc1

laurelfulford (changes requested):
Thanks @dkoo! 

The behaviour for gitftable products is looking a lot better! 

For both giftable and non-giftable products, the error message no longer explains what the issue was - was this deliberate?

Image

Image

I think the styles also need to be fixed, but that could be a separate PR. We have CSS in the new My Account styles that should fix it [here](https://github.com/Automattic/newspack-plugin/blob/6e3c5f85307e8e6bbf3fbce43ff49c16c024f670/src/newspack-ui/scss/elements/woocommerce/_my-account.scss#L18-L47) -- I think we just need to make that more general so it applies anytime these classes are inside of a `.newspack-ui` container, not just the My Account ones.

dkoo:
> I think the styles also need to be fixed, but that could be a separate PR. We have CSS in the new My Account styles that should fix it [here](https://github.com/Automattic/newspack-plugin/blob/6e3c5f85307e8e6bbf3fbce43ff49c16c024f670/src/newspack-ui/scss/elements/woocommerce/_my-account.scss#L18-L47) -- I think we just need to make that more general so it applies anytime these classes are inside of a `.newspack-ui` container, not just the My Account ones.

Good point, I almost didn't notice that it had the wrong styles. [e44f4e3](https://github.com/Automattic/newspack-plugin/pull/4310/commits/e44f4e3e594173ac0f3cea5228b66debfcd3ff90) in the [other PR](https://github.com/Automattic/newspack-plugin/pull/4310) should fix this by applying `newspack-ui` notice styles to all Woo notices that are nested inside a `.newspack-ui` wrapper.

Screenshot 2025-11-21 at 10 10 50 AM

As for the generic "We ran into an error processing this request...", that message should only be shown when the checkout session becomes invalidated due to a payment processing error that happens in the same session where a new customer account is created, as the creation and authentication of the customer account will invalidate the checkout session which started as an anonymous session. Because this isn't the only scenario when the checkout session could become invalidated, the error message has to be fairly generic.

You should still see the "Your card was declined" error the first time you try to complete checkout with an invalid card number, then when you try to complete the checkout again, that should trigger the checkout modal refresh and the generic error message.

laurelfulford (approved):
Thanks for tweaks, @dkoo! 🎉 

I ran through the checkout with a bad card:
* Without login required with a giftable sub
* Without login required with an ungiftable sub
* Without login required with regular product

In all cases:
* I got the credit card declined message
* When I tried with a good card in the same window, I got the generic error message
* When I restarted the transaction (closed the modal and reopened it), it let me run it successfully with a good card

I also tested the same products with login required, just in case - in all cases I just got the credit card declined message, then was able to get through fine with a good card in the same modal (as expected since there isn't the account weirdness when you're logged in!).

The error messages are also looking good! :shipit:

PR #2263, opened 2025-11-27
feat: supress carousel block from feeds NPPM-2413

leogermani:
### All Submissions:

Adds the [same behavior of the Content Loop block](https://github.com/Automattic/newspack-blocks/commit/51f72ebd6a8db158bc93ea2a4ceaf2b044653944) and exclude it from feeds

1. Add a carousel block to a post
2. Make sure it works as expected both in the editor and in the frontend
3. Confirm it's no longer included in feeds

### Other information:

### DIFF
--- a/src/blocks/carousel/view.php
+++ b/src/blocks/carousel/view.php
@@ -13,6 +13,12 @@
  * @return string Returns the post content with latest posts added.
  */
 function newspack_blocks_render_block_carousel( $attributes ) {
+
+	// Don't output the block inside RSS feeds.
+	if ( is_feed() ) {
+		return;
+	}
+
 	static $newspack_blocks_carousel_id = 0;
 	global $newspack_blocks_post_id;
 


rbcorrales (approved):
I was doing this for Pugpig feeds only:
```php
	/**
	 * Prevent carousel blocks from rendering in the Pugpig feed.
	 *
	 * @param string|null $pre_render The pre-rendered content. Default null.
	 * @param array       $parsed_block An associative array of the block being rendered.
	 * @return string|null The pre-rendered content.
	 */
	public static function prevent_carousel_rendering( $pre_render, $parsed_block ) {
		if ( Utils::is_pugpig_feed() && 'newspack-blocks/carousel' === ( $parsed_block['blockName'] ?? '' ) ) {
			return '';
		}

		return $pre_render;
	}
```

But if we don't want to see it anywhere, this is better.

PR #2270, opened 2026-01-06
fix: add space to donate block label in editor

laurelfulford:
### All Submissions:

I noticed this while working on something else:

The Donate block has a new per month/per year frequency text in the "Other" field label, but there's no space before this label in the editor. This PR fixes that.

1. Add the Donate block to the editor and pick the Monthly or Annually tab; adjust it so you can see the 'Other' field, and note the text:

CleanShot 2026-01-06 at 10 44 18

2. Apply this PR and run `npm run build`.
3. Refresh the editor and confirm the text is properly spaced:

CleanShot 2026-01-06 at 10 46 18

### Other information:

### DIFF
--- a/src/blocks/donate/utils.ts
+++ b/src/blocks/donate/utils.ts
@@ -49,11 +49,12 @@ export const getFrequencyLabel = ( frequencySlug: DonationFrequencySlug, hideOnc
 		? hideOnceLabel
 			? ''
 			: __( 'once', 'newspack-blocks' )
-		: ' ' + sprintf(
+		: ' ' +
+		sprintf(
 				// Translators: %s is the frequency (e.g. per month, per year).
 				_x( 'per %s', 'per `Frequency`', 'newspack-blocks' ),
 				frequencySlug
-		  );
+		);
 };
 
 export const getFormattedAmount = ( amount: number, withCurrency = false ) => {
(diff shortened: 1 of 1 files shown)


src/blocks/donate/utils.ts
```diff
@@ -49,11 +49,12 @@ export const getFrequencyLabel = ( frequencySlug: DonationFrequencySlug, hideOnc
 		? hideOnceLabel
 			? ''
 			: __( 'once', 'newspack-blocks' )
-		: sprintf(
-				// Translators: %s is the frequency (e.g. per month, per year).
-				_x( 'per %s', 'per `Frequency`', 'newspack-blocks' ),
-				frequencySlug
-		  );
+		: ' ' +
```
chickenn00dle, line 52: Hmm, I wonder if this is friendly for all languages? I'm just guessing there might be languages where forcing the frequency string to come after the amount string will result in some potentially funky translations.

Maybe it would be better to add the frequency within the `getFormattedAmmount` method below so we can generate these translated strings consistently?

WDYT?

chickenn00dle: I realize this is the case already before this PR, so I'll go ahead and approve. But I do think this is a good chance to address this.

chickenn00dle (approved):

src/blocks/donate/utils.ts
laurelfulford: Thanks @chickenn00dle! 

> Hmm, I wonder if this is friendly for all languages? I'm just guessing there might be languages where forcing the frequency string to come after the amount string will result in some potentially funky translations.

Yeah, that's a good call! I made myself a separate issue to look at this a little closer (NPPM-2509) 🙂 I also see there's a separate `getFrequencyLabelWithAmount` const that possibly has the same missing space issue (and order issue - I'm not sure yet! This is the first time I've noticed this change).

PR #2282, opened 2026-01-24
feat: add tag labels support to Content Loop block

jason10lee:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)?

This PR is one of a set that adds the ability to display selected tags as labels.  Changes live in the `feat/tag-labels` topic branch on the *newspack-blocks*, *newspack-plugin*, and *newspack-theme* projects.

See also:  https://github.com/Automattic/newspack-theme/pull/2613
See also:  https://github.com/Automattic/newspack-plugin/pull/4381

Labels will appear above the headline, after any category or sponsorship labels.

By default, the text of the tag label will be the same as the tag name.  Optionally, the label text can be customized (e.g., to something shorter and more suitable for use as a label).  Additionally, the label color can be customized in the Newspack Theme customizer.

Addresses [NPPD-383: "Breaking News" or "Developing" Label](https://linear.app/a8c/issue/NPPD-383/breaking-news-or-developing-label).

#### Basic operation:

1. Pull [this topic branch](https://github.com/Automattic/newspack-blocks/tree/feat/tag-labels), [the newspack-plugin topic branch](https://github.com/Automattic/newspack-plugin/tree/feat/tag-labels), and [the corresponding newspack-theme topic branch](https://github.com/Automattic/newspack-theme/tree/feat/tag-labels).
2. Create a post tag.  Confirm that you see a 'Display as label' checkbox.
5. Check the 'Display as label' checkbox.  Confirm the 'Label text' field is now visible.
6. Save the tag.
7. Tag a post with your new tag.  This post should be one that shows up in a Content Loop block.
8. Confirm the post now shows a label with the tag name on it in the Content Loop block.
9. Edit the post tag again.  Uncheck the 'Display as label' checkbox.  Confirm the 'Label text' field is disabled and save.
10. Confirm the label is no longer shown on the post.

#### Multiple labels:
1. Enable tag labels on your test tag, as above.
2. Tag your test post with another tag that doesn't have labels enabled.  Confirm only the test tag with labels enabled displays a label on the post.
3. Enable labels on your second test tag.  Confirm it also shows up on the post.

#### Custom label text (flag):
1. Edit one of your test post tags.  Check the 'Display as label' field to enable the 'Label text' field.
2. Add text of your choice (something different from the tag name!) to the 'Label text' field and save.
3. Confirm the test post shows a label with your custom text in all contexts.
4. Edit the post tag and verify your custom text is still shown in the 'Label text' field.
5. Clear the 'Label text' field.  Confirm the original tag name is again shown as a placeholder and save.
6. Confirm the test post shows a label with the original tag name.
7. Edit the post tag, uncheck the 'Display as label' field and save.  Confirm the label's no longer shown.
8. Edit the post tag and add custom text to the 'Label text' field again.  Save.  (Optionally:  check your custom text still shows up!)
9. Edit the post tag and uncheck the 'Display as label' field.  Confirm no label is shown on the test post.

#### Custom label color:
1. Add a tag with labels enabled to a test post.  
2. Ensure you're using a Newspack theme, and open the theme customizer (**Appearance** > **Customize**).
3. Select **Tag Labels** from the menu.
4. Click on 'Select color' to select a color for the label.  If a post tagged with that label is visible in the preview pane on the right, you should see it change!  The text color may change as well to better contrast with the new label color.
5. Hit the 'Publish' button at the top and check your test post.  It should show the new color scheme.

### Other information:


adekbadek (commented):
_(Superseded by the inline review below.)_

src/blocks/homepage-articles/edit.tsx
```diff
@@ -112,6 +112,21 @@ class Edit extends Component< HomepageArticlesProps > {
...
+					) }{ ' ' }
+					{ post.newspack_tag_labels && (
+						<div className="cat-links tag-labels">
+							{ post.newspack_tag_labels.map( ( newspack_tag_label, index ) => {
+								return newspack_tag_label.link ? (
+									<a key={ index } href={ newspack_tag_label.link } className="tag-label flag">
+										{ newspack_tag_label.flag }
+									</a>
+								) : (
+									<span key={ index } className="flag">
```
adekbadek: The `<span>` fallback is missing the `tag-label` class. The `<a>` variant on line 120 correctly has `className="tag-label flag"`, and the PHP `generate_tag_labels()` applies both classes to both elements.

```suggestion
									<span key={ index } className="tag-label flag">
```

Also: `newspack_tag_labels` should be added to the post type definition in `src/types/index.d.ts` (around line 88, next to `newspack_post_sponsors`):
```ts
newspack_tag_labels?: { flag: string; link: string }[] | false;
```

includes/class-newspack-blocks-api.php
```diff
@@ -157,6 +157,31 @@ public static function newspack_blocks_sponsor_info( $object_info ) {
...
+		);
+		if ( ! empty( $tag_labels ) ) {
+			$tag_labels_info = [];
+			foreach ( $tag_labels as $tag_label ) {
+				$tag_label_info_item = [
+					'flag' => $tag_label['flag'],
+					'link' => $tag_label['link'],
+				];
+				$tag_labels_info[] = $tag_label_info_item;
+			}
```
adekbadek: Nit: this loop extracts `flag` and `link`, but `get_labels_for_post()` already returns exactly those fields. Could simplify:

```php
public static function newspack_blocks_get_tag_labels( $object_info ) {
    $tag_labels = Newspack_Blocks::get_tag_labels( $object_info['id'] );
    return ! empty( $tag_labels ) ? array_values( $tag_labels ) : false;
}
```

(`array_values` because `array_filter` in `get_labels_for_post` can leave gaps in keys.)

includes/class-newspack-blocks.php
```diff
@@ -1089,6 +1089,66 @@ public static function newspack_display_sponsors_and_categories( $sponsors ) {
...
+	 * Generates HTML for given tag labels.
+	 *
+	 * @param array $labels Labels to display.
+	 * @param bool  $links  Whether to include links to tag archives.
+	 * @param array $outer_classes Classes to apply to the outer container.
+	 * @param array $inner_classes Classes to apply to the inner container.
+	 *
+	 * @return string       Tag labels as HTML.
+	 */
+	public static function generate_tag_labels( $labels = null, $links = true, $outer_classes = array( 'tag-labels' ), $inner_classes = array( 'tag-label', 'flag' ) ) {
```
adekbadek: This method duplicates `\Newspack\Tag_Labels::generate_html()` from newspack-plugin (the only difference is hardcoded `<div>` vs configurable `$outer_element`). Since `get_tag_labels()` already gates on the plugin being present, and `display_tag_labels` is only called when labels are non-empty, you could delegate directly:

```php
public static function display_tag_labels( $labels = null, $links = true ) {
    if ( empty( $labels ) || ! class_exists( '\Newspack\Tag_Labels' ) ) {
        return;
    }
    \Newspack\Tag_Labels::display( $labels, $links, 'div' );
}
```

That would drop `generate_tag_labels` entirely. Not blocking, but reduces maintenance surface from 3 copies of the same HTML builder to 1.

adekbadek (commented):
Feature works well. A few inline notes, plus one scope suggestion: the **Carousel block** shares the same REST API data as Homepage Articles but does not render tag labels yet -- worth considering as a follow-up so the label experience is consistent across both content blocks.

rbcorrales:
Swapping the labels since this seems to need some additional work.

includes/class-newspack-blocks.php
jason10lee: Agreed per discussion on `newspack-plugin` [#4381](https://github.com/Automattic/newspack-plugin/pull/4381) and addressed  in [c9f3d7](https://github.com/Automattic/newspack-blocks/pull/2282/commits/9c9f3d7e8c23e53cc2b0413d77b1aed8632a740d).

src/blocks/homepage-articles/edit.tsx
jason10lee: Agreed and addressed in [0a96747](https://github.com/Automattic/newspack-blocks/pull/2282/commits/0a967470696e25414b3edabc7e4aa547347ee9c6).

includes/class-newspack-blocks-api.php
jason10lee: Good call; now that these are standardized, there's no need to open up the internals.  Using the structure directly in [0a96747](https://github.com/Automattic/newspack-blocks/pull/2282/commits/0a967470696e25414b3edabc7e4aa547347ee9c6).

includes/class-newspack-blocks.php
dkoo: Nit: Is `Newspack_Blocks::generate_tag_labels` called anywhere? I'm not seeing any callers in any of the repos anymore—I just see `Newspack\Tag_Labels::generate_html` throughout.

dkoo (changes requested):
Echoing @adekbadek's comment that handling for tag labels is missing from the Carousel block. Aside from that, all is working as described.

dkoo (approved):

dkoo (dismissed):
Whoops, hit the "Approved" button too eagerly. Latest changes for the "Show tag labels" attribute look good, but still need to add support for the Post Carousel block (or follow up in a separate PR).

jason10lee:
Thanks, @dkoo!  We'll tackle the Carousel in a separate PR to avoid scope creep.

src/blocks/homepage-articles/block.json
```diff
@@ -84,6 +84,10 @@
 			"type": "boolean",
 			"default": false
 		},
+		"showTagLabels": {
```
wil-gerken, line 87: One thing worth a deliberate decision: `showTagLabels` defaults to `true`, while the other display toggles here (`showCategory`, `showSubtitle`) default to `false`. So labels turn on for every existing Homepage Posts block as soon as a tag is flagged, rather than being opt-in per block. That may be exactly what you want for this feature — flagging it so it's a deliberate choice rather than an inherited default. Happy either way; leaving it to you. _(Non-blocking.)_

includes/class-newspack-blocks.php
```diff
@@ -1096,6 +1096,53 @@ public static function newspack_display_sponsors_and_categories( $sponsors ) {
...
```
wil-gerken: `get_tag_labels()` and `display_tag_labels()` are both in use, but `generate_tag_labels()` doesn't seem to be called anywhere in newspack-blocks. If you meant it as public API for the theme, makes sense to keep — `newspack-theme` already has its own `newspack_generate_tag_labels()` wrapper. If nothing calls it, I'd consider dropping it to keep the public surface tidy. _(Non-blocking.)_

includes/class-newspack-blocks-api.php
```diff
@@ -269,6 +280,7 @@ public static function posts_endpoint( $request ) {
 				'newspack_post_sponsors'            => self::newspack_blocks_sponsor_info( $data ),
 				'newspack_sponsors_show_author'     => Newspack_Blocks::newspack_display_sponsors_and_authors( $sponsors ),
 				'newspack_sponsors_show_categories' => Newspack_Blocks::newspack_display_sponsors_and_categories( $sponsors ),
+				'newspack_tag_labels'               => self::newspack_blocks_get_tag_labels( $data ),
```
wil-gerken, line 283: One small coverage gap: this adds a REST key (`newspack_tag_labels`), a template branch, and three public methods without tests. A small test in `tests/test-homepage-posts-block.php` asserting the `{ flag, link }` shape would lock in the contract — the PHP side passes through whatever `\Newspack\Tag_Labels::get_labels_for_post()` returns without validating it, and the three repos have to agree on that shape. Not blocking — a useful safety net for a cross-repo feature. _(Non-blocking.)_

wil-gerken (approved):
Tested the feature end-to-end in an isolated env with the `feat/tag-labels` branches for the plugin, blocks, and theme, plus seeded tags. Everything I checked works great: per-tag enable, custom vs. default label text, control tags correctly left off, the editor/REST data path, front-end rendering, the block toggle, and the customizer color with auto-contrast (light backgrounds get dark text, dark backgrounds get light text). The defensive `class_exists()` / `method_exists()` guards are spot-on, so this is safe to land ahead of the plugin and theme PRs — it renders nothing until they do. Nice work.

A few non-blocking notes below — none need to hold the merge; the first is a product call that's entirely yours to make.

src/blocks/homepage-articles/block.json
jason10lee: This is a deliberate choice: since this ships without automatically flagging any tags as labels, label display is gated on both this and the publisher intentionally enabling labels per tag.  Defaulting to `true` lets publishers see their flags immediately and on all surfaces where they're supported without having to hunt down the separate blocks and enable each one too.

includes/class-newspack-blocks.php
jason10lee: Dropped `generate_tag_labels()` as we've done in the other PRs for this feature.  Addressed in [1319aa9](https://github.com/Automattic/newspack-blocks/pull/2282/commits/1319aa93a08726e7c1744315445f1c30bf00283c).

includes/class-newspack-blocks-api.php
jason10lee: New tests in [7b1dd8f](https://github.com/Automattic/newspack-blocks/pull/2282/commits/7b1dd8fcc6c5e6f1f6f0c572cc4e37387a6c7224).

src/blocks/homepage-articles/templates/article.php
```diff
@@ -17,6 +17,9 @@ function( $data ) {
 		// Get sponsors for this post.
 		$sponsors = Newspack_Blocks::get_all_sponsors( $post_id );
 
+		// Get tag labels for this post.
+		$tag_labels = Newspack_Blocks::get_tag_labels( $post_id );
+
```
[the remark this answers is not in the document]
jason10lee: Addressed in 29923d53. The lookup is now gated on the block's `showTagLabels` attribute, so `get_tag_labels()` only runs when labels are set to display:

`$tag_labels = ( $attributes['showTagLabels'] ?? true ) ? Newspack_Blocks::get_tag_labels( $post_id ) : null;`

The now-redundant attribute check at the render site was dropped, leaving a single source of truth for the gate.

includes/class-newspack-blocks.php
```diff
@@ -1096,6 +1096,35 @@ public static function newspack_display_sponsors_and_categories( $sponsors ) {
...
+	}
+
+	/**
+	 * Outputs HTML for given tag labels.
+	 *
+	 * @param array $labels Labels to display.
+	 * @param bool  $links  Whether to include links to tag archives.
+	 *
+	 * @return void
+	 */
```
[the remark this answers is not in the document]
jason10lee: Addressed in c8962050 — `display_tag_labels()`'s `@param` is now `array|null` to match the `$labels = null` default.

src/blocks/homepage-articles/edit.tsx
```diff
@@ -112,6 +113,21 @@ class Edit extends Component< HomepageArticlesProps > {
...
 							) }
 						</div>
+					) }{ ' ' }
+					{ showTagLabels && post.newspack_tag_labels && (
+						<div className="cat-links tag-labels">
+							{ post.newspack_tag_labels.map( ( newspack_tag_label, index ) => {
+								return newspack_tag_label.link ? (
+									<a key={ index } href="#" className="tag-label flag">
+										{ newspack_tag_label.flag }
+									</a>
```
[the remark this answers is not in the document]
jason10lee: Keeping `href="#"` here, by convention. Editor-preview anchors in this plugin are inert placeholders: in `homepage-articles/edit.tsx` the title, read-more, image, and tag-label anchors all use `href="#"`, and `carousel/edit.js` follows the same pattern. There is no `preventDefault`-based approach anywhere in the blocks' editor code.

b0f24161 deliberately aligned the tag-label anchor to that convention, so switching just this one anchor to a real link would make it a one-off inconsistency. The preview isn't meant to be navigable.

wil-gerken (approved):
I re-checked the new commits, and all notes are handled: `generate_tag_labels()` dropped, the REST-shape tests added and passing (nice, you covered both the shape and the empty case), and the default-on toggle kept with a clear rationale. The perf gate is a great catch too. Thanks Jason!

PR #2286, opened 2026-02-02
feat(author-profile): add block theme compatibility

rbcorrales:
### All Submissions:

Adds block theme support to the Author Profile block using WordPress block bindings and nested [InnerBlocks](https://developer.wordpress.org/block-editor/how-to-guides/block-tutorial/nested-blocks-inner-blocks/). Publishers can now fully customize the author profile layout in the Site Editor by reordering, removing, or adding blocks within the template.

Main features:
- Author data (name, bio, job title, etc.) is bound to core blocks via a custom block bindings source, enabling real-time preview and full layout control in the Site Editor.
- Social media links are rendered via two new companion blocks in newspack-plugin (Author Social Links, Author Social Link) with circular icon backgrounds.
- The block uses two columns for the layout, and the Avatar column width dynamically adjusts based on the configured avatar size. This allows moving the column to the right too.
- Template structure is locked but can be reset via the toolbar button. Same with the social icons.
- Placeholder labels (`[Author]`, `[Job Title]`, etc.) display in the template editor, with real author data shown in post context.
- When using a block theme, the social icons in the previous version of the block now render SVGs via a fallback to newspack-plugin's icon map when `Newspack_SVG_Icons` (coming from the classic theme) is unavailable.

Closes NPPD-1147.

Related PRs:
- https://github.com/Automattic/newspack-plugin/pull/4448
- https://github.com/Automattic/newspack-block-theme/pull/412

Requires the companion newspack-plugin and newspack-block-theme branches.

**Test in Site Editor (template context):**
1. Activate the block theme. The block's nested mode is automatically enabled for block themes.
2. Go to Appearance > Editor > Templates > Single Post.
3. Insert a new Author Profile block.
4. Select "Contextual" mode when prompted.
5. Verify placeholder labels appear: `[Author]`, `[Job Title]`, `[Bio]`, etc.
6. Customize the layout: reorder blocks via list view, remove unwanted fields, and adjust avatar size.
7. Change avatar size and verify the avatar column width adjusts accordingly.
8. Add/remove social links via the Author Social Links block. Verify circular icon backgrounds and the icon size slider.
9. Test the reset button in the social links block toolbar.

**Test on the frontend (block theme):**
1. Create or edit a post with an author who has social links configured.
2. View the post on the frontend.
3. Verify all author data displays correctly with your customized layout.
4. Verify social icons have circular gray backgrounds with a hover effect (background changes).
5. Test with multiple co-authors.

**Test existing blocks (backwards compatibility):**
The existing Author Profile block is already in production. This PR should not disrupt it and only extend functionality for block themes.

1. Insert or find an existing Author Profile block. It should continue working exactly as before.
2. On a block theme, verify the Author Profile blocks social links (created on the classic theme) render as SVG icons with the same hover behavior.
3. Switch to newspack-theme (classic). Verify the block still renders correctly with social icons from the theme's `Newspack_SVG_Icons`.
4. Confirm that switching themes does not break or alter existing blocks.

**Test edge cases:**
1. Author with no avatar (hide default avatar enabled).
2. Author with no social links.
3. Author with no bio.
4. Post with Custom Byline active (block should show a warning and hide).

### Other information:


src/blocks/author-profile/edit.js
```diff
@@ -2,32 +2,77 @@
...
+					if ( ! key ) {
+						return [ attribute, '' ];
+					}
+					// Handle special cases.
+					if ( key === 'url' || key === 'archive_url' ) {
+						return [ attribute, author.url || '' ];
+					}
+					// "More by [author]" link text.
+					if ( key === 'archive_link_text' ) {
+						return [ attribute, author.name ? `${ __( 'More by', 'newspack-blocks' ) } ${ author.name }` : '' ];
```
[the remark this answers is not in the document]
rbcorrales: Fixed. Using `sprintf` with a translatable string so translators can reorder the placeholder.

[3fcaafae](https://github.com/Automattic/newspack-blocks/pull/2286/commits/3fcaafaefa479322b148feac78d44a63ecc5972d)

src/blocks/author-profile/view.scss
```diff
@@ -107,6 +113,63 @@
...
+				flex-grow: 1;
+			}
+		}
+
+		a {
+			color: inherit;
+
+			&:hover {
+				color: var(--newspack-ui-color-neutral-90);
+			}
```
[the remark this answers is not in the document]
rbcorrales: Added fallback `#1e1e1e` (the resolved value of `--newspack-ui-color-neutral-90` from Newspack UI's color tokens).

[e4608d50](https://github.com/Automattic/newspack-blocks/pull/2286/commits/e4608d50d957c5e47c62dd6f1dce3216fa28fc70)

src/blocks/author-profile/view.scss
```diff
@@ -107,6 +113,63 @@
...
+		}
+
+		.author-archive-link {
+			margin-top: 0.5em;
+			margin-bottom: 0.5em;
+			color: var(--newspack-ui-color-neutral-60, #6c6c6c);
+
+			&:hover {
+				color: var(--newspack-ui-color-neutral-90);
+			}
```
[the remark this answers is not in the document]
rbcorrales: Same fix applied here.

[e4608d50](https://github.com/Automattic/newspack-blocks/pull/2286/commits/e4608d50d957c5e47c62dd6f1dce3216fa28fc70)

src/blocks/author-profile/view.php
```diff
@@ -5,6 +5,68 @@
  * @package WordPress
  */
 
+/**
+ * Register block bindings source for author data.
+ *
+ * This allows core blocks to bind their content to author fields using:
+ * {"metadata":{"bindings":{"content":{"source":"newspack-blocks/author","args":{"key":"name"}}}}}
+ *
+ * Supported keys: name, bio, url, newspack_job_title, newspack_role, newspack_employer
```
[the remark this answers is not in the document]
rbcorrales: Updated the docblock to list all supported keys including `archive_url` and `archive_link_text`.

[a7cb4b20](https://github.com/Automattic/newspack-blocks/pull/2286/commits/a7cb4b20b808ae3976890b3be58d05cb245adcc4)

src/blocks/author-profile/view.php
```diff
@@ -96,35 +165,292 @@ function newspack_blocks_get_author_or_guest_author( $author_id, $avatar_size =
...
+/**
+ * Block render callback.
+ *
+ * @param array    $attributes Block attributes.
+ * @param string   $content Block inner content.
+ * @param WP_Block $block Block instance.
+ * @return string Rendered block HTML.
+ */
+function newspack_blocks_render_block_author_profile( $attributes, $content, $block ) {
+	$is_contextual  = ! empty( $attributes['isContextual'] );
```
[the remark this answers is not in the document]
rbcorrales: Removed. The variable is only needed in `newspack_blocks_get_authors_for_render()` which reads it from `$attributes` independently.

[a7cb4b20](https://github.com/Automattic/newspack-blocks/pull/2286/commits/a7cb4b20b808ae3976890b3be58d05cb245adcc4)

src/blocks/author-profile/view.php
```diff
@@ -96,35 +165,292 @@ function newspack_blocks_get_author_or_guest_author( $author_id, $avatar_size =
...
+ * @param string $html Rendered HTML from inner blocks.
+ * @param array  $author Author data array.
+ * @return string Processed HTML with archive link wrapped.
+ */
+function newspack_blocks_wrap_author_archive_link( $html, $author ) {
+	$url = $author['url'] ?? '';
+	if ( $url ) {
+		$html = preg_replace(
+			'/(<p[^>]*class="[^"]*author-archive-link[^"]*"[^>]*>)(.*?)(<\/p>)/s',
+			'$1<a href="' . esc_url( $url ) . '">$2</a>$3',
```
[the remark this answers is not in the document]
rbcorrales: Added a guard to skip wrapping when the paragraph already contains a link, avoiding nested anchors.

[a7cb4b20](https://github.com/Automattic/newspack-blocks/pull/2286/commits/a7cb4b20b808ae3976890b3be58d05cb245adcc4)

src/blocks/author-profile/edit.js
```diff
@@ -169,17 +364,107 @@ const AuthorProfile = ( { attributes, setAttributes } ) => {
...
+	// Set global author for block bindings in editor.
+	// This allows bound core blocks to access author data.
+	useEffect( () => {
+		const safeIndex = Math.min( previewAuthorIndex, Math.max( 0, authorsToRender.length - 1 ) );
+		const previewAuthor = authorsToRender[ safeIndex ] || null;
+		window.__newspackCurrentAuthor = previewAuthor;
+		return () => {
+			window.__newspackCurrentAuthor = null;
+		};
+	}, [ authorsToRender, previewAuthorIndex ] );
```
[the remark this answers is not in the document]
rbcorrales: Added a `layoutVersion !== 2` guard so the global is only set in nested mode, where bindings are actually used.

[3fcaafae](https://github.com/Automattic/newspack-blocks/pull/2286/commits/3fcaafaefa479322b148feac78d44a63ecc5972d)

src/blocks/author-profile/edit.js
```diff
@@ -2,32 +2,77 @@
...
+ * AuthorContext.Provider. This is a workaround since bindings don't have
+ * direct access to React context.
+ */
+if ( typeof registerBlockBindingsSource === 'function' ) {
+	registerBlockBindingsSource( {
+		name: 'newspack-blocks/author',
+		label: __( 'Author Profile', 'newspack-blocks' ),
+		getValues( { bindings } ) {
+			// Read author from global state set by AuthorContext.Provider.
+			const author = window.__newspackCurrentAuthor || {};
```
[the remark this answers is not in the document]
rbcorrales: The block bindings API doesn't expose `context.rootClientId` or `context.clientId` in `getValues`, so scoping per instance isn't feasible with the current API. In practice, the Author Profile only renders one author at a time (the preview author), and template contexts typically have a single instance. This is a known architectural constraint of the bindings API.

src/blocks/author-profile/edit.js
```diff
@@ -305,52 +646,277 @@ const AuthorProfile = ( { attributes, setAttributes } ) => {
...
+		const previewAuthor = authorsToRender[ safeIndex ];
+
+		// Set global synchronously so bindings have access on first render.
+		// The useEffect handles cleanup when component unmounts.
+		window.__newspackCurrentAuthor = previewAuthor;
+
+		const nestedBlockProps = {
+			...blockProps,
+			className: `${ blockProps.className } wp-block-newspack-blocks-author-profile is-nested-mode`,
+		};
```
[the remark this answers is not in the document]
rbcorrales: This is intentional. Blocks in this repo use a non-standard registration pattern: `block.json` has `"name": "author-profile"` (no namespace), and the full name is constructed at registration time via `registerBlockType('newspack-blocks/' + name, settings)`. Because of this, `useBlockProps()` doesn't generate the `wp-block-newspack-blocks-author-profile` class automatically. This is a repo-wide pattern; all blocks in this repo rely on manual class handling (PHP side uses `Newspack_Blocks::block_classes()`). Fixing this would require updating all blocks to use standard `block.json` with namespaced names, which is out of scope here.

src/blocks/author-profile/editor.scss
```diff
@@ -1,7 +1,41 @@
-@use "view";
 @use "../../shared/sass/mixins";
```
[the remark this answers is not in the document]
rbcorrales: The editor still gets the `view.scss` styles via the JS import chain: `editor.js` imports `index.js`, which imports both `./editor.scss` and `./view.scss`. Both get bundled into `dist/editor.css` by webpack. The `@use "view"` was an unnecessary duplicate that was removed intentionally.

src/blocks/author-profile/edit.js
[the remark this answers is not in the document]
rbcorrales: Correction: after testing with two specific-author blocks on the same page, the issue is reproducible when reordering blocks. The global gets overwritten and child bindings read stale data.

`getValues` does receive `clientId` (of the bound child block) and `select`, so per-instance scoping is feasible by walking up the block tree with `getBlockParents()`.

Fixed by replacing the single `window.__newspackCurrentAuthor` global with a per-instance map (`window.__newspackAuthorsByBlock`) keyed by the parent Author Profile's `clientId`. Each block registers its author on mount and cleans up on unmount, and `getValues` looks up the correct author via `getBlockParents()`.

[4f46c142](https://github.com/Automattic/newspack-blocks/pull/2286/commits/4f46c14257f3e0ac80c25d8b6cb3407bdfb9da14)

src/blocks/author-profile/class-wp-rest-newspack-authors-controller.php
```diff
@@ -444,4 +461,167 @@ function( $acc, $profile ) use ( $author_id ) {
...
+				$avatar = coauthors_get_avatar( $guest_author, 256 );
+				if ( $avatar && ( false === strpos( $avatar, 'avatar-default' ) || ! $avatar_hide_default ) ) {
+					$author_data['avatar'] = $avatar;
+				}
+			}
+
+			return self::fill_guest_author_data( $author_data, $guest_author, $fields );
+		}
+
+		// It's a linked WP user.
```
laurelfulford: When I set up a Guest Author and link them to an existing WordPress user, the existing WordPress user doesn’t show up in the dropdown (expected since the Guest Author is supposed to kind of ‘mask’ who they are, but I need to double-check how this works in trunk). 

However, when I pick that Guest Author, no information shows for them in the block — the image used for the avatar falls back to the post/page author’s, and there’s just placeholders even though this author has a display name, avatar, and bio info.

Image

Image

This seems to work as expected for non-linked Guest authors.

src/blocks/author-profile/class-wp-rest-newspack-authors-controller.php
```diff
@@ -79,9 +82,15 @@ public function get_authors( $request ) {
 		$search              = ! empty( $request->get_param( 'search' ) ) ? $request->get_param( 'search' ) : null; // Fetch authors by search string.
 		$offset              = ! empty( $request->get_param( 'offset' ) ) ? $request->get_param( 'offset' ) : 0; // Offset results (for pagination).
 		$per_page            = ! empty( $request->get_param( 'perPage' ) ) ? $request->get_param( 'perPage' ) : 10; // Number of results to return per page. This is applied to each query, so the actual number of results returned may be up to 2x this number.
```
laurelfulford: I know this isn't part of this PR, but should this be per_page? 😅

src/blocks/author-profile/class-wp-rest-newspack-authors-controller.php
```diff
@@ -444,4 +461,167 @@ function( $acc, $profile ) use ( $author_id ) {
...
+					}
+				}
+
+				$response = new WP_REST_Response( $authors );
+				$response->header( 'x-wp-total', count( $authors ) );
+				return rest_ensure_response( $response );
+			}
+		}
+
+		// Fallback to default post author.
```
laurelfulford, line 491: If Co-Authors is active but doesn't return any $coauthors to a post for some reason, I don't think we want to fall back to the default post author (like the WordPress user who published the story). Say an editor is publishing a story for a guest author, something gets messed up in the data, and the post somehow has no author attached -- the Editor's bio would be the one that ends up showing on the front-end. 

I know it's a weird edge case, but I think it might be 'safer' to return nothing -- it still wouldn't be right exactly, but it would be missing information rather than showing incorrect information. What do you think?

src/blocks/author-profile/view.php
```diff
@@ -96,35 +165,297 @@ function newspack_blocks_get_author_or_guest_author( $author_id, $avatar_size =
...
+/**
+ * Render author profiles using flat template (existing behavior).
+ *
+ * @param array $authors Array of author data.
+ * @param array $attributes Block attributes.
+ * @return string Rendered HTML.
+ */
+function newspack_blocks_render_flat_author_profiles( $authors, $attributes ) {
+	$output         = '';
+	$show_empty_bio = $attributes['showEmptyBio'] ?? false;
```
laurelfulford, line 409: The classic theme does this for the end-of-story bios (skips authors that don't have the author bio text) but this would be a breaking change for the Author Profile block, which will show authors regardless of what information's available.

What do you think about making this an option so you can opt into skipping bio-less profiles? Or even just remove that logic for now, and come back to it if publishers ask for it?

As is, existing Author Profile blocks that show authors without bios disappear after this update in the Classic theme; they also still preview in the editor with both themes.

src/blocks/author-profile/view.scss
```diff
@@ -107,6 +113,63 @@
...
+				flex-basis: auto;
+				flex-grow: 1;
+			}
+		}
+
+		a {
+			color: inherit;
+
+			&:hover {
+				color: var(--newspack-ui-color-neutral-90, #1e1e1e);
```
laurelfulford: Should we be pulling something from the theme for this (like the `--wp--preset--color--base-#` colours), or letting this (and the other link) just pick up the default link styles from the theme/Global Settings?

src/blocks/author-profile/view.php
```diff
@@ -96,35 +165,297 @@ function newspack_blocks_get_author_or_guest_author( $author_id, $avatar_size =
...
+	$url = $author['url'] ?? '';
+	if ( $url ) {
+		$html = preg_replace_callback(
+			'/(<p[^>]*class="[^"]*author-archive-link[^"]*"[^>]*>)(.*?)(<\/p>)/s',
+			function ( $matches ) use ( $url ) {
+				// Avoid creating nested anchors if the paragraph already contains a link.
+				if ( false !== stripos( $matches[2], '<a ' ) ) {
+					return $matches[0];
+				}
+				return $matches[1] . '<a href="' . esc_url( $url ) . '">' . $matches[2] . '</a>' . $matches[3];
```
laurelfulford: I know we know where `$matches[2]` comes from and what it is, but should it be escaped here just for completeness/escaping everything all the time?

src/blocks/author-profile/class-wp-rest-newspack-authors-controller.php
```diff
@@ -444,4 +461,167 @@ function( $acc, $profile ) use ( $author_id ) {
...
+		}
+
+		if ( ! $user ) {
+			return null;
+		}
+
+		// Check if this user has a linked guest author for additional data.
+		$linked_guest_author = self::get_linked_guest_author( $user->user_login );
+
+		if ( $linked_guest_author && class_exists( 'CoAuthors_Guest_Authors' ) ) {
```
laurelfulford, line 579: When I set up a Guest Author and link them to an existing WordPress user, the existing WordPress user doesn’t show up in the dropdown (expected since the Guest Author is supposed to kind of ‘mask’ who they are). 

However, when I pick that Guest Author, no information shows for them in the block — the image used for the avatar falls back to the post/page author’s, and there’s just placeholders even though this author has a display name, avatar, and bio info.

Image

Image

This seems to work as expected for non-linked Guest authors.

src/blocks/author-profile/view.php
```diff
@@ -5,6 +5,68 @@
  * @package WordPress
  */
 
+/**
+ * Register block bindings source for author data.
```
laurelfulford, line 9: In the post/page editor, the block bindings seem to show some kind of placeholder text if empty (the title and job description), or nothing (the employer and description). 

When they do show placeholders, it's always "Author Profile" - not sure if it's possible but it would be clearer what's happening if these could say what the fields are.

When the fields don't show anything, they seem like they're just empty space until you click on the block:

Image

The placeholders look good in the Site Editor!

src/blocks/author-profile/class-wp-rest-newspack-authors-controller.php
```diff
@@ -302,6 +311,8 @@ public static function fill_guest_author_data( $guest_author_data, $guest_author
 					];
 					if ( class_exists( 'Newspack_SVG_Icons' ) ) {
 						$value['svg'] = Newspack_SVG_Icons::get_social_link_svg( $value['url'], 24 );
+					} elseif ( class_exists( '\Newspack\Blocks\Author_Profile_Social\Author_Profile_Social_Block' ) ) {
```
laurelfulford: I get why we're doing it (because the old Author Profile block will need SVGs even after moving to the Block theme), but what do you think about making the Newspack Plugin the one central source?

It does mean we'll end up with the SVGs duplicated across the Classic theme and Newspack Plugin, but we'll have to do that anyway for the block theme. That at least knocks out needing to check for the Classic Theme.

src/blocks/author-profile/index.js
```diff
@@ -55,5 +56,12 @@ export const settings = {
 		default: '',
 	},
 	edit,
-	save: () => null, // to use view.php
+	// Save inner blocks for nested mode (layoutVersion 2).
+	// For flat mode (layoutVersion 1), return null to use server-side rendering only.
+	save: props => {
```
laurelfulford, line 58: This is an edge case, but when you go from the Block theme to the Classic theme, you end up getting errors about blocks like Avatar not being supported.

I'm not sure what the best fix would be but wanted to flag in case it's easy-ish to fix! This is probably something we can log for the future vs. worrying about now.

laurelfulford (changes requested):
This is looking really cool, @rbcorrales! I found some odds and ends, and hopefully noted them on the correct PR. Just let me know if any of this doesn't make any sense!

Edited to add: I'm still working through the newspack-plugin PR but should have something to share tomorrow! 

----

~~Overall, the block tested well following the steps -- I wasn't sure exactly how to test "Author with no avatar (hide default avatar enabled)." I feel like I'm missing something silly, but I didn't see an option to hide the default avatar.~~ D'oh, just figured it out: it's an option in the original block 🤦‍♀️ It doesn't look like it's working in the "classic" version of the block when I added it to the editor on this branch.

For the new version of the block, it ignores whatever you have set as the placeholder in the editor and uses the big G Gravatar icon:

Image

Also for this step:

> Post with Custom Byline active (block should show a warning and hide).

This works like this for me whether or not the Custom Byline includes an actual Author shortcode.

For consistency with how the byline/avatar works, I think it'd be good if the profile still showed if the author is "real". I think that goes beyond what was in the scope (since this how an author bio would work in this situation in the Classic theme!) so it's definitely not a blocker.  

----

Given that the separate block version of the plugin relies on blocks in the Newspack Plugin to work (like avatar and social links), should we make the Newspack Plugin required before the new version of the block is used?

----

Since the block lets you do all the layouts with blocks (🎉 works so well!!), the layouts can probably be removed from the block theme version:
Image

----

For consistency, it would be nice to match the loading state of the OG block: 

Image

----

I'm not sure the best way to fix this given that we want to reuse the Author Profile block for the block theme, but having the different pieces spread across this plugin and Newspack Plugin feels like it'll get confusing, fast 😅 I'm not sure what the best fix would be -- maybe moving this block and the Author List block to the Newspack Plugin?

src/blocks/author-profile/view.php
rbcorrales: Good idea! I changed the default to true so existing blocks keep working. Also added a "Show authors without bio" toggle in the sidebar for the block theme version (v2), so publishers can opt into hiding bio-less profiles if they want.

The toggle also filters the editor preview to match frontend behavior. The classic/v1 path keeps the same default (true), but no toggle for now. We can add one later if needed.

[4a99e26](https://github.com/Automattic/newspack-blocks/pull/2286/commits/4a99e264e610ead2235bf1374f980a342bded20a)

src/blocks/author-profile/class-wp-rest-newspack-authors-controller.php
rbcorrales: Done in[ 5ed084d](https://github.com/Automattic/newspack-blocks/pull/2286/commits/5ed084d2b8ccac6473d94c6fba352b5b8e2c1947). The REST controller now checks the plugin first and falls back to the theme. See the [related discussion](https://github.com/Automattic/newspack-plugin/pull/4448#discussion_r2819793347) for additional context.

src/blocks/author-profile/index.js
rbcorrales: Fixed in [7b8eddb](https://github.com/Automattic/newspack-blocks/pull/2286/commits/7b8eddb6e5066056714af71973c4ed006eed4675). The inner blocks (Avatar, Social Links) are only registered for block themes, so they can't render properly in a classic theme context. The editor now detects when a v2 block (created with a block theme) is opened in a classic theme and shows a warning notice instead of trying to render the inner blocks:

image

On the frontend, `view.php` checks `wp_is_block_theme()` and falls back to the flat/classic render path automatically.

The `block.json` defaults for the flat layout attributes (`showAvatar`, `showBio`, `showSocial`, etc.) are all reasonable, so the fallback should produce a good result even though those settings were never explicitly configured.

src/blocks/author-profile/view.php
rbcorrales: Fixed in [511632a](https://github.com/Automattic/newspack-blocks/pull/2286/commits/511632a5c05cf855dc5336ebe3565218dc34719d). The "Author Profile" text was coming from the binding source's label, which WordPress uses as a generic placeholder when a bound value is empty. Now, empty fields fall back to the same field-specific labels used in the Site Editor (`[Job Title]`, `[Bio]`, etc.) via `getPlaceholderAuthor()`.

src/blocks/author-profile/view.scss
rbcorrales: Updated the nested mode styles to use `--wp--preset--color--contrast` for link hover and `--wp--preset--color--contrast-2` for the archive link, with hex fallbacks from the block theme's `theme.json`. This matches the flat mode, which was already using contrast-2 correctly.

[2a1ff48](https://github.com/Automattic/newspack-blocks/pull/2286/commits/2a1ff48368df8eea9b88c427555669c28411fe0e)

src/blocks/author-profile/class-wp-rest-newspack-authors-controller.php
rbcorrales: Well, changed again 😄 
Now it only uses the plugin's icons map. [Context](https://github.com/Automattic/newspack-plugin/pull/4448#discussion_r2820317415).

src/blocks/author-profile/view.php
rbcorrales: Good call. The content comes from `WP_Block::render()` which already sanitizes bound values via `wp_kses_post()`, so it's technically safe as-is. But following the "escape as late as possible" principle, added `wp_kses_post()` here (`esc_html()` would break any legitimate inline HTML in the paragraph content). 
[8890008](https://github.com/Automattic/newspack-blocks/pull/2286/commits/8890008b74b16c8fbd559771cf4666f8701a9499)

src/blocks/author-profile/class-wp-rest-newspack-authors-controller.php
rbcorrales: When Co-Authors Plus is active, the code now always returns from the CAP branch and never falls through to `$post->post_author`. If `get_coauthors()` somehow returns empty, we show nothing rather than potentially the wrong author: [a0eac0e](https://github.com/Automattic/newspack-blocks/pull/2286/commits/a0eac0eb05ff610c079c89e3400ce8bd7532fd95).

Currently, `get_coauthors()` falls back to the post author when no CAP terms are found, unless `force_guest_authors` is enabled, in which case it can return empty. Either way, a good defensive measure since we shouldn't rely on that internal behavior.

rbcorrales:
@laurelfulford addressing some of the top-level comments from your review here:

> D'oh, just figured it out: it's an option in the original block 🤦‍♀️ It doesn't look like it's working in the "classic" version of the block when I added it to the editor on this branch.

I tested this a bit more on a separate staging site, and I think the "hide default avatar" toggle is broken on trunk too, not a regression from this PR.

The toggle relies on detecting the `avatar-default` CSS class in the avatar HTML (`strpos( $avatar, 'avatar-default' )`). However, WordPress only adds that class when `get_avatar_data()` returns` found_avatar = false`, which essentially never happens because any valid email produces a Gravatar hash and WordPress considers it "found" regardless of whether a real Gravatar image exists.

Tested on trunk with both a guest author and a freshly created WP user without a profile picture. The avatar HTML returns `class='avatar avatar-128 photo'` with no `avatar-default` class. The check always passes, and the toggle has no effect.

Could you confirm whether this is working as expected on trunk for you? It's also possible that Gravatar changed its behavior at some point.

Now, for fixing this, the reliable way to detect a default avatar differs by author type:

- Guest authors (CAP): Check `has_post_thumbnail()` on the guest author post. CAP stores uploaded avatars as featured images, so this is cheap and reliable.
- WP users: The only reliable method is a HEAD request to `gravatar.com/avatar/{hash}?d=404` and checking for a 404 response. This could be cached in a transient to avoid repeated requests, but it does add an external HTTP call on first
  load per author.

Options:

1. Fix guest authors only (no HTTP, partial fix).
2. Fix both with a cached Gravatar HEAD check (full fix, adds one HTTP call per author cached for ~1 week).
3. Leave as a pre-existing bug and track separately.

What do you think? Or am I completely misunderstanding the issue here? 😄 

---

> For consistency with how the byline/avatar works, I think it'd be good if the profile still showed if the author is "real". I think that goes beyond what was in the scope (since this how an author bio would work in this situation in the Classic theme!) so it's definitely not a blocker.

Addressed in [3a00690](https://github.com/Automattic/newspack-blocks/pull/2286/commits/3a00690d5cfcf7e35bba6d1faeaaaae4c0ff2428). When Custom Byline is active, the block now parses the `_newspack_byline` meta for `[Author id=N]` shortcodes and shows profiles for those referenced WordPress users. If the byline is pure text with no author references, it still hides the block and shows the "Custom Byline is active" warning.

PHP side reuses `Newspack\Bylines::get_post_byline_authors()` to resolve the referenced users. If the Bylines class is unavailable (plugin deactivated), it returns empty rather than falling back to post authors, since the byline is the source of truth.

The JS side currently duplicates the shortcode regex from `extractAuthorIdsFromByline()` in `newspack-plugin/src/shared/hooks/use-custom-byline.js` since we can't import across plugin boundaries. Once the Author Profile block moves to `newspack-plugin`, it can reuse the shared `useCustomByline()` and `extractAuthorIdsFromByline()` hooks that the avatar and byline blocks already use.

---

> Given that the separate block version of the plugin relies on blocks in the Newspack Plugin to work (like avatar and social links), should we make the Newspack Plugin required before the new version of the block is used?

Yes, nested mode now requires the Newspack Plugin. But instead of checking for the plugin itself, both PHP and JS now check that the `newspack/avatar` and `newspack/author-profile-social` block dependencies are registered before enabling nested mode: [2e7ff05](https://github.com/Automattic/newspack-blocks/pull/2286/commits/2e7ff05ddf4819867dfb45b6e9175ed6413fbbb6).

If either block is missing, the block automatically falls back to flat rendering in the frontend and the same message from [here](https://github.com/Automattic/newspack-blocks/pull/2286#discussion_r2819914060) in the editor.

---

> Since the block lets you do all the layouts with blocks (🎉 works so well!!), the layouts can probably be removed from the block theme version:

Ok, I looked into this. For the block styles (Default/Centered in the toolbar), WordPress registers these globally per block type, not per instance. The approaches I explored:

1. PHP conditional with `wp_is_block_theme()`: Cleanly avoids registering styles on block themes, but prevents editing styles on v1 blocks that were created before switching to a block theme.
2. `useEffect` + `unregisterBlockStyle`: Per-instance aware, but `unregisterBlockStyle` is global. If a page has both a v1 and v2 block (highly unlikely I guess), the v2 block would remove the style picker from the v1 block too.
3. Injected `<style>` tag or body class: Inject CSS to hide the Styles panel when a nested block is selected. Works per-instance but is a DOM hack tied to WordPress editor internals.
4. Custom `InspectorControls` toggle: Remove core block styles entirely, reimplement the style selector as a custom toggle rendered only for v1 blocks. Clean per-instance control, but loses the core styles preview on hover.

There's no WordPress API for per-instance block style visibility. The styles have no visual effect on nested mode markup, so the other option is to just leave them registered as is. The worst case is a publisher clicks "Centered" on a nested block and nothing visible changes.

What would you prefer? If we accept that we might not always be able to update the styles of legacy author blocks in the block theme, then options 2 or 1 are viable.

---

> For consistency, it would be nice to match the loading state of the OG block:

Fixed in [1cc566c](https://github.com/Automattic/newspack-blocks/pull/2286/commits/1cc566cbb3e78fed4b5ca6eb353afdbd0076f55b). Both loading states (nested and flat mode) now use a shared `<Placeholder>` component with the block icon and "Fetching author info..." text, matching the original block.

---

>  I'm not sure the best way to fix this given that we want to reuse the Author Profile block for the block theme, but having the different pieces spread across this plugin and Newspack Plugin feels like it'll get confusing, fast 😅 I'm not sure what the best fix would be -- maybe moving this block and the Author List block to the Newspack Plugin?

Totally agree, it's not ideal having pieces spread across both repos. The plan is to move the Author Profile block (and likely Author List) to `newspack-plugin` eventually. I initially added the social blocks here in `newspack-blocks`, but since they're new components, it made more sense to start them in the right place, so I moved them to the plugin repo [559d256](https://github.com/Automattic/newspack-blocks/pull/2286/commits/559d256ed038e253ba9cc75c5c31f86878854e29). I'm aware of the awkward cross-repo dependency this creates in the meantime, but I'd like to keep that refactor out of scope for these PRs, though. There's already plenty to review here!

laurelfulford:
Thanks @rbcorrales! I started by reviewing the big comment reply; I'll need to circle back tomorrow for the rest (and I'll try not to cross paths with your work/make too much of a mess of the feedback on this one 😆):

> I think the "hide default avatar" toggle is broken on trunk too... Could you confirm whether this is working as expected on trunk for you? It's also possible that Gravatar changed its behavior at some point.

Oh good catch -- I get the same thing on trunk, too! It seems to work for Author List but not Author Profile -- seeing that, I wonder if it's just a weird UI thing (like the Author List is meant to have the toggle, but the Author Profile is not). For the Author Profile block in it's original form, you're picking specific authors and you can hide the avatar in general for the block, so having a case-specific override seems unnecessary. The more general approach makes more sense for the Author List block since you can show a ton of authors and they all share the same settings. 

I made an issue for this here: NPPM-2626

> Now, for fixing this...

I think it would be nice to have this work in the updated block, but assuming I'm right about the toggle being an accident, it wouldn't be needed for feature parity and could be added later. It's definitely come up for the theme (in bylines and the author bio) but we've never addressed it there 😅 

> When Custom Byline is active, the block now parses the _newspack_byline meta for [Author id=N] shortcodes and shows profiles for those referenced WordPress users. If the byline is pure text with no author references, it still hides the block and shows the "Custom Byline is active" warning.

This works great! I tested the custom byline with an author shortcode and without.

> Yes, nested mode now requires the Newspack Plugin. But instead of checking for the plugin itself, both PHP and JS now check that the newspack/avatar and newspack/author-profile-social block dependencies are registered before enabling nested mode.

This looks good - thank you!

> Ok, I looked into this. For the block styles (Default/Centered in the toolbar), WordPress registers these globally per block type, not per instance... What would you prefer? If we accept that we might not always be able to update the styles of legacy author blocks in the block theme, then options 2 or 1 are viable.

Ok, wow, that was a lot more complicated than I expected! 

I think #1 is probably enough:

* If the content already exists, it's unlikely a publisher will want to edit it and and only change the layout to centred
* Worst case, the CSS class that the block style uses can still be added
* Also, the Author Profile block is mostly used on staff pages on existing sites; I think if publishers wanted to overhaul those as part of moving to the block theme they'd probably be more interested in checking out the new version of the block than changing the old one 🙂 

> Both loading states (nested and flat mode) now use a shared <Placeholder> component with the block icon and "Fetching author info..." text, matching the original block.

I noticed this when doing additional testing of the other PR -- it looks nice, thank you!

> The plan is to move the Author Profile block (and likely Author List) to newspack-plugin eventually...

Eventually is good! Definitely some point after the dust settles on all the other moving pieces 😅

src/blocks/author-profile/class-wp-rest-newspack-authors-controller.php
laurelfulford: Thanks Ramon! This looks good to me!

src/blocks/author-profile/view.php
laurelfulford: This looks good! I tested in with the Classic theme (no option, block shows without bio), and tested it with and without the option in in the Block theme for authors with and without bios. All works as described, and the original block works the same as before 🙂

src/blocks/author-profile/view.scss
laurelfulford: Thanks!

src/blocks/author-profile/view.php
laurelfulford: Thanks for humoring all the sanitizing I got drilled into me on the theme team 😆 Looks good!

src/blocks/author-profile/view.php
laurelfulford: Nice! I like the consistency with the placeholders used in the Site Editor. The only suggestion I have is maybe greying them out still so they still look like placeholders (and not content that will be published) but that's minor! (To me that seemed like less of an issue in the Site Editor where it's more clear pretty much everything is a placeholder!).

src/blocks/author-profile/class-wp-rest-newspack-authors-controller.php
laurelfulford: Thanks Ramon! This looked good in my testing in the Newspack Plugin, I'm good with the approach, so marking this one as done!

src/blocks/author-profile/index.js
laurelfulford: I ran into this almost by accident testing other stuff, and it feels like a good way to handle it: it still works and looks the same as it would normally in the classic theme if you happen to do the weird block theme > classic theme flow, and the fallback in the editor is consistent with the loading state visually, and adds some nice context why it's not working as expected! 🙂

src/blocks/author-profile/class-wp-rest-newspack-authors-controller.php
laurelfulford: I retested this one with everything else and it's working as expected now! Not sure if it got a purposeful fix, an accidental fix, or if I was running into an issue locally the first time that fixed itself but I'll take it!

laurelfulford:
@rbcorrales I think I followed up on everything! TL;DR: everything looks good, I think the outstanding bits are just whatever solution needs to be implemented for the classic block's layout option (I'm find with the easiest approach), and I had a suggestion for greying out the placeholders but that's not a blocker. 

Just let me know if you have any questions about anything!

src/blocks/author-profile/view.php
rbcorrales: Done! Each bound block in the nested template now has a placeholder attribute (e.g. [Job Title], [Bio]) that WordPress renders in its native greyed-out style when the field has no data. This applies in both the Site Editor (template context, all fields) and the post editor (only fields the author hasn't filled in).

image

The placeholder text is derived from the block's list-view name, wrapped in brackets, so there's a single source of truth. Blocks created before this change will show the generic "Author Profile" label from WordPress core. Resetting the layout via the toolbar button or re-adding them should pick up the new placeholders.

[16a474b](https://github.com/Automattic/newspack-blocks/pull/2286/commits/16a474b099b10925bc84ce6dc928e8e411533a6a)

rbcorrales:
Thanks for the thorough review @laurelfulford!

> I think # 1 is probably enough:

Block styles are now registered in PHP via `register_block_style()` instead of JS, wrapped in `! wp_is_block_theme()` so the style picker only appears on classic themes: [ada8703](https://github.com/Automattic/newspack-blocks/pull/2286/commits/ada8703112e9c2dc03234c0561cc8cdd9fe92ef4).

As for the greyed-out placeholders, I implemented that and [commented](https://github.com/Automattic/newspack-blocks/pull/2286#discussion_r2831245381) about it.

Let me know if I'm still missing anything!

thomasguillot:
I made some changes to the nested mode so we're using block settings rather than forced css (note: you need to add a new block to the page to see these changes)

cc @rbcorrales @laurelfulford

thomasguillot:
I removed the "Author source" panel. I didn't understand its purpose. It's simple enough to just delete the block and add a new one.

rbcorrales:
@thomasguillot Thanks for the style updates. I made a few changes after your commits:

- 053b085d fixes a latent bug in `createBoundParagraph`. The new `...extraAttributes` rest parameter collected arguments into an array, so spreading it into the attributes object would have produced numeric keys (`{ 0: value }`) instead of merging properties. No callers ever passed extra attributes, so the bug never fired, but the signature was misleading. Removed the parameter entirely since it's unused ([YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)).
- 95b84ea7 removes dead `.components-button-group` CSS from `editor.scss`. You removed the `ButtonGroup` component in d5d61f2a, but its styles were left behind.
- a7d753a2 normalizes `__next40pxDefaultSize={ true }` to the bare prop form for consistency with the rest of the file.

Also, in 3c305f61, the archive link placeholder was changed from 'More by Author' to 'More by Author name'. This string is translatable, so embedding "Author name" as a literal can confuse translators into translating it as regular words. Could we use 'More by [Author]' instead? The brackets signal that it's a placeholder value, and it would also match the pattern already used for bylines in the Site Editor ('By [Author]').

thomasguillot:
@rbcorrales I updated the strings here: https://github.com/Automattic/newspack-blocks/pull/2286/commits/363d6412b9927f1fa0fa7740ba8d44ae7a9defb4 so the strings are more consistent with other Core blocks.

I’m not keen on using something like `[Author]`. From a developer perspective it makes sense, but most people won’t understand why there are brackets there. Core doesn’t handle it that way, and neither should we.

Screenshot 2026-02-21 at 08 52 56@2x

We need to update this to "By Author Name".

src/blocks/author-profile/view.php
laurelfulford: I know this is a bit in motion still (removing brackets) but this is looking good to me!

src/blocks/author-profile/view.scss
```diff
@@ -2,34 +2,58 @@
...
-	}
+		h3 {
+			margin-bottom: 0.5em;
+		}
 
-	h3 {
-		margin-bottom: 0.5em;
-	}
+		p {
+			font-size: var(--wp--preset--font-size--small, 1em);
```
laurelfulford: @thomasguillot this looks like it's affecting the sizing of some of the paragraphs in the classic version of the block (so it will change how this block looks on live sites). I noticed it with the Job Title and Bio, they're normally 20px but will be switched to 16px.

Trunk is on the left, this PR on the right:

Image

laurelfulford:
@rbcorrales The styling change (removing the Default and Centred buttons in the block theme) looks good!

@thomasguillot I noted one font sizing issue above - I think that's kind of a blocker (we don't want to change how the block looks on live sites), but once that's fixed I think the block is pretty solid! 

What do you both think about finding a hard stop point with this PR, and moving additional feedback to Linear, just so we're working in smaller, more contained PRs? 🙂 If that sounds good I can rubber stamp this one!

src/blocks/author-profile/view.scss
thomasguillot: My bad! Should be fixed? https://github.com/Automattic/newspack-blocks/pull/2286/commits/132120b74e48143dc101c8046c17ba9cd4bfd459

rbcorrales:
> What do you both think about finding a hard stop point with this PR

@laurelfulford I agree! But I think Thomas already fixed the last styling issue you flagged.
Or is there anything else pending for this PR?

rbcorrales:
Created this Linear issue for updating the Byline block placeholder: NPPD-1268.

src/blocks/author-profile/view.scss
laurelfulford: Thanks -- looks good!

laurelfulford:
> Or is there anything else pending for this PR?

That was it 🙂 I just got distracted by other stuff -- marking as approved!

laurelfulford (approved):

PR #2304, opened 2026-03-03
fix(modal-checkout): suppress WooPayments express checkout when Stripe is active

wil-gerken:
### All Submissions:
* [x] Have you followed the Newspack Contributing guideline?
* [x] Does your code follow the WordPress' coding standards and VIP Go coding standards?
* [x] Have you checked to ensure there aren't other open Pull Requests for the same update/change?

When both Stripe and WooPayments are active with express checkout enabled, the modal checkout renders two sets of Apple Pay / Google Pay buttons — one from each gateway.

The fix adds a new `unhook_woocommerce_payments_express_checkout_buttons()` method that runs on `init` priority 20 (after WooPayments attaches its callback at priority 15) and removes WooPayments' `display_express_checkout_buttons` action from `woocommerce_checkout_before_customer_details` during modal checkout. This mirrors the existing `unhook_woocommerce_payments_update_billing_fields()` pattern in the same file.

The unhook is gated on:
1. Being in modal checkout context (`is_modal_checkout()`).
2. WooPayments being installed (`class_exists('WC_Payments')`).
3. Stripe's gateway being enabled AND its express checkout setting being `yes`.

If Stripe isn't providing express checkout, WooPayments remains the only source of Apple Pay / Google Pay in the modal and renders normally.

Also removes now-unused filter code inside `filter_woocommerce_payments_settings()` that targeted `express_checkout_checkout_methods` — WooPayments 10.4+ no longer consults that key for button display (it reads from the new split Apple Pay / Google Pay gateway options instead), so the filter was a silent no-op in modern installs.

Closes NPPM-2652.

**Reproduce the problem (without this PR):**

1. Activate both `woocommerce-gateway-stripe` and `woocommerce-payments`, with express checkout enabled on each.
2. WooPayments must be in a connected state (`WC_Payments_Features::are_payments_enabled()` returns `true`) — if the gateway hasn't been connected, its display hook won't fire.
3. Open the modal checkout via the Donate block (e.g. on `/support-our-publication/`).
4. You will see two sets of Apple Pay / Google Pay buttons

download-1

**Verify the fix (with this PR):**

5. Reload the page. Confirm only a single row of Apple Pay / Google Pay (Stripe's, with the "Pay with Link" button) appears, followed by one "— Or —" divider and the card form.
   
download

**Regression Check:**

6. Deactivate Stripe, leaving only WooPayments active. Reload the page and open the modal. Confirm WooPayments' Apple Pay / Google Pay buttons appear normally. The fix should only suppress when both gateways are providing express checkout.

### Other information:

### DIFF
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -530,8 +530,24 @@ final class Modal_Checkout {
 		if ( ! self::is_modal_checkout() ) {
 			return $settings;
 		}
+		// Disable WooPay (Woo's hosted checkout experience) — not supported in the modal.
 		if ( isset( $settings['platform_checkout'] ) ) {
 			$settings['platform_checkout'] = 'no';
+		}
+		// If Stripe's express checkout is active, suppress WooPayments' express checkout
+		// (payment_request) to prevent duplicate Apple Pay / Google Pay buttons in the modal.
+		// Stripe is preferred: Newspack already manages its express checkout behavior (see
+		// class-woocommerce-gateway-stripe.php), and WooPayments is already partially filtered
+		// above (platform_checkout). WooPayments is suppressed, not the reverse.
+		if ( class_exists( 'WC_Stripe' ) ) {
+			$stripe_settings = get_option( 'woocommerce_stripe_settings', [] );
+			if ( isset( $stripe_settings['express_checkout'] ) && 'yes' === $stripe_settings['express_checkout'] ) {
+				if ( isset( $settings['express_checkout_checkout_methods'] ) ) {
+					$settings['express_checkout_checkout_methods'] = array_values(
+						array_diff( $settings['express_checkout_checkout_methods'], [ 'payment_request' ] )
+					);
+				}
+			}
 		}
 		return $settings;
 	}
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -593,12 +593,12 @@ final class Modal_Checkout {
 		if ( ! isset( $GLOBALS['wp_filter']['woocommerce_checkout_before_customer_details'] ) ) {
 			return;
 		}
-		$filters = $GLOBALS['wp_filter']['woocommerce_checkout_before_customer_details'];
-		foreach ( $filters as $index => $filter ) {
-			$keys = array_keys( $filter );
-			foreach ( $keys as $key ) {
+		$hook      = $GLOBALS['wp_filter']['woocommerce_checkout_before_customer_details'];
+		$callbacks = $hook instanceof \WP_Hook ? $hook->callbacks : (array) $hook;
+		foreach ( $callbacks as $priority => $priority_callbacks ) {
+			foreach ( array_keys( $priority_callbacks ) as $key ) {
 				if ( strpos( $key, 'display_express_checkout_buttons' ) !== false ) {
-					remove_action( 'woocommerce_checkout_before_customer_details', $key, $index );
+					remove_action( 'woocommerce_checkout_before_customer_details', $key, $priority );
 				}
 			}
 		}
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -544,7 +544,7 @@ final class Modal_Checkout {
 			if ( isset( $stripe_settings['express_checkout'] ) && 'yes' === $stripe_settings['express_checkout'] ) {
 				if ( isset( $settings['express_checkout_checkout_methods'] ) ) {
 					$settings['express_checkout_checkout_methods'] = array_values(
-						array_diff( $settings['express_checkout_checkout_methods'], [ 'payment_request' ] )
+						array_diff( (array) $settings['express_checkout_checkout_methods'], [ 'payment_request' ] )
 					);
 				}
 			}
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -536,9 +536,9 @@ final class Modal_Checkout {
 		}
 		// If Stripe's express checkout is active, suppress WooPayments' express checkout
 		// (payment_request) to prevent duplicate Apple Pay / Google Pay buttons in the modal.
-		// Stripe is preferred: Newspack already manages its express checkout behavior (see
-		// newspack-plugin/includes/plugins/class-woocommerce-gateway-stripe.php), and WooPayments
-		// is already partially filtered above (platform_checkout). WooPayments is suppressed, not the reverse.
+		// Stripe is preferred here because Newspack already keys this behavior off the Stripe
+		// gateway settings below, while WooPayments is already partially filtered above via
+		// platform_checkout. WooPayments is suppressed, not the reverse.
 		$stripe_settings = get_option( 'woocommerce_stripe_settings', [] );
 		if ( isset( $stripe_settings['enabled'], $stripe_settings['express_checkout'] )
 			&& 'yes' === $stripe_settings['enabled']
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -177,6 +177,7 @@ final class Modal_Checkout {
 		add_action( 'wp_ajax_nopriv_process_name_your_price_request', [ __CLASS__, 'process_name_your_price_request' ] );
 		add_filter( 'option_woocommerce_woocommerce_payments_settings', [ __CLASS__, 'filter_woocommerce_payments_settings' ] );
 		add_action( 'init', [ __CLASS__, 'unhook_woocommerce_payments_update_billing_fields' ] );
+		add_action( 'init', [ __CLASS__, 'unhook_woocommerce_payments_express_checkout_buttons' ], 20 );
 		add_action( 'wp_enqueue_scripts', [ __CLASS__, 'update_password_strength_message' ], 9999 );
 		add_filter( 'woocommerce_enforce_password_strength_meter_on_checkout', '__return_true' );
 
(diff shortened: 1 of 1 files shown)


includes/class-modal-checkout.php
```diff
@@ -472,9 +472,25 @@ public static function filter_woocommerce_payments_settings( $settings ) {
...
+		// (payment_request) to prevent duplicate Apple Pay / Google Pay buttons in the modal.
+		// Stripe is preferred: Newspack already manages its express checkout behavior (see
+		// class-woocommerce-gateway-stripe.php), and WooPayments is already partially filtered
+		// above (platform_checkout). WooPayments is suppressed, not the reverse.
+		if ( class_exists( 'WC_Stripe' ) ) {
+			$stripe_settings = get_option( 'woocommerce_stripe_settings', [] );
+			if ( isset( $stripe_settings['express_checkout'] ) && 'yes' === $stripe_settings['express_checkout'] ) {
+				if ( isset( $settings['express_checkout_checkout_methods'] ) ) {
+					$settings['express_checkout_checkout_methods'] = array_values(
+						array_diff( $settings['express_checkout_checkout_methods'], [ 'payment_request' ] )
```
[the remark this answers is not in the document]
wil-gerken: Good catch. Used (array) casting instead to guard against the same issue more concisely.

dkoo (changes requested):
Hmm, this isn't working for me but I'm not sure if it's because of my specific testing site environment. When I log the output of `get_option( 'woocommerce_stripe_settings', [] );` I see `'enabled' => 'yes'` and `'express_checkout' => 'yes'` but I don't see an `express_checkout_checkout_methods` key on the settings array. Subsequently, I still see the double express checkout buttons displayed in modal checkout.

It's possible the settings schema has changed in recent versions of the [`woocommerce-gateway-stripe` plugin](https://github.com/woocommerce/woocommerce-gateway-stripe/). I'm running v10.5.2—what version do you have?

includes/class-modal-checkout.php
```diff
@@ -472,9 +472,24 @@ public static function filter_woocommerce_payments_settings( $settings ) {
...
 		}
+		// Disable WooPay (Woo's hosted checkout experience) — not supported in the modal.
 		if ( isset( $settings['platform_checkout'] ) ) {
 			$settings['platform_checkout'] = 'no';
 		}
+		// If Stripe's express checkout is active, suppress WooPayments' express checkout
+		// (payment_request) to prevent duplicate Apple Pay / Google Pay buttons in the modal.
+		// Stripe is preferred: Newspack already manages its express checkout behavior (see
+		// newspack-plugin/includes/plugins/class-woocommerce-gateway-stripe.php), and WooPayments
+		// is already partially filtered above (platform_checkout). WooPayments is suppressed, not the reverse.
```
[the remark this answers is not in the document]
wil-gerken: Thanks. I updated the comment to remove the broken path reference.

wil-gerken:
> Hmm, this isn't working for me but I'm not sure if it's because of my specific testing site environment. When I log the output of `get_option( 'woocommerce_stripe_settings', [] );` I see `'enabled' => 'yes'` and `'express_checkout' => 'yes'` but I don't see an `express_checkout_checkout_methods` key on the settings array. Subsequently, I still see the double express checkout buttons displayed in modal checkout.
> 
> It's possible the settings schema has changed in recent versions of the [`woocommerce-gateway-stripe` plugin](https://github.com/woocommerce/woocommerce-gateway-stripe/). I'm running v10.5.2—what version do you have?

@dkoo,

Sorry for the slow reply on this, and thanks for catching it!

The issue was that `express_checkout_checkout_methods` may not be present in the database, as your testing uncovered.

The original `isset()` guard caused the suppression to silently no-op in that case. WooPayments then fell back to its hardcoded default, which includes `payment_request`, so the duplicate buttons still appeared.

This revised fix initializes from the WooPayments default when the key is absent, so the suppression now applies regardless of the database state. I was able to reproduce your exact scenario locally by removing the key, and confirmed that the fix works in both cases: when the key is present and when it is absent.

Could you please take another look when you get a chance?

includes/class-modal-checkout.php
```diff
@@ -530,9 +530,24 @@ public static function filter_woocommerce_payments_settings( $settings ) {
...
+		// gateway settings below, while WooPayments is already partially filtered above via
+		// platform_checkout. WooPayments is suppressed, not the reverse.
+		$stripe_settings = get_option( 'woocommerce_stripe_settings', [] );
+		if ( isset( $stripe_settings['enabled'], $stripe_settings['express_checkout'] )
+			&& 'yes' === $stripe_settings['enabled']
+			&& 'yes' === $stripe_settings['express_checkout'] ) {
+			// Always write the key — if it's absent from the DB, WooPayments falls back to its
+			// hardcoded default (['payment_request', 'woopay', 'amazon_pay']), bypassing this filter.
+			$methods = (array) ( $settings['express_checkout_checkout_methods'] ?? [ 'payment_request', 'woopay', 'amazon_pay' ] );
+			$settings['express_checkout_checkout_methods'] = array_values( array_diff( $methods, [ 'payment_request' ] ) );
```
dkoo:
```suggestion
			$settings['payment_request'] = 'no';
```

dkoo (changes requested):
Strangely, it's still not working for me! I verified that the settings are filtered by the `filter_woocommerce_payments_settings` callback with the expected values, too.

If I change the filtered value to the inline suggestion, it does successfully remove the WooPay express checkout buttons. However, I'm not sure if this might have any unintended side effects since it's a much farther-reaching settings change.

includes/class-modal-checkout.php
```diff
@@ -565,6 +566,44 @@ public static function unhook_woocommerce_payments_update_billing_fields() {
...
+		if ( ! isset( $GLOBALS['wp_filter']['woocommerce_checkout_before_customer_details'] ) ) {
+			return;
+		}
+		$filters = $GLOBALS['wp_filter']['woocommerce_checkout_before_customer_details'];
+		foreach ( $filters as $index => $filter ) {
+			$keys = array_keys( $filter );
+			foreach ( $keys as $key ) {
+				if ( strpos( $key, 'display_express_checkout_buttons' ) !== false ) {
+					remove_action( 'woocommerce_checkout_before_customer_details', $key, $index );
+				}
```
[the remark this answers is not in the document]
wil-gerken: Thanks. You're right that the direct `foreach` over `$GLOBALS['wp_filter'][...]` was fragile.

Reading `->callbacks` directly is clearer in intent and avoids relying on `WP_Hook`'s `Iterator` semantics, which are meant for use during `do_action()`.

I've applied a minimal version of your suggestion that does just that:

```php
$hook      = $GLOBALS['wp_filter']['woocommerce_checkout_before_customer_details'];
$callbacks = $hook instanceof \WP_Hook ? $hook->callbacks : (array) $hook;
foreach ( $callbacks as $priority => $priority_callbacks ) {
    foreach ( array_keys( $priority_callbacks ) as $key ) {
        if ( strpos( $key, 'display_express_checkout_buttons' ) !== false ) {
            remove_action( 'woocommerce_checkout_before_customer_details', $key, $priority );
        }
    }
}

File: includes/class-modal-checkout.php (hunk 3 above)
Comment (wil-gerken) [reply includes/class-modal-checkout.php]: @dkoo,

Thanks for catching this twice, and thanks for all of your testing!

I'm still smoothing out my local environment, especially related to Woo and Stripe. It turned out the root cause was something my earlier fix was never going to address.

My local env was misleading me: `are_payments_enabled` was false so WooPayments' display hook wasn't firing, which made the old filter look like it was working. Once I could reproduce it reliably, it became clear the filter was targeting a settings key that WooPayments 10.4+ no longer consults for button display.

I've reworked the fix to target the rendering layer. The new approach `remove_action`s WooPayments' `display_express_checkout_buttons` callback from `woocommerce_checkout_before_customer_details` during modal checkout. This is the same pattern as the existing `unhook_woocommerce_payments_update_billing_fields` method in the file.

It's version-independent and gated on Stripe's express checkout being active, so if someone's running WooPayments without Stripe, WooPayments still renders its express buttons in the modal normally.

I also stripped the now-dead `express_checkout_checkout_methods` filter code `filter_woocommerce_payments_settings` is back to just the original `platform_checkout` (WooPay) suppression.

I was able to reproduce and retest, and it's looking good on my end.

Can you take another look when you get a chance?

File: includes/class-modal-checkout.php [hunk 5]
```diff
@@ -565,6 +566,44 @@ public static function unhook_woocommerce_payments_update_billing_fields() {
...
+		$callbacks = $hook instanceof \WP_Hook ? $hook->callbacks : (array) $hook;
+		foreach ( $callbacks as $priority => $priority_callbacks ) {
+			foreach ( array_keys( $priority_callbacks ) as $key ) {
+				if ( strpos( $key, 'display_express_checkout_buttons' ) !== false ) {
+					remove_action( 'woocommerce_checkout_before_customer_details', $key, $priority );
+				}
+			}
+		}
+	}
+
```
dkoo, line 606: NIT: extra line break here

```suggestion
```

includes/class-modal-checkout.php
```diff
@@ -565,6 +566,44 @@ public static function unhook_woocommerce_payments_update_billing_fields() {
...
+	 * at init priority 15; when Stripe's express checkout is also active the modal renders
+	 * duplicate Apple Pay / Google Pay rows. Removing the hook here suppresses WooPayments'
+	 * buttons while leaving Stripe's intact. Hooked on init at priority 20 so WooPayments'
+	 * init callback has already attached its actions.
+	 */
+	public static function unhook_woocommerce_payments_express_checkout_buttons() {
+		if ( ! self::is_modal_checkout() ) {
+			return;
+		}
+		if ( ! class_exists( 'WC_Payments' ) ) {
```
dkoo: NIT: some sites use neither Stripe nor WooPayments—fairly rare for our platform, but possible.

```suggestion
		if ( ! class_exists( 'WC_Payments' ) || ! class_exists( 'WC_Stripe' ) ) {
```

dkoo (approved):
Working for me, now! Left a couple of non-blocking inline comments. Feel free to merge after resolving.

includes/class-modal-checkout.php
wil-gerken: Thanks. I'll check why the linter didn't catch the extra blank line on my end

includes/class-modal-checkout.php
wil-gerken: Great suggestion, and added. Thanks!

PR #2311, opened 2026-03-09
fix: modal checkout for blocks

chickenn00dle:
### All Submissions:

Closes https://linear.app/a8c/issue/NPPD-1188/modal-checkout-testing-and-fixes

This PR gets Modal checkout working with the block theme.

See https://github.com/Automattic/newspack-plugin/pull/4557 for testing instructions

### Other information:

### DIFF
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -138,6 +138,7 @@ final class Modal_Checkout {
 		add_action( 'wp', [ __CLASS__, 'process_checkout_request' ] );
 		add_action( 'wp_ajax_abandon_modal_checkout', [ __CLASS__, 'process_abandon_checkout' ] );
 		add_action( 'wp_ajax_nopriv_abandon_modal_checkout', [ __CLASS__, 'process_abandon_checkout' ] );
+		add_action( 'wp_loaded', [ __CLASS__, 'process_checkout_action' ], 21 );
 
 		add_filter( 'wp_redirect', [ __CLASS__, 'pass_url_param_on_redirect' ] );
 		add_filter( 'woocommerce_cart_product_cannot_be_purchased_message', [ __CLASS__, 'woocommerce_cart_product_cannot_be_purchased_message' ], 10, 2 );
@@ -292,6 +293,35 @@ final class Modal_Checkout {
 		}
 
 		return false;
+	}
+
+	/**
+	 * Process checkout after order button is clicked in the modal.
+	 */
+	public static function process_checkout_action() {
+		if ( ! self::is_modal_checkout() ) {
+			return;
+		}
+
+		$nonce = isset( $_POST['newspack_checkout_nonce'] ) ? sanitize_text_field( wp_unslash( $_POST['newspack_checkout_nonce'] ) ) : '';
+		if ( ! wp_verify_nonce( $nonce, 'newspack_modal_checkout_nonce' ) ) {
+			wp_send_json_error( [ 'message' => __( 'Invalid nonce.', 'newspack-blocks' ) ] );
+			wp_die();
+		}
+
+		if ( isset( $_POST['newspack_blocks_checkout_action'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Missing
+			wc_nocache_headers();
+
+			if ( \WC()->cart->is_empty() ) {
+				wp_safe_redirect( wc_get_cart_url() );
+				exit;
+			}
+
+			wc_maybe_define_constant( 'WOOCOMMERCE_CHECKOUT', true );
+			// Generate a fresh WC nonce so process_checkout() passes its internal verification.
+			$_REQUEST['woocommerce-process-checkout-nonce'] = wp_create_nonce( 'woocommerce-process_checkout' );
+			\WC()->checkout()->process_checkout();
+		}
 	}
 
 	/**
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -299,7 +299,7 @@ final class Modal_Checkout {
 	 * Process checkout after order button is clicked in the modal.
 	 */
 	public static function process_checkout_action() {
-		if ( ! self::is_modal_checkout() ) {
+		if ( ! self::is_modal_checkout() || ! isset( $_POST['newspack_blocks_checkout_action'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Missing
 			return;
 		}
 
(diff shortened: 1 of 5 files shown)


src/modal-checkout/index.js
```diff
@@ -685,7 +685,7 @@ import { domReady, onCheckoutPlaceOrderProcessing } from './utils';
 					// Serialize form and remove fields that shouldn't be included for validation.
 					const serializedForm = $form.serializeArray().filter( item => ! removeFromValidation.includes( item.name ) );
 					// Add 'update totals' parameter so it just performs validation.
-					serializedForm.push( { name: 'woocommerce_checkout_update_totals', value: '1' } );
+					serializedForm.push( { name: 'newspack_blocks_checkout_action', value: '1' } );
 					// Ajax request.
```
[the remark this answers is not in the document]
chickenn00dle: This shouldn't be a problem since Woos checkout handler depends on two POST flags which we are purposefully not setting: https://github.com/woocommerce/woocommerce/blob/edee2521a4850a767ac95dd6ee900a93c2915ae8/plugins/woocommerce/includes/class-wc-form-handler.php#L397-L410

On our end, we are looking for our own flag and setting the relevant woo POST data before triggering checkout ourselves.

includes/class-modal-checkout.php
```diff
@@ -138,6 +138,7 @@ public static function init() {
 		add_action( 'wp', [ __CLASS__, 'process_checkout_request' ] );
 		add_action( 'wp_ajax_abandon_modal_checkout', [ __CLASS__, 'process_abandon_checkout' ] );
 		add_action( 'wp_ajax_nopriv_abandon_modal_checkout', [ __CLASS__, 'process_abandon_checkout' ] );
+		add_action( 'wp_loaded', [ __CLASS__, 'process_checkout_action' ], 21 );
```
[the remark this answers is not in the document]
chickenn00dle: I've added a check to determine if checkout is processing in https://github.com/Automattic/newspack-blocks/pull/2311/commits/651cea1b79ba4d586eef27cff1891c5e2db7628e. So we will keep this hooked after Woos callback to ensure this check runs after Woos. Woos should never run though since we aren't passing the required POST flags, but it is better for us to be safe.

includes/class-modal-checkout.php
```diff
@@ -294,6 +295,39 @@ public static function has_unsupported_payment_gateway() {
...
+
+		$nonce = isset( $_POST['newspack_checkout_nonce'] ) ? sanitize_text_field( wp_unslash( $_POST['newspack_checkout_nonce'] ) ) : '';
+		if ( ! wp_verify_nonce( $nonce, 'newspack_modal_checkout_nonce' ) ) {
+			wp_send_json_error( [ 'message' => __( 'Invalid nonce.', 'newspack-blocks' ) ] );
+			wp_die();
+		}
+
+		wc_nocache_headers();
+
+		if ( \WC()->cart->is_empty() ) {
```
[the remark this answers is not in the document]
chickenn00dle: This is the same behavior as Woos callback, so I'll keep it.

includes/class-modal-checkout.php
```diff
@@ -294,6 +295,39 @@ public static function has_unsupported_payment_gateway() {
 		return false;
 	}
 
+	/**
+	 * Process checkout after order button is clicked in the modal.
+	 */
+	public static function process_checkout_action() {
+		if ( ! self::is_modal_checkout() || ! isset( $_POST['newspack_blocks_checkout_action'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Missing
+			return;
+		}
```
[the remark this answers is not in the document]
chickenn00dle: Updated in https://github.com/Automattic/newspack-blocks/pull/2311/commits/1299629018a029db82bda7f6cbf4a6720b1bfebf

src/blocks/donate/styles/editor.scss
```diff
@@ -35,7 +35,7 @@
 
 	button,
 	.submit-button {
-		font-family: var(--newspack-theme-font-heading);
+		font-family: var(--newspack-theme-font-heading, sans-serif);
```
laurelfulford, line 38: This fallback is a good catch! Probably something we should move to the respective themes in the future (but definitely unrelated to this change).

src/modal-checkout/checkout.scss
```diff
@@ -841,6 +848,20 @@
 				margin: 0;
 			}
 		}
+
+		td,
+		th {
+			border: 1px solid var(--newspack-ui-color-border);
```
laurelfulford: Nit: this could probably use a fallback - it looks like we use `#ddd` [elsewhere](https://github.com/Automattic/newspack-blocks/blob/7f3e399e9c6ea93dbfaf1c7efe8c5308d7d2fe2e/src/modal-checkout/checkout.scss#L518).

includes/class-modal-checkout.php
```diff
@@ -294,6 +295,44 @@ public static function has_unsupported_payment_gateway() {
...
+
+		wc_maybe_define_constant( 'WOOCOMMERCE_CHECKOUT', true );
+
+		// If this is a validation-only request, set the flag that tells process_checkout() to only validate the order.
+		if ( isset( $_POST['is_validation_only'] ) ) {
+			$_POST['woocommerce_checkout_update_totals'] = '1';
+		}
+		// Generate a fresh WC nonce so process_checkout() passes its internal verification.
```
laurelfulford: Nit: might be good to add a little context about the block theme being the cause of this workaround (either here or with the `newspack_modal_checkout_nonce`), just in case it's ever flagged as unneeded 😅

src/modal-checkout/checkout.scss
```diff
@@ -841,6 +848,20 @@
...
+			border-width: 0 0 1px;
+			word-break: break-all;
+		}
+
+		tr:last-child {
+			td,
+			th {
+				border-bottom: 0;
+			}
```
laurelfulford: There's a bit of a regression with these styles. It looks like the issue is that `tr` can be the last child within `thead`, `tbody`, or `tfoot` but not be the last row in the table.

Here's an example of the Transaction details in trunk vs. this PR, both with the Classic Theme:

Image

Image

I think the fix could just be checking the last `tr` withing `tfoot`, or check for the last `tr` in the last direct child of the `table` (like `table > *:last-child > tr`)?

laurelfulford (changes requested):
Thanks for taking this on, @chickenn00dle! 

It's looking really good overall. I hammered all of your testing steps, plus: NYP, free trials, coupon codes, and my BFFs, PayPal and Braintree. 

I ran into a couple minor issues but it's looking really good in my testing overall!

----

PayPal and Braintree work with this PR with both themes! 🎉 Like now, you have to use the `[woocommerce_checkout]` shortcode, and not the block-based checkout. When you use the latter, PayPal doesn't load, and Braintree won't load the inputs.

I'm going to spin up an issue for this as it'd be nice to have this work with the block-based checkout, but as-is, it meets feature parity 🙂 

----

Minor nit: In the block theme, the transaction details are picking up an underline on the Remove Coupon link that you don't get in the Classic theme: 

Image

src/modal-checkout/checkout.scss
chickenn00dle: Added in https://github.com/Automattic/newspack-blocks/pull/2311/commits/992b3e27549964af158579ea47a53549043fc0bd

includes/class-modal-checkout.php
chickenn00dle: Great idea! Expanded the comment in https://github.com/Automattic/newspack-blocks/pull/2311/commits/992b3e27549964af158579ea47a53549043fc0bd

src/modal-checkout/checkout.scss
chickenn00dle: Good catch! Fixed in https://github.com/Automattic/newspack-blocks/pull/2311/commits/992b3e27549964af158579ea47a53549043fc0bd

chickenn00dle:
Thanks for the review @laurelfulford!  This is ready for another round of review now.

chickenn00dle:
Sorry @laurelfulford, I lied. For some reason all of my changes didn't end up in that commit... Gotta re-push them.

chickenn00dle:
Okay, this is ready for real now: https://github.com/Automattic/newspack-blocks/pull/2311/changes/1f32275679cd9ef2845513acf33b89ac230ba7d3

Sorry for the confusion @laurelfulford

src/modal-checkout/checkout.scss
laurelfulford: Thanks!

includes/class-modal-checkout.php
laurelfulford: Comment looks good -- thank you!

src/modal-checkout/checkout.scss
laurelfulford: Table looks good!

laurelfulford (approved):
Thanks @chickenn00dle! This is all looking good (including the coupon link in the block theme) -- thank you!

PR #2325, opened 2026-03-26
ci: fix dependabot auto-merge for grouped updates

adekbadek:
## Summary

- Fix Dependabot auto-merge workflow to handle grouped (multi-dependency) updates
- `dependabot/fetch-metadata@v2` doesn't reliably populate `update-type` for grouped PRs, so the approve/merge steps were silently skipped
- Adds a fallback condition: if `dependency-group` is set and `update-type` doesn't contain `semver-major`, approve and auto-merge

## Test plan

- [ ] Verify the workflow triggers correctly on the next grouped Dependabot PR
- [ ] Verify major version bumps in grouped updates are NOT auto-approved

### DIFF
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -17,11 +17,34 @@ jobs:
         with:
           github-token: "${{ secrets.GITHUB_TOKEN }}"
 
+      - name: Check for major version bumps in PR body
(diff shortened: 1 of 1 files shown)


adekbadek:
This will be handled in the monorepo. See https://github.com/Automattic/newspack-workspace/pull/70. Closing as part of the monorepo cutover.

PR #2337, opened 2026-05-04
fix: ensure buttons pick up all colours, fonts sizes, font class styles

laurelfulford:
### All Submissions:

This PR originated as a way to fix an issue with the new font options coming in WP 7.0, but it also fixes a couple minor issues with button styles:
* It makes sure the theme's baked-in font sizes work on the front-end
* It makes sure WP 7.0's new font face option works on the checkout button block
* It adds a `has-background` CSS class when custom background colours are applied, not just colours from the default palette; ditto `has-text-color`. 
* It cleans up some extra spacing getting added in between CSS classes

**Edited to add:**

This PR updates `block_classes()` in `includes/class-newspack-blocks.php`, which is shared by all server-rendered blocks. The new `array_filter()` on the return value also affects:

* `homepage-articles` (Content Loop)
* `carousel` (Content Carousel)
* `author-profile` (and the `author-profile-card` template)

Closes NEWS-1754 

1. Test on a site running the latest WP 7.0 RC.
2. Under Appearance > Fonts, click Install Fonts, allow Google Fonts, pick at least one font and install a variant by checking at least one weight/style, and clicking 'Install':

image

3. Create a new post or page, and add a few different checkout button blocks. Make sure to include:
* A block using colours from the theme (background and text)
* A block using custom colours
* A block using font sizes from the theme (like 'Large')
* A block using a custom font size
* In the 'Typography' panel, click the `...` menu and select 'Font', then use it to change the font-family on one of the buttons.
4. Publish and view on the front end and note what's not working -- it should be the baked in font sizes, and font family. If you view source, you'll also see some funky styles (like `background-color: primary`)
5. Apply this PR.
6. Refresh the front end and confirm the block styles are all being picked up - the only one that should be a mismatch with the editor is the padding, but this is due to missing editor styles from the theme; that'll be fixed in a separate PR ([here](https://github.com/Automattic/newspack-theme/pull/2686)).
7. Ensure that blocks with custom background and text colours have the `has-background` and `has-text-color` classes, respectively. 

### Other information:

### DIFF
--- a/includes/class-newspack-blocks.php
+++ b/includes/class-newspack-blocks.php
@@ -396,7 +396,7 @@ class Newspack_Blocks {
 			$classes = array_merge( $classes, $extra );
 		}
 
-		return implode( ' ', array_filter( $classes ) );
+		return implode( ' ', array_filter( $classes, 'strlen' ) );
 	}
 
 	/**
--- a/includes/class-newspack-blocks.php
+++ b/includes/class-newspack-blocks.php
@@ -396,7 +396,7 @@ class Newspack_Blocks {
 			$classes = array_merge( $classes, $extra );
 		}
 
-		return implode( ' ', $classes );
+		return implode( ' ', array_filter( $classes ) );
 	}
 
 	/**
--- a/src/blocks/checkout-button/view.php
+++ b/src/blocks/checkout-button/view.php
@@ -98,9 +98,9 @@ function render_callback( $attributes ) {
 
 	$button = sprintf(
 		'<button class="%1$s" style="%2$s" type="submit">%3$s</button>',
-		$button_classes,
-		$button_styles,
-		$text
+		esc_attr( $button_classes ),
+		esc_attr( $button_styles ),
+		wp_kses_post( $text )
 	);
 
 	// Generate hidden fields for the form.
@@ -168,7 +168,7 @@ function render_callback( $attributes ) {
 		[
 			'wp-block-button',
 			( $font_size || isset( $style['typography']['fontSize'] ) ) ? 'has-custom-font-size' : '',
-			$width ? ' has-custom-width wp-block-button__width-' . esc_attr( $width ) : '',
+			$width ? 'has-custom-width wp-block-button__width-' . esc_attr( $width ) : '',
 		]
 	);
 	return sprintf(
--- a/src/blocks/checkout-button/view.php
+++ b/src/blocks/checkout-button/view.php
@@ -49,6 +49,7 @@ function render_callback( $attributes ) {
 	$background_color           = $attributes['backgroundColor'] ?? '';
 	$gradient                   = $attributes['gradient'] ?? '';
 	$font_size                  = $attributes['fontSize'] ?? '';
+	$font_family                = $attributes['fontFamily'] ?? '';
 	$style                      = $attributes['style'] ?? [];
 	$text_align                 = $attributes['textAlign'] ?? '';
 	$width                      = $attributes['width'] ?? '';
@@ -74,8 +75,6 @@ function render_callback( $attributes ) {
 	$button_styles = Newspack_Blocks::block_styles(
 		$attributes,
 		[
-			$background_color ? 'background-color:' . esc_attr( $background_color ) . ';' : '',
-			$font_size ? 'font-size:' . esc_attr( $font_size ) . ';' : '',
 			$width ? 'width:' . esc_attr( $width ) . '%;' : '',
 			$button_color ? 'color:' . esc_attr( $button_color ) . ';' : '',
 		]
@@ -86,8 +85,11 @@ function render_callback( $attributes ) {
 		$attributes,
 		[
 			'wp-block-button__link',
-			$background_color ? 'has-background has-' . esc_attr( $background_color ) . '-background-color' : '',
-			$gradient ? 'has-background has-' . esc_attr( $gradient ) . '-gradient-background' : '',
+			( $background_color || $gradient || isset( $style['color']['background'] ) || isset( $style['color']['gradient'] ) ) ? 'has-background' : '',
+			$background_color ? 'has-' . esc_attr( $background_color ) . '-background-color' : '',
+			$gradient ? 'has-' . esc_attr( $gradient ) . '-gradient-background' : '',
+			$font_size ? 'has-' . esc_attr( $font_size ) . '-font-size' : '',
+			$font_family ? 'has-' . esc_attr( $font_family ) . '-font-family' : '',
 			$text_align ? 'has-text-align-' . esc_attr( $text_align ) : '',
 			isset( $style['border']['radius'] ) && $style['border']['radius'] === 0 ? 'no-border-radius' : '',
 			$button_color ? 'has-text-color has-' . esc_attr( $button_color ) . '-color' : '',
--- a/src/blocks/checkout-button/view.php
+++ b/src/blocks/checkout-button/view.php
@@ -92,7 +92,8 @@ function render_callback( $attributes ) {
 			$font_family ? 'has-' . esc_attr( $font_family ) . '-font-family' : '',
 			$text_align ? 'has-text-align-' . esc_attr( $text_align ) : '',
 			isset( $style['border']['radius'] ) && $style['border']['radius'] === 0 ? 'no-border-radius' : '',
-			$button_color ? 'has-text-color has-' . esc_attr( $button_color ) . '-color' : '',
+			( $button_color || isset( $style['color']['text'] ) ) ? 'has-text-color' : '',
+			$button_color ? 'has-' . esc_attr( $button_color ) . '-color' : '',
 		]
 	);
 
--- a/src/blocks/checkout-button/view.php
+++ b/src/blocks/checkout-button/view.php
@@ -76,7 +76,6 @@ function render_callback( $attributes ) {
 		$attributes,
 		[
 			$width ? 'width:' . esc_attr( $width ) . '%;' : '',
-			$button_color ? 'color:' . esc_attr( $button_color ) . ';' : '',
 		]
 	);
 
--- a/src/blocks/checkout-button/view.php
+++ b/src/blocks/checkout-button/view.php
@@ -47,6 +47,7 @@ function render_callback( $attributes ) {
 	\Newspack_Blocks::enqueue_view_assets( 'checkout-button' );
 
 	$background_color           = $attributes['backgroundColor'] ?? '';
+	$text_color                 = $attributes['textColor'] ?? '';
 	$gradient                   = $attributes['gradient'] ?? '';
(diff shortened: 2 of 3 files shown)


src/blocks/checkout-button/view.php
```diff
@@ -74,8 +75,6 @@ function render_callback( $attributes ) {
 	$button_styles = Newspack_Blocks::block_styles(
 		$attributes,
 		[
-			$background_color ? 'background-color:' . esc_attr( $background_color ) . ';' : '',
-			$font_size ? 'font-size:' . esc_attr( $font_size ) . ';' : '',
 			$width ? 'width:' . esc_attr( $width ) . '%;' : '',
 			$button_color ? 'color:' . esc_attr( $button_color ) . ';' : '',
 		]
```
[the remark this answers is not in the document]
laurelfulford: Should be fixed in [a7a6249](https://github.com/Automattic/newspack-blocks/pull/2337/commits/a7a62498e7a2248aed41e61d126e5200eea22feb).

includes/class-newspack-blocks.php
```diff
@@ -396,7 +396,7 @@ public static function block_classes( $type, $attributes = array(), $extra = arr
 			$classes = array_merge( $classes, $extra );
 		}
 
-		return implode( ' ', $classes );
+		return implode( ' ', array_filter( $classes ) );
```
chickenn00dle: Non-blocking: `array_filter()` here is cross-cutting — all callers of `block_classes()` (carousel, author-profile, homepage-articles, author-profile-card template) get this empty-string trim, not just checkout-button. Worth a one-line PR-description callout for reviewers of those other blocks.

Also: bare `array_filter()` strips all falsy values, so a legitimate `'0'` class would be dropped. None of the current callers can produce `'0'`, but `array_filter( $classes, 'strlen' )` would be more explicit about intent.

src/blocks/checkout-button/view.php
```diff
@@ -86,11 +84,15 @@ function render_callback( $attributes ) {
 		$attributes,
 		[
 			'wp-block-button__link',
-			$background_color ? 'has-background has-' . esc_attr( $background_color ) . '-background-color' : '',
-			$gradient ? 'has-background has-' . esc_attr( $gradient ) . '-gradient-background' : '',
+			( $background_color || $gradient || isset( $style['color']['background'] ) || isset( $style['color']['gradient'] ) ) ? 'has-background' : '',
+			$background_color ? 'has-' . esc_attr( $background_color ) . '-background-color' : '',
+			$gradient ? 'has-' . esc_attr( $gradient ) . '-gradient-background' : '',
+			$font_size ? 'has-' . esc_attr( $font_size ) . '-font-size' : '',
+			$font_family ? 'has-' . esc_attr( $font_family ) . '-font-family' : '',
```
chickenn00dle, line 90: Non-blocking: font-family / font-size slugs go directly into class names via `esc_attr()`. If a theme ever ships a slug with uppercase or unusual characters (e.g. `Inter`), the emitted `has-Inter-font-family` won't match what core writes (`has-inter-font-family`). In practice the editor lowercases slugs, but `sanitize_html_class()` is one extra line and would harden it.

src/blocks/checkout-button/view.php
```diff
@@ -86,11 +84,15 @@ function render_callback( $attributes ) {
...
 			$text_align ? 'has-text-align-' . esc_attr( $text_align ) : '',
```
chickenn00dle, line 91: Non-blocking: no `has-custom-font-family` companion. A `style.typography.fontSize` (custom) path gets `has-custom-font-size` further down (~line 170), but a `style.typography.fontFamily` (custom) path has no parallel hook. `wp_style_engine_get_styles` still injects the inline CSS, but the asymmetry means themes have nothing to target for custom font-families. Consider parity.

chickenn00dle (approved):
Tested locally and the changes look good — approving. A few non-blocking suggestions inline, plus two out-of-diff observations below.

**`src/blocks/checkout-button/view.php:171`** — pre-existing leading space in `' has-custom-width …'` produces double-spaces in the imploded class list (`array_filter` drops empties but doesn't trim). Since this PR cleans up extra spacing between CSS classes, it'd be worth fixing here too.

**`src/blocks/checkout-button/view.php:99-104`** — `$button_classes` / `$button_styles` are interpolated via `%1$s` / `%2$s` without an outer `esc_attr()`. Individual segments are escaped before imploding, so values are safe today, but a defense-in-depth `esc_attr()` at the `sprintf` site (or moving escaping into `block_classes()` itself) would prevent future contributors from missing it.

src/blocks/checkout-button/view.php
```diff
@@ -86,11 +84,15 @@ function render_callback( $attributes ) {
...
 			isset( $style['border']['radius'] ) && $style['border']['radius'] === 0 ? 'no-border-radius' : '',
-			$button_color ? 'has-text-color has-' . esc_attr( $button_color ) . '-color' : '',
+			( $button_color || isset( $style['color']['text'] ) ) ? 'has-text-color' : '',
```
adekbadek: **Suggestion:** The render callback reads `$attributes['backgroundColor']` (line 88) but never `$attributes['textColor']`. So a *named preset* text colour picked via the block's Text control lands in `textColor` (the slug) and is silently dropped on the front end — no `has-text-color`, no `has-{slug}-color`. The new `( $button_color || isset( $style['color']['text'] ) )` guard covers the *custom-hex* text colour, and `$button_color` covers a *preset* link colour stored under `style.elements.link.color.text` — but not the plain `textColor` attribute, which is what the block's own colour UI actually writes for a preset.

**Confirmed at runtime:** a button with `textColor: "accent"` on a light (`base-2`) background renders with the default white text — effectively invisible — while the editor renders it correctly (`has-base-color`/`has-accent-color`). So there's an editor↔front-end mismatch for the most common text-colour case.

This is *pre-existing* (trunk's `view.php` also doesn't read `textColor`), so the PR doesn't regress anything — but the PR title is "pick up all colours" and it's reworking exactly these lines, so it'd be the natural place to add `$text_color = $attributes['textColor'] ?? '';`, fold it into the `has-text-color` guard, and emit `has-{slug}-color`. Worth also reconsidering the `style.elements.link.color.text` branch: it only fires for the `var:preset|color|{slug}` form (a custom hex there is dropped), and `elements.link` is the `core/button` storage path — for this `<button>` block the colour UI writes to `textColor` / `style.color.text`, so that branch may be dead code.

adekbadek (approved):
Tested this on WP 7.0-RC3 with the block theme: added a page with checkout buttons covering theme palette colours, custom hex colours, theme font size, custom font size, the Cardo theme font family and a custom gradient. Front-end output is correct in every case now — the right `has-{slug}-*` classes, custom values still emitted inline by the style engine, and no more `background-color: primary`-style garbage or stray double-spaces in the class list. Nice fix.

One thing worth deciding on before merge (the `textColor` note inline). Not blocking.

One more out-of-diff observation: no render test for this block (cf. `tests/test-donate-block.php` for the donate block), and this is exactly the class/style serialization that's regressed a couple of times on this branch. A focused test — assert preset attrs produce the `has-{slug}-*` classes with no slug-as-value inline CSS, `style.color.background`/`gradient` trigger `has-background`, and `Newspack_Blocks::block_classes()` emits no leading/trailing/double spaces — would be cheap insurance.

includes/class-newspack-blocks.php
laurelfulford: God call, @chickenn00dle! I've added strlen in [e24b44c](https://github.com/Automattic/newspack-blocks/pull/2337/commits/e24b44cdf11f8841042a680fdc3e4cc89e252264), and updated the PR description to mention the other blocks that will be affected.

src/blocks/checkout-button/view.php
laurelfulford: > In practice the editor lowercases slugs...

It looks like it does in some cases, but not for dynamic blocks (like the Core search block, or this one): if you make one of the colour slugs capitalized in the theme.json, it will come out that way in the front-end and not match the CSS class.

Ditto weird characters -- like `*`. Core doesn't remove them from dynamic Core block classes, even if it does remove them from the classes generated for CSS (they get replaced with a `-`).

Based on this, this block acts like a dynamic block in Core, so I'm going to leave this as is for now. It doesn't sound like theme authors are running into this issue with wacky slug choices 🙂 

Thanks for flagging this though, @chickenn00dle! It was kind of cool to see how this stuff is being handled under the hood!

src/blocks/checkout-button/view.php
laurelfulford: Thanks @chickenn00dle! I'm going to hold off on this one as well for now:

Unlike options like colours or font size, there is no free-form font family input. Any fonts registered through Appearance > Fonts or from a block theme's theme.json generate a CSS class that will be used to assign the fonts. There isn't actually a pathway to get the font assigned via a `style` attribute using this dropdown, which is how custom colours & fonts are applied, and why CSS classes for them have a bit more of a utility. 

If it turns out there IS a use case for this, we can absolutely add it down the road, though!

src/blocks/checkout-button/view.php
laurelfulford: Thanks @adekbadek! 

I couldn't quite reproduce this problem on in RC 4 with the block theme, but the change makes sense from a code perspective (the attribute that's written is actually getting used) -- updated in [a3cbd87](https://github.com/Automattic/newspack-blocks/pull/2337/commits/a3cbd8780a3444a52caca3d1c7770382b1fc7291). 

With this fix, I got a little paranoid and kept `style.elements.link.color.text` as a fallback - I'm not sure if it was ever used and is dead, or is some old remnant of something that could show up in published blocks. After duking it out with Claude, we're both vaguely paranoid about it being used 😅 In my testing right now with the code as is, it gets populated with the same value as `$attributes['textColor']` somehow -- when the code outputs CSS classes for both, you get two identical text colour classes, both with the classic and block theme. 

Hope this is an okay compromise but we can revisit in a future PR if it seems safe to cut entirely!

laurelfulford:
Thank you both for your reviews, @chickenn00dle and @adekbadek! 

I ended up adding both out-of-diff requests, since I think they're solid additions to this PR:
* The leading space on the `has-custom-width` CSS class was removed and class/text escaping was added in [9f78413](https://github.com/Automattic/newspack-blocks/pull/2337/commits/9f78413bc24ed254592b1e15f2a93ced4b19d9e8). 
* I added some tests in [3cb95ce](https://github.com/Automattic/newspack-blocks/pull/2337/commits/3cb95ceb0ce823637d0fc6cfa87ae1cddae862df) to cover the requested cases to avoid these issues getting reintroduced.

I'm going to go ahead and merge this but it'll go to alpha before getting released -- so if any of this sounds wrong or should be revisited, please let me know!

PR #2338, opened 2026-05-05
fix(recaptcha): bypass on Check Payments path in modal checkout

chickenn00dle:
### All Submissions:

When a reader chooses **Check Payments** in the modal checkout, the "Complete payment" button could freeze indefinitely on sites running reCAPTCHA v2 invisible. The cheque gateway has no client-side tokenization to serialize the submit, so the v2 widget could race with checkout fragment refreshes and never produce a token — leaving the form stuck.

This change bypasses reCAPTCHA on the cheque path in the modal checkout, both client-side (the widget is never executed) and server-side (token verification is skipped). Other gateways are unaffected and continue to be protected by reCAPTCHA. The cheque path remains low-risk: no charge is processed, orders are created on hold, and reader account creation/sign-in remain protected by their own captcha layers.

Closes NPPM-2619.

1. On a site with reCAPTCHA v2 invisible enabled and the Check Payments WooCommerce gateway active, open a modal checkout for any product (donation, subscription, or one-time purchase).
2. Fill in billing details and proceed to the payment step.
3. Select **Check Payments** as the payment method. Inspect the `form.checkout` element in DevTools and confirm `data-skip-recaptcha="1"` is present.
4. Click **Complete payment**. The order should submit immediately and land on the thank-you screen with a new on-hold order in WooCommerce.
5. On the same modal, switch back to a card gateway (Stripe / WooPayments). Confirm `data-skip-recaptcha` is removed from the form.
6. Submit a card order and verify reCAPTCHA still fires (no regression for card paths).
7. Toggle "Edit billing details" mid-checkout. Confirm `data-skip-recaptcha` stays set during the edit step regardless of the previously selected payment method, then resolves to the correct value when returning to the payment step.
8. Outside modal checkout (standard `/checkout/` page), submit a cheque order and confirm reCAPTCHA still fires — the bypass is scoped to modal checkout only.

### Other information:

### DIFF
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -1584,8 +1584,14 @@ final class Modal_Checkout {
 		if ( 'checkout' !== $context ) {
 			return $should_verify;
 		}
-		// Skip captcha verification if we're in the modal checkout and this is a validation-only request,
-		// which happens when the user updates their billing details on the first screen of the modal checkout.
+		// All bypasses below are scoped to the modal checkout context so a
+		// crafted POST to standard /checkout/ can't disable reCAPTCHA via
+		// is_validation_only or payment_method=cheque.
+		if ( ! self::is_modal_checkout() ) {
+			return $should_verify;
+		}
+		// Skip captcha verification on the validation-only request fired when
+		// the user updates billing details on the first screen of the modal.
 		if ( self::is_validation_only() ) {
 			return false;
 		}
@@ -1593,11 +1599,9 @@ final class Modal_Checkout {
 		// client-side tokenization to serialize the submit, so v2 invisible
 		// races with updated_checkout resets and freezes the button (NPPM-2619).
 		// Mirrors the data-skip-recaptcha toggle on the modal checkout form.
-		if ( self::is_modal_checkout() ) {
-			$payment_method = isset( $_POST['payment_method'] ) ? sanitize_text_field( wp_unslash( $_POST['payment_method'] ) ) : ''; // phpcs:ignore WordPress.Security.NonceVerification.Missing
-			if ( 'cheque' === $payment_method ) {
-				return false;
-			}
+		$payment_method = isset( $_POST['payment_method'] ) ? sanitize_text_field( wp_unslash( $_POST['payment_method'] ) ) : ''; // phpcs:ignore WordPress.Security.NonceVerification.Missing
+		if ( 'cheque' === $payment_method ) {
+			return false;
 		}
 		return $should_verify;
 	}
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -1595,7 +1595,12 @@ final class Modal_Checkout {
 	 * @return bool True if the request is for validation only.
 	 */
 	private static function is_validation_only() {
-		return boolval( filter_input( INPUT_POST, 'is_validation_only', FILTER_SANITIZE_NUMBER_INT ) );
+		// Read $_POST directly (rather than filter_input) so the recaptcha
+		// bypass branch this gates is exercisable from unit tests —
+		// filter_input(INPUT_POST, ...) returns null under PHP CLI. Nonce
+		// verification is owned by Woo's checkout flow upstream of every
+		// call site.
+		return ! empty( $_POST['is_validation_only'] ); // phpcs:ignore WordPress.Security.NonceVerification.Missing
 	}
 
 	/**
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -1584,8 +1584,20 @@ final class Modal_Checkout {
 		if ( 'checkout' !== $context ) {
 			return $should_verify;
 		}
+		// Skip captcha verification if we're in the modal checkout and this is a validation-only request,
+		// which happens when the user updates their billing details on the first screen of the modal checkout.
 		if ( self::is_validation_only() ) {
 			return false;
+		}
+		// Skip captcha on the Check Payments path. The cheque gateway has no
+		// client-side tokenization to serialize the submit, so v2 invisible
+		// races with updated_checkout resets and freezes the button (NPPM-2619).
+		// Mirrors the data-skip-recaptcha toggle on the modal checkout form.
+		if ( self::is_modal_checkout() ) {
+			$payment_method = isset( $_POST['payment_method'] ) ? sanitize_text_field( wp_unslash( $_POST['payment_method'] ) ) : ''; // phpcs:ignore WordPress.Security.NonceVerification.Missing
+			if ( 'cheque' === $payment_method ) {
+				return false;
+			}
 		}
 		return $should_verify;
 	}
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -1624,10 +1624,10 @@ final class Modal_Checkout {
 	 * billing-edit step and offline gateways with no client-side tokenization
 	 * (e.g. Check Payments). Also covers AJAX checkout (e.g. Apple Pay).
 	 *
-	 * All bypasses are scoped to the modal checkout context and to logged-in
-	 * readers so a crafted POST to standard /checkout/ — or an unauthenticated
-	 * POST to the modal — can't disable reCAPTCHA to spam the reader/order
-	 * tables.
+	 * All bypasses are scoped to the modal checkout context and gated on a valid
+	 * modal-checkout nonce so a crafted POST to standard /checkout/ — or a blind
+	 * POST to the modal without the nonce — can't disable reCAPTCHA to spam the
+	 * reader/order tables.
 	 *
 	 * @param bool   $should_verify Whether to verify the captcha.
 	 * @param string $url The URL from which the verification request originated.
@@ -1640,10 +1640,16 @@ final class Modal_Checkout {
 		if ( ! self::is_modal_checkout() ) {
 			return $should_verify;
 		}
-		// Require an authenticated reader for any bypass. Reader Activation
-		// auto-registers and logs in the reader when they submit billing on the
-		// first modal screen, so the bypasses still apply on the real flow.
-		if ( ! is_user_logged_in() ) {
+		// Require a valid modal-checkout nonce for any bypass. The nonce is
+		// rendered into the modal checkout form (form-checkout.php) and submitted
+		// with the checkout request, so the real modal flow — including guest
+		// donors converting through the modal (NPPM-2619), who are not yet
+		// registered/logged in at woocommerce_checkout_process time — still hits
+		// the bypass. A crafted POST to standard /checkout/ with modal_checkout=1
+		// but no valid nonce falls through to standard reCAPTCHA verification,
+		// closing the spam vector without locking out unauthenticated readers.
+		$nonce = isset( $_POST['newspack_checkout_nonce'] ) ? sanitize_text_field( wp_unslash( $_POST['newspack_checkout_nonce'] ) ) : ''; // phpcs:ignore WordPress.Security.NonceVerification.Missing
+		if ( ! wp_verify_nonce( $nonce, 'newspack_modal_checkout_nonce' ) ) {
 			return $should_verify;
 		}
 		// Skip captcha verification on the validation-only request fired when
--- a/includes/class-modal-checkout.php
+++ b/includes/class-modal-checkout.php
@@ -134,6 +134,21 @@ final class Modal_Checkout {
 	];
 
 	/**
+	 * Offline payment gateways with no client-side tokenization. These race with
+	 * Woo's updated_checkout reset under reCAPTCHA v2 invisible and freeze the
+	 * submit button (NPPM-2619), so reCAPTCHA verification is bypassed for them
+	 * inside the modal checkout. Mirrored in JS via the localized
+	 * `newspackBlocksModalCheckout.recaptcha_bypass_gateways` array.
+	 *
+	 * @var string[]
+	 */
+	private static $recaptcha_bypass_gateways = [
+		'bacs', // Direct bank transfer.
+		'cheque',
+		'cod', // Cash on delivery.
+	];
+
+	/**
 	 * Initialize hooks.
 	 */
 	public static function init() {
@@ -272,6 +287,23 @@ final class Modal_Checkout {
 		 * @param array $supported_gateways
 		 */
 		return apply_filters( 'newspack_blocks_modal_checkout_supported_gateways', self::$supported_gateways );
+	}
+
+	/**
+	 * Get the list of payment gateways for which reCAPTCHA verification should
+	 * be bypassed inside the modal checkout. Filterable via
+	 * `newspack_blocks_modal_checkout_recaptcha_bypass_gateways`.
+	 *
+	 * @return string[] Bypass gateway IDs.
+	 */
+	public static function get_recaptcha_bypass_gateways() {
+		/**
+		 * Filters the list of payment gateways that bypass reCAPTCHA verification
+		 * inside the modal checkout.
+		 *
+		 * @param string[] $bypass_gateways Bypass gateway IDs.
+		 */
+		return apply_filters( 'newspack_blocks_modal_checkout_recaptcha_bypass_gateways', self::$recaptcha_bypass_gateways );
 	}
 
 	/**
@@ -844,14 +876,15 @@ final class Modal_Checkout {
 			'newspack-blocks-modal-checkout',
 			'newspackBlocksModalCheckout',
 			[
-				'ajax_url'              => admin_url( 'admin-ajax.php' ),
-				'nyp_nonce'             => wp_create_nonce( 'newspack_checkout_name_your_price' ),
-				'checkout_nonce'        => wp_create_nonce( 'newspack_modal_checkout_nonce' ),
-				'newspack_class_prefix' => self::get_class_prefix(),
-				'is_checkout_complete'  => function_exists( 'is_order_received_page' ) && is_order_received_page(),
-				'divider_text'          => esc_html__( 'Or', 'newspack-blocks' ),
-				'is_error'              => ! is_checkout() && ! is_order_received_page(),
-				'labels'                => [
+				'ajax_url'                  => admin_url( 'admin-ajax.php' ),
+				'nyp_nonce'                 => wp_create_nonce( 'newspack_checkout_name_your_price' ),
+				'checkout_nonce'            => wp_create_nonce( 'newspack_modal_checkout_nonce' ),
+				'newspack_class_prefix'     => self::get_class_prefix(),
+				'is_checkout_complete'      => function_exists( 'is_order_received_page' ) && is_order_received_page(),
+				'divider_text'              => esc_html__( 'Or', 'newspack-blocks' ),
+				'is_error'                  => ! is_checkout() && ! is_order_received_page(),
+				'recaptcha_bypass_gateways' => array_values( self::get_recaptcha_bypass_gateways() ),
+				'labels'                    => [
 					'billing_details'  => self::get_modal_checkout_labels( 'billing_details' ),
 					'shipping_details' => self::get_modal_checkout_labels( 'shipping_details' ),
 					'gift_recipient'   => self::get_modal_checkout_labels( 'gift_recipient' ),
@@ -1574,7 +1607,15 @@ final class Modal_Checkout {
 	}
 
 	/**
-	 * Prevent reCAPTCHA from being verified for AJAX checkout (e.g. Apple Pay).
+	 * Selectively bypass reCAPTCHA verification in the modal checkout for paths
+	 * that can't safely serialize with the v2 widget — the validation-only
+	 * billing-edit step and offline gateways with no client-side tokenization
+	 * (e.g. Check Payments). Also covers AJAX checkout (e.g. Apple Pay).
+	 *
+	 * All bypasses are scoped to the modal checkout context and to logged-in
+	 * readers so a crafted POST to standard /checkout/ — or an unauthenticated
+	 * POST to the modal — can't disable reCAPTCHA to spam the reader/order
+	 * tables.
 	 *
 	 * @param bool   $should_verify Whether to verify the captcha.
 	 * @param string $url The URL from which the verification request originated.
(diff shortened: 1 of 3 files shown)


src/modal-checkout/index.js
```diff
@@ -106,6 +106,17 @@ import { domReady, onCheckoutPlaceOrderProcessing } from './utils';
...
+					// races with updated_checkout resets and freezes the button (NPPM-2619).
+					// Skip while editing billing details — setEditingDetails owns the attr there.
+					if ( ! $form.find( '[name="is_validation_only"]' ).length ) {
+						if ( 'cheque' === selected ) {
+							$form.attr( 'data-skip-recaptcha', '1' );
+						} else {
+							$form.removeAttr( 'data-skip-recaptcha' );
+						}
+					}
 				}
```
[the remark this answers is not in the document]
chickenn00dle: Addressed in d492471. Added an explicit `handlePaymentMethodSelect()` call at the end of the `setEditingDetails(false)` branch so the cheque bypass is reapplied when returning from the edit-billing step.

includes/class-modal-checkout.php
```diff
@@ -1546,9 +1546,21 @@ public static function recaptcha_verify_captcha( $should_verify, $url, $context
 		if ( 'checkout' !== $context ) {
 			return $should_verify;
 		}
+		// Skip captcha verification if we're in the modal checkout and this is a validation-only request,
+		// which happens when the user updates their billing details on the first screen of the modal checkout.
 		if ( self::is_validation_only() ) {
```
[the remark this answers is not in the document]
chickenn00dle: Addressed in d492471. Restructured `recaptcha_verify_captcha` so all bypasses (validation-only and cheque) are gated on `is_modal_checkout()`. A crafted POST to standard `/checkout/` will no longer disable reCAPTCHA via `is_validation_only=1` or `payment_method=cheque`.

src/modal-checkout/index.js
```diff
@@ -106,6 +106,17 @@ import { domReady, onCheckoutPlaceOrderProcessing } from './utils';
 					const selected = $( 'input[name="payment_method"]:checked' ).val();
 					$( '.wc_payment_method' ).removeClass( 'selected' );
 					$( '.wc_payment_method.payment_method_' + selected ).addClass( 'selected' );
+					// Skip reCAPTCHA on the Check Payments path. The cheque gateway has no
+					// client-side tokenization to serialize the submit, so v2 invisible
+					// races with updated_checkout resets and freezes the button (NPPM-2619).
+					// Skip while editing billing details — setEditingDetails owns the attr there.
+					if ( ! $form.find( '[name="is_validation_only"]' ).length ) {
+						if ( 'cheque' === selected ) {
```
adekbadek: **Blocker:** v2-invisible + cheque: `data-skip-recaptcha` makes `Place order` a dead button — On a site running v2 invisible reCAPTCHA (the NPPM-2619 environment), this attribute toggle doesn't unblock the cheque submit — it just changes the failure mode. The newspack-plugin reCAPTCHA setup renders a visible clone (`#place_order_clone`) and CSS-hides the real `#place_order` inside modal checkout (`newspack-plugin/src/other-scripts/recaptcha/style.scss:29-36`). The clone's click handler in `newspack-plugin/src/other-scripts/recaptcha/index.js:96-107` calls `e.preventDefault(); e.stopImmediatePropagation(); handleSubmit(e);`. `handleSubmit` only triggers a submit path when reCAPTCHA needs to run — with `data-skip-recaptcha=1` set, it takes the else branch (removes `data-recaptcha-validated`) and falls through. No `grecaptcha.execute()`, no `form.submit()`, no AJAX — just a click that registers nothing.

Verified structurally on this env:

1. `wp eval` over `apply_filters( 'woocommerce_order_button_html', $html )` produced `<button type="submit" ... id="place_order_clone" id="place_order" ...>` (two `id` attrs — pre-existing duplication in `Recaptcha::order_button_html`).
2. Fed to a real HTML parser, the first `id` wins; `document.getElementById('place_order_clone')` resolves to the clone and the click handler attaches.
3. CSS hides `#place_order` so the user only sees the clone.
4. The clone's click handler then bails as described above when `data-skip-recaptcha=1`.

A fix needs to either (a) make the clone-click handler fall through to the default form-submit when `data-skip-recaptcha` is set; (b) skip rendering the v2 widget entirely for the cheque path (tear it down on `payment_method_selected`); or (c) submit the form explicitly from the JS toggle when cheque is chosen. Recommend (a) — keeps the rest of the contract intact and is the smallest change.

includes/class-modal-checkout.php
```diff
@@ -1546,9 +1546,25 @@ public static function recaptcha_verify_captcha( $should_verify, $url, $context
...
+		// the user updates billing details on the first screen of the modal.
 		if ( self::is_validation_only() ) {
 			return false;
 		}
+		// Skip captcha on the Check Payments path. The cheque gateway has no
+		// client-side tokenization to serialize the submit, so v2 invisible
+		// races with updated_checkout resets and freezes the button (NPPM-2619).
+		// Mirrors the data-skip-recaptcha toggle on the modal checkout form.
+		$payment_method = isset( $_POST['payment_method'] ) ? sanitize_text_field( wp_unslash( $_POST['payment_method'] ) ) : ''; // phpcs:ignore WordPress.Security.NonceVerification.Missing
+		if ( 'cheque' === $payment_method ) {
```
adekbadek: **Suggestion:** Hardcoded `'cheque'` — other tokenization-less gateways share the same root cause — The PR description identifies the root cause as *"the cheque gateway has no client-side tokenization to serialize the submit"* — but that property is shared by other offline gateways already in `$supported_gateways` (`bacs`, `cod`). A publisher using Direct bank transfer or Cash on delivery will hit the same v2-invisible race. Consider matching against a small allowlist (`['cheque', 'bacs', 'cod']`) mirrored in JS — or extracting a single source of truth (a `RECAPTCHA_BYPASS_GATEWAYS` constant exposed through the existing `newspackBlocksModalCheckout` localized object, plus a `newspack_modal_checkout_recaptcha_bypass_gateways` filter for custom offline gateways). That also keeps the PHP and JS halves from drifting silently if the gateway list ever changes.

src/modal-checkout/index.js
```diff
@@ -106,6 +106,17 @@ import { domReady, onCheckoutPlaceOrderProcessing } from './utils';
 					const selected = $( 'input[name="payment_method"]:checked' ).val();
 					$( '.wc_payment_method' ).removeClass( 'selected' );
 					$( '.wc_payment_method.payment_method_' + selected ).addClass( 'selected' );
+					// Skip reCAPTCHA on the Check Payments path. The cheque gateway has no
+					// client-side tokenization to serialize the submit, so v2 invisible
+					// races with updated_checkout resets and freezes the button (NPPM-2619).
+					// Skip while editing billing details — setEditingDetails owns the attr there.
+					if ( ! $form.find( '[name="is_validation_only"]' ).length ) {
```
adekbadek: **Suggestion:** Edit-mode detection by DOM probing is brittle — `! $form.find( '[name="is_validation_only"]' ).length` infers edit-mode by probing the hidden input that `setEditingDetails(true)` injects. Works today, but couples this branch to an implementation detail of a function ~400 lines away. The form already triggers `editing_details` (line 577); listening for it and caching the state in a closure variable would be cleaner — and would remove the need for the explicit `handlePaymentMethodSelect()` re-invocation at line 575.

includes/class-modal-checkout.php
```diff
@@ -1546,9 +1546,25 @@ public static function recaptcha_verify_captcha( $should_verify, $url, $context
 		if ( 'checkout' !== $context ) {
 			return $should_verify;
 		}
+		// All bypasses below are scoped to the modal checkout context so a
```
adekbadek: **Suggestion:** Function docblock is stale — The function's docblock above (`/** Prevent reCAPTCHA from being verified for AJAX checkout (e.g. Apple Pay). */`) no longer describes what it does — with this PR it also covers the validation-only billing edit step and the cheque path. Consider something like *"Selectively bypass reCAPTCHA verification in the modal checkout for paths that can't safely serialize with the v2 widget"*.

includes/class-modal-checkout.php
adekbadek: **Blocker:** Abuse vector: unauthenticated spam path via crafted POST — `is_modal_checkout()` is satisfied by any caller including `modal_checkout=1` (a public marker, not a signed token), so a crafted POST to `/checkout/` with `modal_checkout=1` and `payment_method=cheque` bypasses reCAPTCHA entirely. No money moves and orders land on-hold, but with Reader Activation enabled each request can create a customer/reader record — an attacker can spam the users table with arbitrary emails, and the on-hold order queue with junk. The bypass intentionally disables reCAPTCHA on a gateway it was protecting, with no compensating control. Mitigations to consider: a per-IP rate limit on the cheque path, a signed nonce on the modal-checkout marker, or scoping the bypass to authenticated/logged-in readers when the underlying race only affects v2 invisible.

includes/class-modal-checkout.php
```diff
@@ -1546,9 +1546,25 @@ public static function recaptcha_verify_captcha( $should_verify, $url, $context
...
 		if ( self::is_validation_only() ) {
 			return false;
 		}
+		// Skip captcha on the Check Payments path. The cheque gateway has no
+		// client-side tokenization to serialize the submit, so v2 invisible
+		// races with updated_checkout resets and freezes the button (NPPM-2619).
+		// Mirrors the data-skip-recaptcha toggle on the modal checkout form.
+		$payment_method = isset( $_POST['payment_method'] ) ? sanitize_text_field( wp_unslash( $_POST['payment_method'] ) ) : ''; // phpcs:ignore WordPress.Security.NonceVerification.Missing
+		if ( 'cheque' === $payment_method ) {
+			return false;
```
adekbadek, line 1670: **Suggestion:** Small PHP unit test would lock the scoping invariant — The filter callback is pure and trivial to unit-test: set `$_POST`/`$_REQUEST`, call `Modal_Checkout::recaptcha_verify_captcha( true, $url, 'checkout' )`, assert the four cases — non-modal POST → unchanged; modal + `is_validation_only` → false; modal + `payment_method=cheque` → false; modal + any other gateway → unchanged. Cheap insurance on the scoping invariant promised in the inline security comment.

includes/class-modal-checkout.php
```diff
@@ -1546,9 +1546,25 @@ public static function recaptcha_verify_captcha( $should_verify, $url, $context
...
+		// Skip captcha verification on the validation-only request fired when
+		// the user updates billing details on the first screen of the modal.
 		if ( self::is_validation_only() ) {
 			return false;
 		}
+		// Skip captcha on the Check Payments path. The cheque gateway has no
+		// client-side tokenization to serialize the submit, so v2 invisible
+		// races with updated_checkout resets and freezes the button (NPPM-2619).
+		// Mirrors the data-skip-recaptcha toggle on the modal checkout form.
+		$payment_method = isset( $_POST['payment_method'] ) ? sanitize_text_field( wp_unslash( $_POST['payment_method'] ) ) : ''; // phpcs:ignore WordPress.Security.NonceVerification.Missing
```
adekbadek, line 1668: **Nit:** `filter_input` would let the `phpcs:ignore` go away — Other reads in this file (`is_validation_only()` at 1520, `is_express_checkout()` at 1687) use `filter_input( INPUT_POST, ..., FILTER_SANITIZE_* )`. Switching to `filter_input( INPUT_POST, 'payment_method', FILTER_SANITIZE_FULL_SPECIAL_CHARS )` would let the `phpcs:ignore WordPress.Security.NonceVerification.Missing` come off (the rule doesn't fire on `filter_input`). Purely cosmetic.

src/modal-checkout/index.js
```diff
@@ -106,6 +106,17 @@ import { domReady, onCheckoutPlaceOrderProcessing } from './utils';
...
+						if ( 'cheque' === selected ) {
+							$form.attr( 'data-skip-recaptcha', '1' );
+						} else {
+							$form.removeAttr( 'data-skip-recaptcha' );
+						}
+					}
 				}
 				$( 'input[name="payment_method"]' ).change( handlePaymentMethodSelect );
```
adekbadek, line 136: **Nit:** Three overlapping listeners for `handlePaymentMethodSelect` — The handler is wired to `input.change`, `document.payment_method_selected`, and `document.updated_checkout`. `payment_method_selected` is the canonical Woo event and fires after Woo's own bookkeeping; the raw `.change()` binding is redundant on top of it. The handler is idempotent so this is harmless, but a one-line comment noting the deliberate triple-bind (or trimming the `.change()`) would help future readers.

adekbadek (changes requested):
Two things need attention before merge: on v2 invisible the fix doesn't actually unblock the cheque submit (browser-verified end-to-end), and the server-side bypass opens an unprotected spam path on the cheque gateway. Details in the inlines.

src/modal-checkout/index.js
chickenn00dle: Addressed in ffe69aa4. Added a capture-phase `click` listener on `document` for `#place_order_clone` that intercepts ahead of the clone's element-level listener when `data-skip-recaptcha` is set, calls `stopImmediatePropagation`, and force-submits via `form.requestSubmit()`. The bypass now actually unblocks the submit instead of just changing the failure mode.

includes/class-modal-checkout.php
chickenn00dle: Addressed in ffe69aa4. Replaced the hardcoded `'cheque'` with a `bacs/cheque/cod` allowlist exposed via a new `newspack_blocks_modal_checkout_recaptcha_bypass_gateways` filter, and mirrored it to JS through the existing `newspackBlocksModalCheckout` localized data (`recaptcha_bypass_gateways`). PHP is the single source of truth so the two halves can't drift.

src/modal-checkout/index.js
chickenn00dle: Addressed in ffe69aa4. `handlePaymentMethodSelect` now defers to an `isEditingDetails` closure that's updated by a top-level `editing_details` event listener. The DOM probe for `is_validation_only` is gone, and the explicit `handlePaymentMethodSelect()` re-invocation inside `setEditingDetails(false)` was removed — the event listener re-applies the payment-method toggle on the edit→non-edit transition.

includes/class-modal-checkout.php
chickenn00dle: Addressed in ffe69aa4. Rewrote the docblock to describe what the function actually covers — validation-only billing edit, offline gateways with no client-side tokenization, and AJAX checkout — and called out the scoping (modal-only + authenticated reader) as the security invariant.

includes/class-modal-checkout.php
chickenn00dle: Addressed in ffe69aa4. Gated all bypasses on `is_user_logged_in()`. Reader Activation auto-registers and logs in the reader on the billing step, so the real modal flow still hits the bypass; an unauthenticated crafted POST to `/checkout/` with `modal_checkout=1` + `payment_method=cheque` now falls through to the standard reCAPTCHA verification. The test added in 6 locks this invariant.

includes/class-modal-checkout.php
chickenn00dle: Addressed in ffe69aa4. Added `tests/test-modal-checkout-recaptcha.php` covering: non-checkout context, non-modal request, unauthenticated modal, authenticated validation-only, each bypass gateway (`cheque`/`bacs`/`cod`) via dataProvider, a non-bypass gateway, modal with no bypass signals, and the filter customization path.

includes/class-modal-checkout.php
chickenn00dle: Won't fix — kept the `$_POST` direct access with `phpcs:ignore`. `filter_input(INPUT_POST, ...)` returns null under PHP CLI (it reads from the original HTTP request, not the current `$_POST`), so switching would mean the bypass branch can't be exercised by the unit test added in 6. Left an inline comment with the rationale next to the `phpcs:ignore`. Nonce verification is owned by the WooCommerce checkout flow upstream of this filter.

src/modal-checkout/index.js
chickenn00dle: Addressed in ffe69aa4. Added a docblock on `handlePaymentMethodSelect` explaining why the triple-bind on `input.change` / `payment_method_selected` / `updated_checkout` is intentional (distinct firing scenarios; handler body is idempotent) — left as-is rather than trimming `.change()` to keep the initial-bind path explicit.

chickenn00dle:
Thanks for the review @adekbadek!

I've addressed all the feedback and this should be ready for another look. That said, you make a good point about a potential abuse vector. I gated this to logged in only, but am wondering if maybe this approach is not the right way to go after all. WDYT?

src/modal-checkout/index.js
adekbadek: Verified end-to-end in the env: clone-click on the post-PR form state now produces `submitFires: 1, ajaxCount: 1, POST /?wc-ajax=checkout` (was 0/0 pre-fix). Implementation is structurally sound — capture-phase always fires before the clone's element-level listener regardless of attach order, and `requestSubmit()` triggers the regular submit event so Woo's AJAX checkout takes over normally. Worth a brief comment that this is a deliberate monkey-patch of newspack-plugin's recaptcha handler from newspack-blocks; a cleaner home for the fix would be in newspack-plugin's `recaptcha/index.js` `handleSubmit` (branch on `data-skip-recaptcha` in the else path), but the cross-repo coordination cost is probably not worth a one-off bug fix.

includes/class-modal-checkout.php
adekbadek: The unit tests verify the gate exactly as documented, but the premise *"Reader Activation auto-registers and logs in the reader when they submit billing on the first modal screen"* doesn't hold empirically:

- `verify_recaptcha_on_checkout` is hooked on `woocommerce_checkout_process` priority 10 (`newspack-plugin/includes/class-recaptcha.php:37`).
- That fires before WC instantiates the customer via `wc_create_new_customer` (which is where RAS gets its hooks via `woocommerce_new_customer_data` / `woocommerce_created_customer`).
- Walked the real flow against a freshly-set-up env: guest fills billing → clicks Continue → `is_user_logged_in()` stays false (no `wordpress_logged_in_*` cookie, no `logged-in` body class).
- Captured the actual `/?wc-ajax=checkout` response body when a guest tries to submit cheque with the post-PR state in place:
  ```json
  { "result": "failure",
    "messages": "...<li>Please complete the challenge to continue.</li>...",
    "refresh": false, "reload": false }
  ```
  That *"Please complete the challenge to continue."* string comes from `Recaptcha::verify_captcha`'s failure path — confirming the captcha check ran, no token was present (client correctly skipped `grecaptcha.execute`), and the submit failed.

So the bug transforms from freeze to captcha error, but the cheque path still doesn't complete for new donors converting through the modal — which is the actual NPPM-2619 audience.

Three paths forward:

1. **Drop the auth gate.** WC's checkout-nonce + the existing `newspack_modal_checkout_nonce` already form a defense-in-depth (confirmed empirically — even with WC nonce scraped, a crafted POST still bounces on a second layer). The original abuse-vector concern is already mitigated upstream.
2. **Replace the auth gate with explicit `wp_verify_nonce( $_POST['newspack_modal_checkout_nonce'], 'newspack_modal_checkout_nonce' )`.** The nonce is already in the localized data and rendered in the form template; this promotes it to a hard server-side check inside the bypass branch, closing the abuse vector without locking out legitimate guests.
3. Move the bypass to run after `wc_create_new_customer` so the auth gate sees the registered reader. Harder retrofit — the captcha check site is owned by newspack-plugin and hooks earlier in the order.

Path 2 looks cleanest.

adekbadek (changes requested):
Blocker 1 verified in browser — clone-click now fires the form submit + AJAX (was 0/0 pre-fix). Blocker 2's auth gate has a real cost though: `verify_recaptcha_on_checkout` runs on `woocommerce_checkout_process` (before `wc_create_new_customer`, before any RAS auto-registration), so guests submitting cheque on a v2-invisible site fail with a captcha error rather than getting through — captured directly from the `/?wc-ajax=checkout` response on the env. Inlines have the evidence + mitigation paths.

includes/class-modal-checkout.php
chickenn00dle: Thanks Adam — appreciate you walking the real flow and capturing the actual `/?wc-ajax=checkout` response. You're right that the auth gate doesn't hold: `verify_recaptcha_on_checkout` runs on `woocommerce_checkout_process` (pri 10) before `wc_create_new_customer`, so the reader is still logged-out when the filter fires — the freeze just becomes a captcha error for the exact NPPM-2619 audience.

Went with **Path 2** in 7508ba45: dropped `is_user_logged_in()` and gated the bypass on `wp_verify_nonce( $_POST['newspack_checkout_nonce'], 'newspack_modal_checkout_nonce' )`. The nonce is already rendered into the modal form (`form-checkout.php`) and serialized with the checkout submit, so the real modal flow — guests included — still hits the bypass, while a blind POST to `/checkout/` with `modal_checkout=1` and no valid nonce falls through to standard reCAPTCHA. Updated the unit tests to lock the new invariant (missing nonce + garbage nonce both fall through). WDYT?

src/modal-checkout/index.js
chickenn00dle: Thanks for verifying this end-to-end, Adam! Added the note in 7508ba45 — the capture-phase `#place_order_clone` listener now carries a comment calling out that it's a deliberate cross-repo monkey-patch of newspack-plugin's recaptcha handler, and that the cleaner home would be a `data-skip-recaptcha` branch in newspack-plugin's `handleSubmit`. Agreed the cross-repo coordination isn't worth it for this one-off, so I kept it here with the rationale documented.

chickenn00dle:
Thanks for the review @adekbadek! This should be good for another look now.

includes/class-modal-checkout.php
adekbadek: Verified end-to-end:

- WP CLI filter test, 9 cases: missing-nonce and garbage-nonce → pass-through; valid-nonce + cheque/bacs/cod → bypassed; valid-nonce + stripe → pass-through; non-modal flag → pass-through.
- Direct guest AJAX POST without `newspack_checkout_nonce`: rejected at `check_ajax_referer( 'newspack_modal_checkout_nonce' )` (line 547) with `"Invalid nonce."` — never reaches the captcha bypass gate.
- Direct guest AJAX POST with the nonce: `{"result":"success","redirect":"...order-received/69/...","order_id":69}`. The NPPM-2619 audience (first-time donors converting by cheque on the modal) gets through cleanly.

The nonce being enforced at two layers (referer check + bypass gate) is even stricter than the commit message promises — abuse vector closed at the door. Approving.

adekbadek (approved):
Path 2 verified end-to-end. Filter behaves correctly on all 9 cases (WP CLI), and a direct AJAX POST as a guest with the modal nonce now creates the cheque order (`result: success, order_id: 69`) — the exact outcome that returned the captcha error last round. The nonce also turns out to be enforced at the AJAX-referer level upstream of the bypass gate, so the abuse vector is closed at two layers.

